PR #2, opened 2021-03-11
Add read_upto and rename read to read_exactly

talex5:
`read` would keep waiting until it had filled the buffer. When reading e.g. an HTTP request from a network connection, you don't know how long it will be and should just process whatever data arrives.


avsm:
One thing I think we should consider is to expose separate higher level functions for operations over files and network sockets. They're just fundamentally different: you want seeking/offsets in files and buffering, and short read/writes in network operations with precise control over flushing.

PR #3, opened 2021-03-11
Fix stack-overflow in scheduler

talex5:
There's no need to call the scheduler again manually after `continue`. The coroutine will return to the scheduler anyway when done.


talex5:
@avsm : I'm not entirely sure how this is supposed to work. Does this look right to you?

avsm:
This looks right; I'd planned to make complete_rw_req and the schedule function mutually recursive anyway, since there needs to be a direct control flow line back to the scheduler in this case.  We could do with a test case that spawns a few billion fibres to test the stack safety...

PR #26, opened 2021-05-17
Grand Central Dispatch Backend

patricoferris:
This draft PR exists to make it easier to track the [Grand Central Dispatch](https://developer.apple.com/documentation/DISPATCH) (GCD) backend for Eio, the work on the PR-ed branch should be treated as *highly* experimental and unstable for now.


talex5:
As the whole repository is experimental, I'm happy to merge this as long as CI is passing :-)

It will probably need some new library (`eio-main`?) that picks a suitable backend for the platform it's compiled for / running on. Then you can write cross-platform programs using `Eio_main.run ...`.

lib_eio_gcd/dqueue.ml
```diff
@@ -0,0 +1,92 @@
...
+  file_offset : int;
+  len : int;
+  id : int;
+  buff : Dispatch.Data.t ref;
+  data : 'a;
+}
+
+type 'a request = Close of Unix.file_descr * 'a | Rw of 'a rw
+
+type eof = Eof | Eof_with_data | Not_eof
```
avsm, line 14: This type could probably use a little more explanation. I guess there's no `Not_eof_with_data` since that circumstance can never happen, but could it with a TCP half close when the other side has indicated it'll never send more data after the current chunk is complete?

lib_eio_gcd/eio_gcd.ml
```diff
@@ -0,0 +1,297 @@
...
+    if res < 0 then
+      failwith "Dispatch file closing error"
+end
+
+let openfile path flags mode =
+  FD.of_unix (Unix.openfile path flags mode)
+
+(* User data handed back in the completion queue *)
+type data = 
+  | Read of  (int * int Suspended.t)
```
avsm, line 46:
```suggestion
  | Read of  int * int Suspended.t
```

You can remove the extra boxing here I think

patricoferris:
Closing this for now as it has fallen very far behind

PR #27, opened 2021-05-18
Convert expect tests to MDX

talex5:
This looks a bit nicer and avoids pulling in two separate expect-test frameworks.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -18,8 +18,8 @@
  (synopsis "Eio implementation for Linux using io-uring")
  (description
    "An eio implementation for Linux using io-uring.")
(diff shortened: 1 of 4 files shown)


talex5:
This avoids the need to install the following libraries (!):

- octavius
- jane-street-headers
- stdio
- ppx_sexp_conv
- ppx_js_style
- ppx_here
- ppx_enumerate
- ppx_compare
- ppx_cold
- ppx_optcomp
- ppx_hash
- ppx_assert
- ppx_base
- jst-config
- time_now
- ppx_inline_test
- ppx_expect

PR #52, opened 2021-06-15
Split Linux support out to eio_linux library

talex5:
This leaves eunix as a place for common code shared with other backends. At the moment, that's just the sleep queue (with cancellation support) and the suspended type (with tracing).

The tests that weren't specific to Linux have been converted to MDX tests using `Eio_main`, so they can test other platforms too.

These changes might be useful for e.g. @patricoferris's work in #26.


patricoferris:
Thanks @talex5 this is useful! Would a PR/multiple PRs be welcome to optionally build most of the linux stack if we are indeed on a linux machine, I'm simply ignoring the linux-related directories (`ocaml-uring`, `lib_eio_linux`, the linux specific tests etc.) right now but I think it would be better to use `enabled_if` where possible instead? That way mac/linux/windows etc. will be able to just run `dune build` from the root ?

talex5:
Yes, that would be very welcome!

PR #68, opened 2021-07-26
Add eio_luv backend

talex5:
Use [luv](https://github.com/aantron/luv) to provide a cross-platform default backend.


talex5:
@patricoferris: might be interesting to benchmark this against the GCD backend (will need some build changes to get it to select only eio-luv on macos).

talex5:
Note: it probably won't build on other platforms yet due to https://github.com/ocaml/dune/issues/4821

PR #72, opened 2021-08-05
add non-deterministic to abstract domain socket test

patricoferris:
The `eio_luv` library works well on macOS as far as I can tell 🎉 

It should be easy to setup with the fixes coming to `dune.2.9.1` with the `enabled_if`s. The only test that failed was the abstract domain socket one as they aren't supported on macOS. This PR disables that test from running on a regular `dune runtest` using `non-deterministic=command` (it might be a nice feature in mdx to support per block disabling depending on the OS). Alternatively if there are enough differences we could split out linux specific ones to a different `.md` document and use `enabled_if` in the `mdx` stanza?

### DIFF
--- a/tests/test_network.md
+++ b/tests/test_network.md
@@ -89,8 +89,9 @@ Handling one connection on a Unix domain socket:
 Exception: Graceful_shutdown.
 ```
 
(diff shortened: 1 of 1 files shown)


talex5:
Thanks!

PR #83, opened 2021-10-12
Effect handlers have their own module now

sidkshatriya:
See https://github.com/ocaml-multicore/ocaml-multicore/pull/687


talex5:
Thanks!

PR #84, opened 2021-10-13
Rename Obj.Effect_handlers to EffectHandlers

SquidDev:
ocaml-multicore/ocaml-multicore#687 moved the effect handlers module out of `Obj` and so eio no longer builds on the main 4.12+domains branch.

Apologies if this is not the right thing to do - I haven't been following all the multicore discussion, just looking forward to trying it out :).


sidkshatriya:
See https://github.com/ocaml-multicore/eio/pull/83 for the same change!

Issue #87, opened 2021-11-03
eio fails to install due to vendor conflicts


mattpallissard:
Howdy, I'm running into an issue when installing eio.  It appears that there is a conflict with the vendored uring.  Calling submodule deinit resolves the issue.

```
[ERROR] The compilation of eio_main.~dev failed at "dune build -p eio_main -j 3
        --promote-install-files=false @install".

#=== ERROR while compiling eio_main.~dev ======================================#
# context     2.1.0 | linux/x86_64 | ocaml-variants.4.12.0+domains | pinned(git+file:///home/matt/usr/src/mis
c/eio#main#6e739586)
# path        ~/usr/opam/4.12.0+domains/.opam-switch/build/eio_main.~dev
# command     ~/usr/opam/opam-init/hooks/sandbox.sh build dune build -p eio_main -j 3 --promote-install-files
=false @install
# exit-code   1
# env-file    ~/usr/opam/log/eio_main-1151034-395744.env
# output-file ~/usr/opam/log/eio_main-1151034-395744.out
### output ###
# Error: Conflict between the following libraries:
# - "uring" in _build/default/ocaml-uring/lib/uring
# - "uring" in /home/matt/usr/opam/4.12.0+domains/lib/uring
#   -> required by library "eio_linux" in
#      /home/matt/usr/opam/4.12.0+domains/lib/eio_linux
```
```
~  opam --version
2.1.0
~  dune --version
2.9.1
```

PR #88, opened 2021-11-03
Update installation steps in readme.

mattpallissard:
This is to account for issue #87 

If this approach is incorrect, feel free to disregard.

edit: spelling.


talex5 (commented):
Thanks! This fix works as long as uring doesn't change, but we'll probably want to update that alongside this project during development, so I've made an alternative fix in #89 using pin-depends instead of vendoring.

talex5:
Closing in favour of #89.

PR #90, opened 2021-11-08
Implicit cancellation

talex5:
Instead of requiring every cancellable operation to pass a `~sw` argument, give each fibre a default switch and use that. It's too easy to forget to make something cancellable and clutters up the code.

This should enable further API improvements, but the change is big enough that it's probably worth reviewing on its own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,8 @@ This is an unreleased repository, as it's very much a work-in-progress.
 * [Testing with Mocks](#testing-with-mocks)
 * [Fibres](#fibres)
 * [Tracing](#tracing)
-* [Switches, Errors, and Cancellation](#switches-errors-and-cancellation)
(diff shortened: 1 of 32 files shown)


talex5:
I've now pushed some more changes:

- Cancellation has been split out of Switch (though every switch now contains a cancellation context).
- The remaining cancellable operations (e.g. awaiting promises) now use the cancellation context.
- Many other operations no longer take a switch argument (e.g. `Fibre.both`).
- `Fibre.both` creates a new cancellation context, so if one of its fibres fails it just re-raises that exception. It doesn't cancel its parent switch.

Here's the diff for gemini-eio to use the new API:
https://gitlab.com/talex5/gemini-eio/-/compare/master...cancellation?w=1

PR #99, opened 2021-11-22
Make enqueue thread-safe

talex5:
Note that the deadlock detection code had to be deleted. Now that we can be woken up from another domain, we can never be sure we're stuck.

This is just the first part. The next step is to make `Waiters` thread-safe, so promises, streams, etc can be used across domains. Then optimise that using lock-free structures.


talex5:
Added thread-safe promises, streams and semaphores. There is a new `make bench` target to test them:

    dune exec -- ./bench/bench_promise.exe
    Reading a resolved promise: 4.684 ns
    use_domains,   n_iters, ns/iter, promoted/iter
          false,  1000000,   964.73,       26.0096
           true,   100000, 13833.80,       15.7142

    dune exec -- ./bench/bench_stream.exe
    use_domains,  n_iters, capacity, ns/iter, promoted/iter
          false, 10000000,        1,  150.95,        0.0090
          false, 10000000,       10,   76.55,        0.0041
          false, 10000000,      100,   52.67,        0.0112
          false, 10000000,     1000,   51.13,        0.0696
           true,  1000000,        1, 4256.24,        1.0048
           true,  1000000,       10,  993.72,        0.2526
           true,  1000000,      100,  280.33,        0.0094
           true,  1000000,     1000,  287.93,        0.0168

    dune exec -- ./bench/bench_semaphore.exe
    use_domains,  n_iters, ns/iter, promoted/iter
          false, 10000000,   43.36,        0.0001
           true, 10000000,  303.89,        0.0000

Note: this is just a starting point and not well optimised.

Other changes:

- Hook is now defunctionalised (simpler and slightly faster).
- `Promise.state` no longer exposes the internals.
- `Switch.await` has moved to `Waiters`.

PR #111, opened 2021-12-10
Restore domains test

talex5:
The fix to ocaml-multicore has been back-ported to 4.12+domains and is now being used by CI.


talex5:
Thanks @ctk21 for the backport!

PR #117, opened 2021-12-14
Allow to set SO_REUSEPORT option

bikallem:
This PR adds the option of setting SO_REUSEPORT socket option in linux_uring backend.

For libuv/luv backend, it seems it is not easy to set it as yet. This is related to another outstanding issue related to SO_REUSEADDR option in libuv backend. Perhaps once luv releases luv-unix, we can revisit this issue again. For now in-order to ensure the repo builds
I have followed a similar approach as SO_REUSEADDR, i.e. dummy function `luv_reuse_port` analogous to `luv_resuse_addr`.

SO_REUSEPORT can be used to build scalable, high performance network servers in linux version 3.9 and over. This is so because reuseport is used to enable socket sharding on multiple threads or child processes. Listening sockets with SO_REUSEPORT enabled are further
load balanced by the kernel; so we get a performant load balancer for free.

References:
https://www.nginx.com/blog/socket-sharding-nginx-release-1-9-1/
https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ

SO_RESUSEADDR/SO_RESUSEPORT luv issue:
https://github.com/aantron/luv/issues/120

### DIFF
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -802,7 +802,7 @@ module Objects = struct
   let net = object
     inherit Eio.Net.t
 
-    method listen ~reuse_addr ~backlog ~sw listen_addr =
+    method listen ~reuse_addr ~reuse_port  ~backlog ~sw listen_addr =
       let socket_domain, socket_type, addr =
         match listen_addr with
         | `Unix path         ->
@@ -825,6 +825,8 @@ module Objects = struct
       end;
       if reuse_addr then
         Unix.setsockopt sock_unix Unix.SO_REUSEADDR true;
+      if reuse_port then
+        Unix.setsockopt sock_unix Unix.SO_REUSEPORT true;
       let sock = FD.of_unix ~sw ~seekable:false sock_unix in
       Unix.bind sock_unix addr;
       Unix.listen sock_unix backlog;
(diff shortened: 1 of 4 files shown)


lib_eio_linux/eio_linux.ml
```diff
@@ -1017,21 +1021,21 @@ let rec run ?(queue_depth=64) ?(block_size=4096) main =
               fn st k;
               schedule st
             )
-          | ERead args -> Some (fun k -> 
+          | ERead args -> Some (fun k ->
               let k = { Suspended.k; fibre } in
               enqueue_read st k args;
               schedule st)
-          | Close fd -> Some (fun k -> 
+          | Close fd -> Some (fun k ->
```
bikallem, line 1028: It seems my nvim settings has automatically removed trailing newlines and made the diff large. Hope this is okay.

lib_eio_linux/eio_linux.ml
```diff
@@ -821,6 +821,10 @@ module Objects = struct
       end;
       if reuse_addr then
         Unix.setsockopt sock_unix Unix.SO_REUSEADDR true;
+
+      if reuse_port then
+        Unix.setsockopt sock_unix Unix.SO_REUSEPORT true;
+
       let sock = FD.of_unix ~sw ~seekable:false sock_unix in
```
talex5, line 828:
```suggestion
      if reuse_port then
        Unix.setsockopt sock_unix Unix.SO_REUSEPORT true;
      let sock = FD.of_unix ~sw ~seekable:false sock_unix in
```

Minor style nit: removed the blank lines for consistency with the rest of the eio code. You can use a comment if you want to make a block of code stand out, but I don't think we want that here.

lib_eio_linux/eio_linux.ml
talex5: Sure. You added them in the first place ;-)

talex5 (approved):
Thanks!

bikallem:
rebased to address merge conflicts.

talex5:
Thanks!

PR #118, opened 2021-12-15
Improve scheduling of forks

talex5:
- Simplify the code by removing the old `Fork` effect, and implementing it instead in terms of `Fork_ignore`.
- Rename `Fork_ignore` to `Fork`, and the same for the functions. This is the most used version.
- The old `Fiber.fork` is now `Fibre.fork_promise` (and might get removed later).
- Extend the lock-free queue to allow the consuming domain to insert items at the head of the queue.
- When forking, we now schedule the caller at the head of the run-queue, ahead of other work.

Some arguments in favour of the new scheduling order:

- This is the most flexible arrangement (you can add yields gets the other combinations).
- The new order seems a bit more natural (the README examples look better).
- Might be better for the cache in some cases (domainslib works this way).
- It means that `Fibre.both f g` starts `f` and `g` in similar contexts. Previously, `f` started before any queued items, while `g` started after, which was a bit inconsistent.

This also adds some documentation about scheduling order to the ocamldoc (note that the previous documentation for `fork` was wrong) and adds a `rationale.md` file explaining the choice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,10 +267,10 @@ For example:
 ```ocaml
 # Eio_main.run @@ fun _env ->
   Switch.run (fun sw ->
-    Fibre.fork_ignore ~sw
+    Fibre.fork ~sw
       (fun () -> for i = 1 to 3 do traceln "i = %d" i; Fibre.yield () done);
     traceln "First thread forked";
-    Fibre.fork_ignore ~sw
+    Fibre.fork ~sw
       (fun () -> for j = 1 to 3 do traceln "j = %d" j; Fibre.yield () done);
     traceln "Second thread forked; top-level code is finished"
   );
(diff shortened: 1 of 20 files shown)


talex5:
Note: the `bench_yield` benchmark got a bit slower with large numbers of fibres.

It spawns a load of fibres in a loop, and each one yields in a loop. But in the old scheduling it takes a long time to start all the fibres, because all existing fibres have to yield once before we spawn the next fibre. i.e. we do

spawn fibre 1
yield 1
spawn fibre 2
yield 2
yield 1
spawn fibre 3
yield 3
yield 2
yield 1
...

This means that there aren't as many fibres running at once as there should be, allowing it to run faster. I think the new version is better.

PR #122, opened 2021-12-21
lib_eio_linux: raise End_of_file

bikallem:
Eio_linux.read_into doesn't explicitly raise `End_of_file` exception
when read returns 0. Rather, the assertion will be hit in
Eio.Flow.read_into `assert(got > 0)`. This commit ensures that
`Eio_linux.read_into` raises `End_of_file` when read returns 0.

This commit is in similar spirit to `Eio_Luv.read_into` where it
explicitly raises `End_of_file` when read returns 0.

Signed-off-by: Bikal Lem <gbikal+git@gmail.com>

### DIFF
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -753,9 +753,11 @@ module Objects = struct
              (should be fixed in Linux 5.14) *)
           await_readable fd
         );
-        let got = read_upto fd chunk max_len in
-        Cstruct.blit chunk_cs 0 buf 0 got;
-        got
+        match read_upto fd chunk max_len with
+        | 0 -> raise End_of_file
+        | got ->
+            Cstruct.blit chunk_cs 0 buf 0 got;
+            got
 
       method read_methods = []
 


lib_eio_linux/eio_linux.ml
```diff
@@ -753,9 +753,11 @@ module Objects = struct
              (should be fixed in Linux 5.14) *)
           await_readable fd
         );
-        let got = read_upto fd chunk max_len in
-        Cstruct.blit chunk_cs 0 buf 0 got;
-        got
+        match read_upto fd chunk max_len with
```
talex5, line 756: The docs for `read_upto` claim that it raises `End_of_file`. Why isn't that happening?

https://github.com/ocaml-multicore/eio/blob/f22bbafeb22924d5ea5efa96bc2ae734342a4049/lib_eio_linux/eio_linux.ml#L468

bikallem: Ah, I think I missed that while reading the code. Eio_luv is implemented a bit differently, it seems. I will close this PR since it is quite redundant then.

PR #129, opened 2022-01-04
lib_eio: add support for cancelling Eio.Std.Promise.t

bikallem:
Adds support for cancelling `Eio.Std.Promise.t`. Conceptually the mechanism to `cancel' is based on `Promise.break`. 

Signed-off-by: Bikal Lem <gbikal+git@gmail.com>

### DIFF
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -109,13 +109,16 @@ module Std : sig
     (** [resolve t (Ok x)] is [fulfill t x] and
         [resolve t (Error ex)] is [break t ex]. *)
 
+    val cancel : 'a u -> unit
+    (** [cancel t] cancels promise [t]. *)
+
     val fulfilled : 'a -> 'a t
     (** [fulfilled x] is a promise that is already fulfilled with result [x]. *)
 
     val broken : exn -> 'a t
     (** [broken x] is a promise that is already broken with exception [ex]. *)
 
-    val state : 'a t -> [`Unresolved | `Fulfilled of 'a | `Broken of exn]
+    val state : 'a t -> [`Unresolved | `Fulfilled of 'a | `Cancelled | `Broken of exn]
     (** [state t] is the current state of [t].
         If the state is [`Unresolved] then it may change in future, otherwise it won't.
         If another domain has access to the resolver then the state may have already
--- a/lib_eio/promise.ml
+++ b/lib_eio/promise.ml
@@ -24,6 +24,7 @@
 type 'a state =
   | Fulfilled of 'a
   | Broken of exn
+  | Cancelled
   | Unresolved of ('a, exn) result Waiters.t * Mutex.t
   (* The Unresolved state's mutex box contains:
      - Full access to the Waiters.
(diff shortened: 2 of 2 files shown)


talex5:
What is the point of this? How is it different to just `Promise.break r Cancelled`?

bikallem:
The motivation is to provide a function analogous to `Lwt.cancel` in `Promise` module. It is convenient and useful to to having this function in `Promsie` module. I will be using it in eio version of cohttp at least and most likely in `resto` and `tezos-node` node in the future. So having this functionality in one place in useful since I won't have to re-invent `cancel` in all the places that I would be using it.

> Promise.break r Cancelled?
I understood `` `Broken`` to mean failure (which is handled differently in my use cases) rather than cancellation, is this not the case? Also, wouldn't it be better to encapsulate cancellation logic from the API surface?

talex5:
Cancellation in Eio works completely differently to cancellation in Lwt. This is deliberate. We're not intending to make Eio a copy of Lwt's API. The Lwt docs say (http://ocsigen.org/lwt/5.5.0/api/Lwt#2_Cancellation):

> Note: cancelation has proved difficult to understand, explain, and maintain, so use of these functions is discouraged in new code.

But in any case, this PR doesn't do what Lwt does. Lwt lets the holder of the *promise* (not the resolver) cancel it, which also cancels it for all other holders of the promise. This is very bad: it's the actor who makes a promise who should get to choose whether to break it.

Cancelling a resolver (as this PR does) is equivalent to just breaking the promise using `break` and I don't see how this PR would help you.

bikallem:
> Lwt lets the holder of the _promise_ (not the resolver) cancel it, which also cancels it for all other holders of the promise.

Ah, yes, indeed. I don't understand all the mechanics of `Lwt.cancel` but it certainly seems to be doing more than what this PR proposes for eio. https://github.com/ocsigen/lwt/blob/master/src/core/lwt.ml#L1424 

More precisely, my use case is to handle success, cancellation and error cases differently. https://github.com/bikallem/ocaml-cohttp/blob/cohttp-eio/cohttp-eio/src/io.mli#L21-L23 

`success` and `error` cases are straight forward and I have used Fulfilled and Broken respectively. For cancellation use case, my initial approach matches your suggestion, i.e. use Break with special exn type for cancellation(https://github.com/bikallem/ocaml-cohttp/blob/cohttp-eio/cohttp-eio/src/io.mli#L21-L23). It would be convenient and useful to have this encoded/encapsulated in the Promise module itself. 

Regarding the `cancel` function, just handling cancel cases but never being able to actually cancel seems to not make much sense. So if there is a desire to maintain symmetry in the Promise api, I believe we need a `cancel` function. Granted it won't match exactly the functionality of its corresponding analogy Lwt.cancel; but mostly its intention, i.e. cancelling an unfulfilled promise. 

Thankfully I haven't come across use cases where I have had to use or abstract over Lwt.cancel - `cancel: 'a t -> unit` yet, i.e.
```
module IO : sig
  type +a t 
  val cancel: 'a t -> unit 
end 
```.
Such use cases is out of scope of this PR. 

In summary, having `cancellation` functionality as introduced by this PR enhances the usability of the eio Promise module.

bikallem:
Closing this PR as I realize now Promise is quite different from Lwt.

PR #130, opened 2022-01-05
Add Luv polling functions

Sudha247:
Adds Luv polling functions wrapped in Eio, mainly to be used for Lwt integration.

### DIFF
--- a/lib_eio_luv/eio_luv.ml
+++ b/lib_eio_luv/eio_luv.ml
@@ -269,6 +269,37 @@ module Stream = struct
     | bufs -> write t bufs
 end
 
+module Poll = struct
+  let poll_readable fd =
+    let poll = Luv.Poll.init ~loop:(get_loop ()) (Obj.magic fd) |> or_raise in
+    let r = enter (fun t k ->
+        Fibre_context.set_cancel_fn k.fibre (fun ex ->
+            Luv.Poll.stop poll |> or_raise;
+            enqueue_failed_thread t k ex
+          );
+        Luv.Poll.start poll [`READABLE;] (fun r ->
+            Luv.Poll.stop poll |> or_raise;
+            if Fibre_context.clear_cancel_fn k.fibre then enqueue_thread t k r
+          )
+      ) in
+    ignore (or_raise r : Luv.Poll.Event.t list)
+
+  let poll_writable fd =
+    let poll = Luv.Poll.init ~loop:(get_loop ()) (Obj.magic fd) |> or_raise in
+    let r = enter (fun t k ->
+        Fibre_context.set_cancel_fn k.fibre (fun ex ->
+            Luv.Poll.stop poll |> or_raise;
+            enqueue_failed_thread t k ex
+          );
+        Luv.Poll.start poll [`WRITABLE;] (fun r ->
+            Luv.Poll.stop poll |> or_raise;
+            if Fibre_context.clear_cancel_fn k.fibre then enqueue_thread t k r
+          )
+      )
+    in
+    ignore (or_raise r : Luv.Poll.Event.t list)
+end
+
 let sleep_until due =
   let delay = 1000. *. (due -. Unix.gettimeofday ()) |> ceil |> truncate |> max 0 in
   enter @@ fun st k ->
--- a/lib_eio_luv/eio_luv.mli
+++ b/lib_eio_luv/eio_luv.mli
@@ -73,6 +73,19 @@ module File : sig
 
   val mkdir : mode:Luv.File.Mode.t list -> string -> unit or_error
   (** Wraps {!Luv.File.mkdir} *)
+
+end
+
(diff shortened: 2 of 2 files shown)


lib_eio_luv/eio_luv.ml
```diff
@@ -126,6 +126,26 @@ let await_with_cancel ~request fn =
...
+let await_with_cancel_poll ~poll fn =
+  enter (fun st k ->
+    let cancel_reason = ref None in
+    Eio.Private.Fibre_context.set_cancel_fn k.fibre (fun ex ->
+        cancel_reason := Some ex;
+        match Luv.Poll.stop poll with
+        | Ok () -> ()
+        | Error e -> Log.debug (fun f -> f "Cancel failed: %s" (Luv.Error.strerror e))
+      );
+    fn st.loop (fun v ->
```
talex5, line 138: From http://docs.libuv.org/en/v1.x/poll.html#c.uv_poll_start it looks like it will keep calling the callback as long as the condition is true. I suspect you need to stop polling here, but I haven't tested it.

lib_eio_luv/eio_luv.ml
```diff
@@ -126,6 +126,26 @@ let await_with_cancel ~request fn =
         )
     )
 
+let await_with_cancel_poll ~poll fn =
+  enter (fun st k ->
+    let cancel_reason = ref None in
+    Eio.Private.Fibre_context.set_cancel_fn k.fibre (fun ex ->
+        cancel_reason := Some ex;
+        match Luv.Poll.stop poll with
+        | Ok () -> ()
```
talex5, line 135: You need to make sure that the continuation is called in this case. For files, `Request.cancel` results in the callback being called with a cancelled result. I suspect this is not the case for `Poll.stop`.

lib_eio_luv/eio_luv.ml
```diff
@@ -228,6 +248,23 @@ module File = struct
     await_with_cancel ~request (fun loop -> Luv.File.mkdir ~loop ~request ~mode path)
 end
 
+module Poll = struct
+
+  let poll_readable fd callback =
+    let p = Luv.Poll.init ~loop:(get_loop ()) (Obj.magic fd) in
+    match p with
+      | Ok poll ->
```
talex5, line 256: Strange indentation here.

You can simplify this with `|> or_raise` like the rest of the code uses, which will also avoid throwing away the error message.

lib_eio_luv/eio_luv.ml
```diff
@@ -228,6 +248,23 @@ module File = struct
...
+        await_with_cancel_poll ~poll (fun _loop _fibre -> Luv.Poll.start poll [`READABLE;] (fun _ -> callback ()))
```
talex5, line 257: You can't throw away the calling fibre here. It needs to be resumed with the result.

lib_eio_luv/eio_luv.mli
```diff
@@ -73,6 +73,19 @@ module File : sig
 
   val mkdir : mode:Luv.File.Mode.t list -> string -> unit or_error
   (** Wraps {!Luv.File.mkdir} *)
+
+end
+
+module Poll : sig
+
+  val poll_readable : Unix.file_descr -> (unit -> unit) -> unit
```
talex5, line 81:
```suggestion
  val poll_readable : Unix.file_descr -> unit
```
The function should just suspend the calling fibre until the result is ready. We don't need to use callbacks now we have effects (even if that is convenient for integrating with Lwt's API).

talex5 (changes requested):
Thanks - getting polling working will be very useful for `Lwt_eio`.

I think you might find it easier to base this code on the `Stream` module rather than on the `File` one, as polling in luv seems to be start/stop based rather than one-shot. I haven't used it though, and could be wrong.

Has this code been tested with `Lwt_eio`? It would be good to add a test case (e.g. in `lib_eio_luv/tests/poll.md`) to make sure it keeps working.

Sudha247:
Thanks for the useful comments @talex5! I haven't looked at the `Stream` module, will see if we can base poll on that.

Sudha247:
I've rewritten `Poll` basing on `Stream`. Here's a branch of Lwt_eio that works with the Luv backend: https://github.com/Sudha247/lwt_eio/tree/luv. I'll try to add a test for this in eio.

talex5 (commented):
Thanks - this looks better! I pushed a few minor changes (e.g. fixed indentation with ocp-indent, removed the unused `type t`, added the type to `ignore`, formatted it to look more like the rest of the code).

As this isn't directly wrapping the Poll interface (just using it), perhaps it should go somewhere else in the `eio_luv.mli` file? It might be better to call the functions `await_readable` etc, for consistency with eio_linux. But that still wouldn't directly allow code to work with both, so we'll still have to think about how to do that.

PR #132, opened 2022-01-05
Don't lose exceptions when cancelled

talex5:
Before, if a fibre failed when its parent context was cancelled then we just reported the parent's cancellation exception. But that could hide other errors, in particular errors during cleanup.

Now, cancellation contexts track whether they were cancelled by their parents.


talex5:
Closing in favour of #134.

PR #139, opened 2022-01-07
Add eio.unix module for Unix integration

### LINKED ISSUE #127: Provide Eio_unix API
Eio needs to work in browsers and unikernels, so `Eio` itself must not depend on `Unix`.

Provide an `Eio_unix` module to extend the basic API with Unix-specific features, such as extracting a `Unix.file_descr` from a flow, waiting for an FD to become readable, etc.

At the moment the `Eio_linux` and `Eio_luv` backends provide separate APIs for this, but there should be a shared API both can implement.

talex5:
For now, this provides `await_readable` and `await_writable`. This allows `Lwt_eio` to work with either backend.

Closes #127.

/cc @Sudha247

### DIFF
--- a/lib_eio/unix/dune
+++ b/lib_eio/unix/dune
@@ -1,0 +1,4 @@
+(library
+ (name eio_unix)
+ (public_name eio.unix)
+ (libraries eio unix))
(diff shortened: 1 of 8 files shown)


Sudha247:
This is neat, thanks!

PR #141, opened 2022-01-12
lib_eio: implement Stream.is_empty

bikallem:
Attempts to implement Eio.Stream.is_empty function. Not 100% sure if the below implementation is adequate since I am not super familiar with Waiters yet. 

The use case is to implement Cohttp Body.is_empty function when using Eio.Stream as a http body.

### DIFF
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -308,7 +308,13 @@ module Stream : sig
       it returns [None] if the stream is empty rather than waiting.
       Note that if another domain may add to the stream then a [None]
       result may already be out-of-date by the time this returns. *)
-end  
+
+  val length : 'a t -> int
+  (** [length t] returns the number of items currently in [t]. *)
+
+  val is_empty : 'a t -> bool
+  (** [is_empty t] is [length t = 0]. *)
+end
 
 (** Cancelling other fibres when an exception occurs. *)
 module Cancel : sig
@@ -734,7 +740,7 @@ module Private : sig
     type 'a enqueue = ('a, exn) result -> unit
     (** A function provided by the scheduler to reschedule a previously-suspended thread. *)
 
-    type _ eff += 
+    type _ eff +=
       | Suspend : (Fibre_context.t -> 'a enqueue -> unit) -> 'a eff
       (** [Suspend fn] is performed when a fibre must be suspended
           (e.g. because it called {!Promise.await} on an unresolved promise).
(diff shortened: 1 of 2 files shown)


lib_eio/stream.ml
```diff
@@ -118,3 +118,10 @@ let take_nonblocking t =
     end;
     Mutex.unlock t.mutex;
     Some v
+
+let is_empty t =
+  let e = ref false in
+  Mutex.lock t.mutex;
+  e := Queue.is_empty t.items && Waiters.is_empty t.writers;
```
talex5, line 125:
```suggestion
  let e = Queue.is_empty t.items && Waiters.is_empty t.writers in
```
Can't see any reason to use a ref here.

talex5 (commented):
The documentation needs to clarify what `is_empty` does for a stream with a capacity of 0.

It would seem natural to define a `length` operation that returns the number of items in the stream (`0 <= length t <= capacity t`), and define `is_empty t = (length t = 0)`. The `is_empty` in this PR is doing something a little different: it also counts a stream as non-empty if there is a writer waiting. That might be more useful, but it should be mentioned. Perhaps calling the function something like `item_available` would be clearer.

bikallem:
As per suggestions, I have now added `length` and used it to define `is_empty`. The new version doesn't take into account waiters since - as you noted - that would have leaked the abstraction implementation details on to the users of the module.

lib_eio/eio.mli
```diff
@@ -308,7 +308,13 @@ module Stream : sig
       it returns [None] if the stream is empty rather than waiting.
       Note that if another domain may add to the stream then a [None]
       result may already be out-of-date by the time this returns. *)
-end  
+
+  val length : 'a t -> int
+  (** [length t] returns the length of items available in [t]. *)
```
talex5, line 313:
```suggestion
  (** [length t] returns the number of items of items currently in [t]. *)
```

lib_eio/eio.mli
```diff
@@ -308,7 +308,13 @@ module Stream : sig
...
+
+  val is_empty : 'a t -> bool
+  (** [is_empty t] returns [true] if [t] is empty, [false] otherwise. *)
```
talex5, line 316:
```suggestion
  (** [is_empty t] is [length t = 0]. *)
```

talex5 (approved):
Looks good. I added some suggestions to clarify the doc-strings.

PR #142, opened 2022-01-12
Add take_all

bikallem:
Draft: This build on top of pr #141 and adds `to_seq` function to convert Stream items to `Seq.t`.

Perhaps this should `to_list` instead and return a list?

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ This is an unreleased repository, as it's very much a work-in-progress.
   * [Example: Worker Pool](#example-worker-pool)
 * [Design Note: Determinism](#design-note-determinism)
 * [Examples](#examples)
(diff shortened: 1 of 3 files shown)


talex5 (changes requested):
Yes, this should return a list. Otherwise, you need to change the locking to lock on every access, not just the first one.

I suggest the name `take_all` for this behaviour of removing everything. `to_list` sounds like it shouldn't change the original.

bikallem:
Note: the loop function uses TRMC attribute so it allocates less and I believe more performant than a tail rec function. TRMC is available in the mainline OCaml 5.0.

bikallem:
closing this in favour of https://github.com/ocaml-multicore/eio/pull/146

PR #149, opened 2022-01-17
Remove dependency on ppxlib

talex5:
The only thing using this was Ctf, for the packet format, which never changes. Instead, just inline the generated code. Eventually this code will be replaced with eventring anyway.

This is also useful as a step towards OCaml 5.00, as ppxlib doesn't support that yet.


talex5:
With this change, these packages get removed from Eio's transitive dependencies:

- base (needed by parsexp)
- num
- ocaml-compiler-libs
- parsexp (needed by sexplib)
- ppx_cstruct
- ppx_derivers
- ppxlib
- sexplib (needed by ppx_cstruct)
- sexplib0

PR #150, opened 2022-01-17
EffectHandlers -> Effect

patricoferris:
I had this PR + the recently added inlining of `ppx_cstruct` locally for a compatible version of Eio with OCaml 5, not sure if Eio wants to change just yet and break compatibility with 4.12+domains.

There's also a bit of a hack around getting that one `linux` test to not be compiled on non-linux platforms which afaict is still not fixed upstream in dune?

### DIFF
--- a/lib_eio/cancel.ml
+++ b/lib_eio/cancel.ml
@@ -1,4 +1,4 @@
-open EffectHandlers
+open Effect
 
 exception Cancelled = Exn.Cancelled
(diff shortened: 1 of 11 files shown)


talex5:
I applied most of this in #151, but with a compatibility module.

Note that in that PR I dropped the test work-around, because it make the test executable part of all packages, not just `eio_linux`.

patricoferris:
Thanks! 

Ah I see, hopefully `dune.3` will have it fixed then :))

PR #161, opened 2022-01-27
Add more functions to Buf_read

talex5:
This adds some more common functions and syntax (matching the Angstrom API): `peek_char`, `skip`, `pair`, `map`, `bind`, `*>` and `<*`.


talex5:
This PR now also adds some fuzz testing using crowbar, fixes a few problems that found, improves error reporting, and adds a section about parsing to the README.

PR #165, opened 2022-02-02
Add secure_random device to stdenv

talex5:
Provide access to the system's secure-random feature. On Linux, this uses `getrandom` (which can block the whole domain, but only if run at early boot before the random system is initialised).

I thought about taking a switch argument and allowing the device to be closed, as some systems might want to open `/dev/urandom`. However, neither Linux not Luv need it so I left it out for now. Applications probably shouldn't be reading so much random data that keeping it open is useful.

Based on @patricoferris's branch, but I decided to use the regular flow interface as it is similar enough.

### DIFF
--- a/lib_eio/eio.ml
+++ b/lib_eio/eio.ml
@@ -36,6 +36,7 @@ module Stdenv = struct
     clock : Time.clock;
     fs : Dir.t;
     cwd : Dir.t;
+    secure_random : Flow.source;
   >
 
   let stdin  (t : <stdin  : #Flow.source; ..>) = t#stdin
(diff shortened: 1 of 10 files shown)


patricoferris:
Thanks, this looks good to me. Took it for a spin on top of the hacky UDP+resolver code in the STUN client, all working locally it would seem: https://github.com/patricoferris/ocaml-stun/commit/9a1e3d8698ee151a0db5eef4a554f58d0eaef204

talex5:
Thanks for checking :-) I'll merge once CI is finished.

PR #171, opened 2022-02-04
UDP interface

patricoferris:
This PR hopes to add UDP to Eio's networking interface, for now I'm still not sure about the actual interface so I'm opening this PR early (without uring support for now), in particular I have the following questions:

 - How to distinguish between "connectionful" and "connectionless" protocols, right now everything is just put into `Sockaddr.t` should they be separated ? I did something along those lines [here](https://github.com/patricoferris/eioio/blob/a285d2894cbd19c24ec317c0d0c36154d06289eb/lib_eio/net.ml#L116-L136).
 - Should Eio expose a "connected" UDP socket (like [luv's UDP.Connected](https://aantron.github.io/luv/luv/Luv/UDP/Connected/index.html) or [Tokio's `connect` function for UDP sockets](https://github.com/tokio-rs/tokio/blob/master/tokio/src/net/udp.rs)) which starts to feel more like an `Eio.Flow.two_way`?
 - At the moment the `endpoint` function always binds the host and port to the socket, maybe that shouldn't be the case? 
 - Should the `recv` pass through things like the sender address ?
 - How to handle/expose multicasting? 

Sorry that there's more questions than answers 😅

### DIFF
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -920,15 +920,15 @@ module Net : sig
     method virtual accept : sw:Switch.t -> <Flow.two_way; Flow.close> * Sockaddr.stream
   end
 
-  class virtual ['addr] datagram_socket : object
-    method virtual send : 'addr -> Cstruct.t -> unit
-    method virtual recv : Cstruct.t -> 'addr * int
+  class virtual datagram_socket : object
+    method virtual send : Sockaddr.datagram -> Cstruct.t -> unit
+    method virtual recv : Cstruct.t -> Sockaddr.datagram * int
   end
 
   class virtual t : object
     method virtual listen : reuse_addr:bool -> reuse_port:bool -> backlog:int -> sw:Switch.t -> Sockaddr.stream -> listening_socket
     method virtual connect : sw:Switch.t -> Sockaddr.stream -> <Flow.two_way; Flow.close>
-    method virtual datagram_socket : sw:Switch.t -> Sockaddr.datagram -> Sockaddr.datagram datagram_socket
+    method virtual datagram_socket : sw:Switch.t -> Sockaddr.datagram -> datagram_socket
   end
 
   (** {2 Out-bound Connections} *)
@@ -976,15 +976,15 @@ module Net : sig
 
   (** {2 Datagram Sockets} *)
 
-  val datagram_socket : sw:Switch.t -> #t -> Sockaddr.datagram -> Sockaddr.datagram datagram_socket
-  (** [datagram_socket ~sw t addr] creates a new, datagram socket that data can be sent to
+  val datagram_socket : sw:Switch.t -> #t -> Sockaddr.datagram -> datagram_socket
+  (** [datagram_socket ~sw t addr] creates a new datagram socket that data can be sent to
       and received from. The new socket will be closed when [sw] finishes. *)
 
-  val send : 'addr datagram_socket -> 'addr -> Cstruct.t -> unit
+  val send : datagram_socket -> Sockaddr.datagram -> Cstruct.t -> unit
   (** [send sock addr buf] sends the data in [buf] to the address [addr] using the 
-      the socket [sock]. *)
+      the datagram socket [sock]. *)
 
-  val recv : 'addr datagram_socket -> Cstruct.t -> 'addr * int
+  val recv : datagram_socket -> Cstruct.t -> Sockaddr.datagram * int
   (** [recv sock buf] receives data from the socket [sock] putting it in [buf]. The number of bytes received is 
       returned along with the sender address and port. If the [buf] is too small then excess bytes may be discarded
       depending on the type of the socket the message is received from. *)
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -895,11 +895,19 @@ module Net : sig
 
   (** Network addresses. *)
   module Sockaddr : sig
-    type t = [
+    type stream = [
       | `Unix of string
       | `Tcp of Ipaddr.v4v6 * int
+    ]
+    (** Socket addresses that we can build a {! Flow.two_way} for i.e. stream-oriented
+        protocols. *)
+
+    type datagram = [
       | `Udp of Ipaddr.v4v6 * int
     ]
+    (** Socket addresses that are message-oriented. *)
+
+    type t = [ stream | datagram ]
 
     val pp : Format.formatter -> t -> unit
   end
@@ -909,30 +917,30 @@ module Net : sig
   class virtual listening_socket : object
     inherit Generic.t
     method virtual close : unit
-    method virtual accept : sw:Switch.t -> <Flow.two_way; Flow.close> * Sockaddr.t
+    method virtual accept : sw:Switch.t -> <Flow.two_way; Flow.close> * Sockaddr.stream
   end
 
   class virtual endpoint : object
-    method virtual send : Sockaddr.t -> Cstruct.t -> unit
+    method virtual send : Sockaddr.datagram -> Cstruct.t -> unit
     method virtual recv : Cstruct.t -> (Ipaddr.v4v6 * int) option * int
   end
 
   class virtual t : object
-    method virtual listen : reuse_addr:bool -> reuse_port:bool -> backlog:int -> sw:Switch.t -> Sockaddr.t -> listening_socket
-    method virtual connect : sw:Switch.t -> Sockaddr.t -> <Flow.two_way; Flow.close>
-    method virtual endpoint : sw:Switch.t -> Sockaddr.t -> endpoint
+    method virtual listen : reuse_addr:bool -> reuse_port:bool -> backlog:int -> sw:Switch.t -> Sockaddr.stream -> listening_socket
+    method virtual connect : sw:Switch.t -> Sockaddr.stream -> <Flow.two_way; Flow.close>
+    method virtual endpoint : sw:Switch.t -> Sockaddr.datagram -> endpoint
   end
 
   (** {2 Out-bound Connections} *)
 
-  val connect : sw:Switch.t -> #t -> Sockaddr.t -> <Flow.two_way; Flow.close>
+  val connect : sw:Switch.t -> #t -> Sockaddr.stream -> <Flow.two_way; Flow.close>
   (** [connect ~sw t addr] is a new socket connected to remote address [addr].
 
       The new socket will be closed when [sw] finishes, unless closed manually first. *)
 
   (** {2 Incoming Connections} *)
 
-  val listen : ?reuse_addr:bool -> ?reuse_port:bool -> backlog:int -> sw:Switch.t -> #t -> Sockaddr.t -> listening_socket
+  val listen : ?reuse_addr:bool -> ?reuse_port:bool -> backlog:int -> sw:Switch.t -> #t -> Sockaddr.stream -> listening_socket
   (** [listen ~sw ~backlog t addr] is a new listening socket bound to local address [addr].
 
       The new socket will be closed when [sw] finishes, unless closed manually first.
@@ -947,7 +955,7 @@ module Net : sig
   val accept :
     sw:Switch.t ->
     #listening_socket ->
-    <Flow.two_way; Flow.close> * Sockaddr.t
+    <Flow.two_way; Flow.close> * Sockaddr.stream
   (** [accept ~sw socket] waits until a new connection is ready on [socket] and returns it.
 
       The new socket will be closed automatically when [sw] finishes, if not closed earlier.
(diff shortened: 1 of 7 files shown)


talex5:
Yeah, I'm not sure what the best approach for this is. Possibly we should just expose a fairly low-level socket interface and let people doing UDP sort it out themselves.

Sockets could inherit from `two_way` to provide a simple read/write interface, and use the extended methods (recv/send) to pass the remote address and flags for people who need that.

Might be worth talking to someone who works on mirage-tcpip too.

talex5:
@djs55 might have thoughts on this (VPNKit needs UDP), and @yomimono has also done lots of work in this area.

Mirage-tcpip defines this interface for UDP (https://github.com/mirage/mirage-tcpip/blob/main/src/core/udp.mli):
```ocaml
  type callback = src:ipaddr -> dst:ipaddr -> src_port:int -> Cstruct.t -> unit Lwt.t
  val listen : t -> port:int -> callback -> unit
  val unlisten : t -> port:int -> unit
  val input: t -> src:ipaddr -> dst:ipaddr -> Cstruct.t -> unit Lwt.t
  val write: ?src:ipaddr -> ?src_port:int -> ?ttl:int -> dst:ipaddr ->
    dst_port:int -> t -> Cstruct.t -> (unit, error) result Lwt.t
```

I think this is more designed for the case where you control the network device. Using it on Unix isn't very nice: doing a `write` call requires setting up and then tearing down a socket for each packet sent (if you're not also listening for incoming packets).

yomimono:
> How to distinguish between "connectionful" and "connectionless" protocols, right now everything is just put into Sockaddr.t should they be separated ? I did something along those lines [here](https://github.com/patricoferris/eioio/blob/a285d2894cbd19c24ec317c0d0c36154d06289eb/lib_eio/net.ml#L116-L136).

Personally I don't find much value in separating "connectionless" protocols out, since you have to ask the kernel to route packets to your application with something like `connect` anyway if you want to ever receive. In environments like MirageOS where you don't have to do this, it can be ergonomic to do something different, but it's very awkward otherwise :(

> Should the recv pass through things like the sender address ?

Strong yes.

djs55:
In the vpnkit world packets from the VM are [pattern-matched]( https://github.com/moby/vpnkit/blob/master/src/hostnet/slirp.ml#L516) like this:
```ocaml
    (* UDP to localhost *)
    | Ipv4 { src; dst; ihl; dnf; raw; ttl;
             payload = Udp { src = src_port; dst = dst_port; len;
                             payload = Payload payload; _ }; _ } ->
```
and relayed to the outside world using calls like [Uwt.Udp.bind](https://github.com/moby/vpnkit/blob/master/src/hostnet/hostnet_udp.ml#L234) (to fix a source address, although here I don't seem to care what it is)
```ocaml
                  Udp.bind ~description:(description datagram) (Ipaddr.(V4 V4.any), 0)
                  >>= fun server ->
```
and [Uwt.Udp.sendto](https://github.com/moby/vpnkit/blob/master/src/hostnet/hostnet_udp.ml#L256) to send the packet
```ocaml
        Udp.sendto flow.server datagram.intercept ~ttl datagram.payload >|= fun () ->
```
and then replies are received again with
```ocaml
        Udp.recvfrom server buf
        >>= fun (n, from) ->
```
The Uwt code is transitional... I've got a PR to switch to `luv`. I guess then I'll make another PR to switch to this :) 

Speaking entirely selfishly about the Mirage interfaces, it's a bit awkward for my use-case. I'd like to listen on all ports, so when TCP SYN packets arrive I have to sneakily call `listen` before inputting the packet to add an entry to the internal hashtable. I think in my ideal world I would pattern-match on incoming traffic and maintain a set of active TCP PCBs myself, and input packets to them without going via the normal interfaces. Anyway I'm not sure that was entirely relevant!

I agree with @yomimono that `recv` should include the sender address :)

patricoferris:
Thank you all for the quick feedback :)) 

I also managed to speak to @avsm about this too so I'll try and add those thoughts (feel free to correct me!). 

> Personally I don't find much value in separating "connectionless" protocols out, since you have to ask the kernel to route packets to your application with something like connect anyway if you want to ever receive.

To a certain extent I agree, I think there are quite a few protocols that work over UDP and TCP, although the application logic would be different. For example, when experimenting with a STUN client ([which works over both UDP and TCP (DTLS and TLS)](https://datatracker.ietf.org/doc/html/rfc8489#section-6.2)) this distinction made it difficult [to implement](https://github.com/patricoferris/ocaml-stun/blob/main/src/stun-lwt-unix/conduit_udp.ml) (or at least confusing for someone with less networking experience)... but maybe I was just missing something?

The problem at the moment I guess is that if `Udp` is bundled into `Sockaddr.t` then does `connect` have to raise an exception for `Udp` addresses? It can't really provide a _flow_ for reasons outlined below.

> Sockets could inherit from two_way to provide a simple read/write interface, and use the extended methods (recv/send) to pass the remote address and flags for people who need that.

Having this simplification from a user's perspective would be very nice. But, and I'm paraphrasing my own interpretation of @avsm's comments, there should be a distinction in that writing data with a UDP socket is not a _flow_. There are some semantics associated with that because they're byte streams? I don't think we can directly use the flow interface with UDP because they're different (a UDP read could be partial, a TCP read can raise `End_of_file`)... but maybe that's not what you had in mind @talex5 ? Maybe it was more that an `Eio.Socket` is the more general case and the user should be aware of the differences when programming with it i.e. if your socket is TCP then a read could raise end of file but it is UDP then it is really just a receive that could be partial? There would also be a low-level socket API with a `connect` that returns an `Eio.Socket.t` rather than a flow?

(Sender addresses are a go then 👍)

talex5:
> The problem at the moment I guess is that if Udp is bundled into Sockaddr.t then does connect have to raise an exception for Udp addresses? It can't really provide a flow for reasons outlined below.

`Sockaddr.t` is a polymorphic variant, so just use whatever subset makes sense. Might need some minor changes to keep this usable. e.g. `Sockaddr.pp` should probably be `val pp : Format.formatter -> [< t] -> unit`.

> There are some semantics associated with [inheriting from two_way] because they're byte streams?

Yes, OK, probably best to keep them separate, even if POSIX doesn't.

lib_eio/eio.mli
```diff
@@ -714,24 +714,33 @@ module Net : sig
...
+        protocols. *)
+
+    type datagram = [
+      | `Udp of Ipaddr.v4v6 * int
+    ]
+    (** Socket addresses that are message-oriented. *)
+
+    type t = [ stream | datagram ]
 
     val pp : Format.formatter -> t -> unit
```
talex5, line 731:
```suggestion
    val pp : Format.formatter -> [< t] -> unit
```
Otherwise printing an address forces it to have the full type.

lib_eio/eio.mli
```diff
@@ -740,18 +749,32 @@ module Net : sig
...
+    (sw:Switch.t -> <Flow.two_way; Flow.close> -> Sockaddr.stream -> unit) ->
     unit
-  (** [accept socket fn] waits for a new connection to [socket] and then runs [fn ~sw flow client_addr] in a new fibre,
+  (** [accept_sub socket fn] waits for a new connection to [socket] and then runs [fn ~sw flow client_addr] in a new fibre,
       using {!Fibre.fork_on_accept}.
       [flow] will be closed automatically when the sub-switch is finished, if not already closed by then. *)
 
+  class virtual endpoint : object
+    method virtual send : Sockaddr.datagram -> Cstruct.t -> unit
+    method virtual recv : Cstruct.t -> (Ipaddr.v4v6 * int) option * int
```
talex5, line 760:
```suggestion
    method virtual recv : Cstruct.t -> 'addr * int
```
Or `Sockaddr.t` if you don't want a type parameter. Shouldn't be an option.

lib_eio_luv/eio_luv.ml
```diff
@@ -475,6 +475,43 @@ let luv_ip_addr_to_eio addr =
...
+        Fibre_context.set_cancel_fn k.fibre (fun ex ->
+            Luv.UDP.recv_stop (Handle.get "recv_into:cancel" sock) |> or_raise;
+            enqueue_failed_thread t k ex
+          );
+        Luv.UDP.recv_start (Handle.get "recv_start" sock) ~allocate:(fun _ -> buf) (fun r ->
+            Luv.UDP.recv_stop (Handle.get "recv_stop" sock) |> or_raise;
+            if Fibre_context.clear_cancel_fn k.fibre then enqueue_thread t k r
+          )
+      ) in
+    match r with
```
talex5, line 492: The [Luv docs](https://aantron.github.io/luv/luv/Luv/UDP/index.html#val-recv_start) say:
> None usually indicates EAGAIN in libuv; libuv still calls the callback, in order to give the C user a chance to deallocate the data buffer. Since this is not usually an issue in OCaml, it is usually safe to simply ignore calls to the callback with sender address None.

tests/test_network.md
```diff
@@ -148,6 +148,36 @@ Calling accept when the switch is already off:
...
+  let e1 = `Udp (Eio.Net.Ipaddr.V4.loopback, 8081) in
+  let e2 = `Udp (Eio.Net.Ipaddr.V4.loopback, 8082) in
+  Fibre.both
+    (fun () ->
+      let e = Eio.Net.endpoint ~sw net e2 in
+      let buf = Cstruct.create 20 in
+      traceln "Waiting to receive data on %a" Eio.Net.Sockaddr.pp e2;
+      let addr, recv = Eio.Net.recv e buf in
+      traceln "Received message from %a: %s"
+      (Fmt.option pp_addr) addr
```
talex5, line 168:
```suggestion
      Eio.Net.Sockaddr.pp addr
```

lib_eio_luv/eio_luv.ml
patricoferris: Yep I saw that. What's the best way to handle `EAGAIN`? Just recursively call `recv` again ?

talex5: Just don't do anything in that case I think, not even `recv_stop`.

lib_eio_linux/eio_linux.ml
```diff
@@ -871,6 +927,18 @@ let net = object
     let sock = FD.of_unix ~sw ~seekable:false ~close_unix:true sock_unix in
     Low_level.connect sock addr;
     (flow sock :> <Eio.Flow.two_way; Eio.Flow.close>)
+
+  method datagram_socket ~sw = function
+    | `Udp (host, port) ->
+      let host = Eio_unix.Ipaddr.to_unix host in
+      let addr = Unix.ADDR_INET (host, port) in
+      let sock_unix = Unix.socket Unix.PF_INET Unix.SOCK_DGRAM 0 in
+      (* TODO: Should expose these as arguments. *)
```
patricoferris, line 936: Do you think this should be exposed in the function type like for listening connections?

lib_eio/eio.mli
```diff
@@ -740,18 +749,33 @@ module Net : sig
...
+  (** [recv sock buf] receives data from the socket [sock] putting it in [buf]. The number of bytes received is 
+      returned along with the sender address and port. If the [buf] is too small then excess bytes may be discarded
+      depending on the type of the socket the message is received from. *)
+
   class virtual t : object
-    method virtual listen : reuse_addr:bool -> reuse_port:bool -> backlog:int -> sw:Switch.t -> Sockaddr.t -> listening_socket
-    method virtual connect : sw:Switch.t -> Sockaddr.t -> <Flow.two_way; Flow.close>
+    method virtual listen : reuse_addr:bool -> reuse_port:bool -> backlog:int -> sw:Switch.t -> Sockaddr.stream -> listening_socket
+    method virtual connect : sw:Switch.t -> Sockaddr.stream -> <Flow.two_way; Flow.close>
+    method virtual datagram_socket : sw:Switch.t -> Sockaddr.datagram -> Sockaddr.datagram datagram_socket
```
talex5, line 775: When I suggested the type parameter, I was thinking of something like this:
```suggestion
    method virtual datagram_socket : 'a. sw:Switch.t -> ([< Sockaddr.datagram] as 'a) -> 'a datagram_socket
```
If we only want UDP datagrams, the type could be simplified though. According to my Linux's unix(7) man-page you can use Unix datagrams for reliable datagrams that preserve message boundaries, though I don't think I've ever used this.

lib_eio_linux/eio_linux.ml
talex5: I'm happy to keep them internal until someone complains.

talex5 (approved):
Looks good. The `datagram_socket` type parameter isn't really being used here. If we only support UDP datagrams, then feel free to get rid of it for now.

It might be useful if we add support for Unix datagrams (so e.g. you know statically that reading from a Unix socket will give you a Unix source address, and that you can only send to UDP addresses over a UDP socket).

patricoferris:
Thanks for the review. The rebase was a bit painful because of the `eio.mli` changes. 

I simplified the `datagram_socket` to not take the parameter which I think simplifies the API quite a bit, hopefully that was what you meant :))

talex5 (approved):
I pushed a commit with a couple of minor changes:

- Close the socket if `bind` fails.
- In the tests, ensure listening socket is ready before sending.

If that looks OK to you, then this is ready to merge.

patricoferris:
Good changes, thank you ! :))

talex5:
OK, it's in - thanks!

PR #197, opened 2022-02-23
Add more documentation

talex5:
In particular, explain more low-level details about how cancellation works.

Based on feedback from @gasche, Rahul and others on https://discuss.ocaml.org.

### DIFF
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -32,6 +32,8 @@ module Switch : sig
       Calling {!fail} cancels all fibers attached to the switch and, once they
       have exited, reports the error.
 
+      Note: this concept is known as a "nursery" or "bundle" in some other systems.
+
       Example:
       {[
          Switch.run (fun sw ->
@@ -45,6 +47,8 @@ module Switch : sig
   type t
   (** A switch contains a group of fibers and other resources (such as open file handles). *)
 
+  (** {2 Switch creation} *)
+
   val run : (t -> 'a) -> 'a
   (** [run fn] runs [fn] with a fresh switch (initially on).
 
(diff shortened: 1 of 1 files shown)


lib_eio/eio.mli
```diff
@@ -87,18 +103,24 @@ module Switch : sig
...
   val null_hook : hook
   (** A dummy hook. Removing it does nothing. *)
 
   val on_release_cancellable : t -> (unit -> unit) -> hook
-  (** Like [on_release], but the handler can be removed later. *)
+  (** Like [on_release], but the handler can be removed later.
+
+      For example, opening a file will call [on_release_cancellable] to ensure the file is closed later.
+      However, if the file is manually closed before that, it will use {!remove_hook} to remove the hook,
+      which is no longer needed. *)
```
gasche, line 116: Note: I would find it more natural to have an API like:

```
module Hook : sig
  type t

  val make : (unit -> unit) -> t
  (* [null_hook] is equivalent to [make ignore] *)

  val remove : t -> unit
end

val on_release_cancellable : t -> hook -> unit
(* or just on_release_hook ? *)
```

(There could also be a function to run a hook, to `val run : t -> unit`, but I understand that this is not necessary for internal usage of hooks, only if users want to use them for other things.)

talex5: I did have it as separate module originally. I got rid of it because it seems distracting in the generated docs (it was the first module you saw when looking at Eio's docs, yet most users won't care about it).

A `make` function won't work, because internally a hook is a `Lwt_dllist.node` in some pre-existing linked list.

gasche: You could still have `make`, I think? `make` would create a one-element circular dllist, and `on_release_cancellable` would splice this dllist into another dllist. But this looses the property that all hooks are "reachable" from some fiber context. Is this property important? (This separation also makes it possible for users to define their hooks for their own purposes, not to attach them to a context -- if they also have a `run` function. But I don't know if this has any use.)

talex5: Lwt_dllist doesn't provide a way to do what you want (https://github.com/mirage/lwt-dllist/blob/master/src/lwt_dllist.mli). Adding a node to a list could be useful in some cases, but here it would just add extra complications (what if I try to add a hook that's already in another list?).

odoc just shows things in the same order as the mli file as far as I know. It wouldn't be a huge problem having it as a separate module, but it just doesn't seem important enough for that.

PR #204, opened 2022-02-28
Disable linux mdx test on non-linux platforms

patricoferris:
[`enabled_if` for `mdx` stanzas was added in dune `2.9`](https://github.com/ocaml/dune/blob/main/CHANGES.md#290-29062021) so I think it's okay to do this, otherwise it won't work on macOS (or windows) for example.


talex5:
Thanks!

PR #207, opened 2022-03-16
Add readdir feature

### LINKED ISSUE #206: API request: readdir
Given `Eio.Dir.t` it should be possible to list the contents of the directory

patricoferris:
This PR starts the implementation of a `readdir` feature for Eio. I opened it mainly to have a conversation about the API.
 
  - What should be returned when you read the contents of a directory? This PR only returns the names of the contents, libuv does expose the `kind` of the entries which probably would be nice to add?
  - What should the name of the function be, here it is `read_dir` but I have seen `readdir` to follow the convention used by the C function, `contents` used by `Bos` etc.
 
At the moment I don't see a way to provide an asynchronous version with uring? I think with the 5.17 kernel there will be a `getdents64` uring submission function which could then be used.

Will fix #206

### DIFF
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -1186,7 +1186,7 @@ module Dir : sig
       it automatically when [fn] returns (if it hasn't already been closed by then). *)
 
   val read_dir : #t -> path -> string list
-  (** [read_dir t path] reads directory entries for [t/path].*)
+  (** [read_dir t path] reads directory entries for [t/path]. The entries are sorted using {! String.compare}.*)
 end
 
 (** The standard environment of a process. *)
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -1118,6 +1118,7 @@ module Dir : sig
       path -> <rw; Flow.close>
     method virtual mkdir : perm:Unix_perm.t -> path -> unit
     method virtual open_dir : sw:Switch.t -> path -> t_with_close
+    method virtual read_dir : path -> path list
   end
   and virtual t_with_close : object
     inherit t
@@ -1183,6 +1184,9 @@ module Dir : sig
   val with_open_dir : #t -> path -> (<t; Flow.close> -> 'a) -> 'a
   (** [with_open_dir] is like [open_dir], but calls [fn dir] with the new directory and closes
       it automatically when [fn] returns (if it hasn't already been closed by then). *)
+
+  val read_dir : #t -> path -> string list
+  (** [read_dir t path] reads directory entries for [t/path].*)
 end
 
 (** The standard environment of a process. *)
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -654,8 +654,15 @@ module Low_level = struct
   type _ Effect.t += Free : Uring.Region.chunk -> unit Effect.t
   let free_fixed buf = Effect.perform (Free buf)
 
-  type _ Effect.t += Unblock : (unit -> 'a) -> 'a Effect.t
-  let unblock fn = Effect.perform (Unblock fn)
+  let unblock fn =
+    let f fiber enqueue =
+      match Fiber_context.get_error fiber with
+      | Some err -> enqueue (Error err)
+      | None ->
+        let _t : Thread.t = Thread.create (fun () -> enqueue (try Ok (fn ()) with exn -> Error exn)) () in
+        ()
+    in
+    Effect.perform (Eio.Private.Effects.Suspend f)
 
   let splice src ~dst ~len =
     let res = enter (enqueue_splice ~src ~dst ~len) in
@@ -1089,7 +1096,7 @@ class dir fd = object
     Switch.run (fun sw ->
       (* Using openat2 to preserve the permission checking, but not using the returned
          file descriptor. Uring doesn't support opendir. *)
-      let dir =
+      let _dir =
         wrap_errors path @@ fun () ->
         Low_level.openat2 ~sw ~seekable:false ?dir:fd path
           ~access:`R
@@ -1097,7 +1104,6 @@ class dir fd = object
           ~perm:0
           ~resolve:Uring.Resolve.beneath
       in
-      Fun.protect ~finally:(fun () -> FD.close dir) @@ fun () ->
       wrap_errors path @@ fun () ->
       let handle = Low_level.unblock (fun () -> Unix.opendir path) in
       Low_level.unblock (fun () -> read_all [] handle)
@@ -1305,17 +1311,6 @@ let rec run ?(queue_depth=64) ?n_blocks ?(block_size=4096) ?polling_timeout ?fal
           | Low_level.Free buf -> Some (fun k ->
               Low_level.free_buf st buf;
               continue k ()
-            )
-          | Low_level.Unblock f -> Some (fun k ->
-              let k = { Suspended.k; fiber } in
-              let _t = Thread.create (fun () ->
-                let v = ref None in
-                run ~queue_depth ~n_blocks ~block_size ?polling_timeout ?fallback:None (fun _ -> v := Some (try Ok (f ()) with exn -> Error exn));
-                match Option.get !v with
-                  | Ok v -> enqueue_thread st k v
-                  | Error exn -> enqueue_failed_thread st k exn
-              ) () in
-              schedule st
             )
           | _ -> None
       }
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -654,16 +654,6 @@ module Low_level = struct
   type _ Effect.t += Free : Uring.Region.chunk -> unit Effect.t
   let free_fixed buf = Effect.perform (Free buf)
 
-  let unblock fn =
-    let f fiber enqueue =
-      match Fiber_context.get_error fiber with
-      | Some err -> enqueue (Error err)
-      | None ->
-        let _t : Thread.t = Thread.create (fun () -> enqueue (try Ok (fn ()) with exn -> Error exn)) () in
-        ()
-    in
-    Effect.perform (Eio.Private.Effects.Suspend f)
-
   let splice src ~dst ~len =
     let res = enter (enqueue_splice ~src ~dst ~len) in
     Log.debug (fun l -> l "splice returned");
@@ -797,7 +787,7 @@ module Low_level = struct
           ~resolve:Uring.Resolve.beneath
 
     let getdents dir =
-      unblock (fun () -> eio_getdents (FD.get "getdents" dir))
+      Eio_unix.run_in_systhread (fun () -> eio_getdents (FD.get "getdents" dir))
 end
 
 external eio_eventfd : int -> Unix.file_descr = "caml_eio_eventfd"
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -654,6 +654,9 @@ module Low_level = struct
   type _ Effect.t += Free : Uring.Region.chunk -> unit Effect.t
   let free_fixed buf = Effect.perform (Free buf)
 
+  type _ Effect.t += Unblock : (unit -> 'a) -> 'a Effect.t
+  let unblock fn = Effect.perform (Unblock fn)
+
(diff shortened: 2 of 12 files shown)


talex5:
> What should be returned when you read the contents of a directory? This PR only returns the names of the contents, libuv does expose the kind of the entries which probably would be nice to add?

Yes, `getdents` on Linux gives you the kind, which saves having to stat each entry in some cases. Though it can also return `DT_UNKNOWN` if the FS doesn't support it.

> At the moment I don't see a way to provide an asynchronous version with uring?

We can use systhreads for things that uring doesn't handle. We'll probably need to do that anyway for stuff like `gethostbyname`.

hexagonrecursion:
1. Why is it `read_dir : Eio.Dir.t -> path -> string list`? I expected `read_dir : Eio.Dir.t -> string list` - "given Eio.Dir.t give me its contents"
1. You should consider how the API will be used. If you give me a list of strings, I will usually have to check each one to see whether it is a directory or a file because the API for working with directories and files is different. How about a list of the following?
    ```ocaml
    type dirent = [ 
    | `File of path
    | `Dir of Eio.Dir.t
    ]
    ```
1. What if a directory contains 10**12 files ([see](https://stackoverflow.com/questions/466521/how-many-files-can-i-put-in-a-directory)) and I don't want a list that big in memory? There should be an API to read the contents of the directory asynchronously just like we are able to read a file asynchronously

patricoferris:
Thanks both for the comments :)) 

> Though it can also return DT_UNKNOWN if the FS doesn't support it.

Yep, this is also exposed in `luv` https://aantron.github.io/luv/luv/Luv/File/Dirent/Kind/index.html#type-t.UNKNOWN 

> We can use systhreads for things that uring doesn't handle. We'll probably need to do that anyway for stuff like gethostbyname.

Forgot this was an option, I'll have a go! 

> Why is it read_dir : Eio.Dir.t -> path -> string list? I expected read_dir : Eio.Dir.t -> string list - "given Eio.Dir.t give me its contents"

My understanding is that an `Eio.Dir.t` is not a directory, but instead [a file-system access capability](https://github.com/ocaml-multicore/eio#filesystem-access). Maybe `Dir` is a bit misleading, perhaps `Eio.Fs` is better, but maybe @talex5 has good reason for the naming scheme? In the case of `cwd` it ensures we can't access things like `../../some/dir` whereas the full-system access capability `fs` would allow such a thing but note that it doesn't actually impact the path, it just checks what we're trying to access with the `path`.

There's also https://github.com/ocaml-multicore/eio#design-note-capabilities for some more context on this design. 

> How about a list of the following?
> ```ocaml
> type dirent = [ 
> | `File of path
> | `Dir of Eio.Dir.t
> ]
> ```

Yep returning the kinds makes sense. Maybe following the layout of `luv` so:

```ocaml
type dirent = {
  kind : kind;
  name : string;
}
and kind = [ `Unknown | `Dir | `File ... ]
```

`` `Dir of Eio.Dir.t`` isn't quite right given the explanation of `Eio.Dir.t` above.

> What if a directory contains 10**12 files ([see](https://stackoverflow.com/questions/466521/how-many-files-can-i-put-in-a-directory)) and I don't want a list that big in memory?

Great point, I'll see what's available in the backend APIs to make this possible :))

hexagonrecursion:
Question: given:
```
path/to/foo/bar.txt

```

patricoferris:
Currently it would be`bar.txt` which I think is inline with [readdir(3)](https://man7.org/linux/man-pages/man3/readdir.3.html), [Sys.readdir](https://v3.ocaml.org/p/ocaml-base-compiler/4.13.1/doc/Stdlib/Sys/index.html#val-readdir) and [Luv.File.readdir](https://aantron.github.io/luv/luv/Luv/File/index.html#val-readdir) so if we want paths to the directory entries, it probably shouldn't be called `read_dir`. Reconstructing the path (once we have a better `path` type) wouldn't be too hard and maybe we could provide a convenience function built on top of `read_dir` for this? Alternatively, I suppose, it won't be too hard to provide a `Path.filename` function too.

talex5:
>> Why is it read_dir : Eio.Dir.t -> path -> string list? I expected read_dir : Eio.Dir.t -> string list - "given Eio.Dir.t give me its contents"
>
> My understanding is that an Eio.Dir.t is not a directory, but instead [a file-system access capability](https://github.com/ocaml-multicore/eio#filesystem-access). Maybe Dir is a bit misleading, perhaps Eio.Fs is better, but maybe @talex5 has good reason for the naming scheme?

I just copied Rust's name for it (https://docs.rs/cap-std/0.22.1/cap_std/fs/struct.Dir.html). They have e.g. 

```rust
impl Dir {
  fn read_dir<P: AsRef<Path>>(&self, path: P) -> Result<ReadDir>
```

The Eio API also matches the POSIX one (e.g. `openat2`, where you pass a base FD and a path relative to that). The reason for having both is mostly to do with symlinks: you can't access the parent of the base FD, but you can follow symlinks within the subtree.

So the idea is that a `Dir.t` *is* a directory, and doesn't allow access to its parent. However, `fs` is an exception to this rule because it represents the current directory but also allows access to everything (via `..`, symlinks, or an absolute path).

patricoferris:
I pushed a commit which tries to use systhreads now they're fixed in OCaml trunk. It's meant to generate a discussion about how best to use them to provide non-blocking, asynchronous versions of unsupported functions in Uring.

Right now the implementation uses a new effect, `Unblock` which takes a function to "unblock". It runs this function in a newly created systhread and the result is then enqueued to the run queue. I don't know if this is the general idea that you had in mind @talex5 ? I had a look at Lwt and Mwt which are more complicated with pools of threads and workers, but I thought thanks to effects we might be able to simplify that a fair bit.

Also the `readdir` function reuses `openat2` to do the necessary filesystem containment checking, but then opens a directory handle anyway which probably isn't particularly efficient.

lib_eio_linux/eio_linux.ml
```diff
@@ -1065,6 +1068,29 @@ class dir fd = object
...
+         file descriptor. Uring doesn't support opendir. *)
+      let dir =
+        wrap_errors path @@ fun () ->
+        Low_level.openat2 ~sw ~seekable:false ?dir:fd path
+          ~access:`R
+          ~flags:Uring.Open_flags.(cloexec + path + directory)
+          ~perm:0
+          ~resolve:Uring.Resolve.beneath
+      in
+      Fun.protect ~finally:(fun () -> FD.close dir) @@ fun () ->
```
talex5, line 1088: This isn't needed. `dir` already gets closed when `sw` finishes.

lib_eio_linux/eio_linux.ml
```diff
@@ -1065,6 +1068,29 @@ class dir fd = object
...
+      with End_of_file -> List.rev acc |> List.filter (function ".." | "." -> false | _ -> true)
+    in
+    Switch.run (fun sw ->
+      (* Using openat2 to preserve the permission checking, but not using the returned
+         file descriptor. Uring doesn't support opendir. *)
+      let dir =
+        wrap_errors path @@ fun () ->
+        Low_level.openat2 ~sw ~seekable:false ?dir:fd path
+          ~access:`R
+          ~flags:Uring.Open_flags.(cloexec + path + directory)
```
talex5, line 1084:
```suggestion
          ~flags:Uring.Open_flags.(cloexec + directory)
```
Using `path` is OK for the code as it is, but to avoid opening it twice we'll need to remove this.

tests/test_fs.md
```diff
@@ -31,6 +31,11 @@ let try_mkdir dir path =
   | () -> traceln "mkdir %S -> ok" path
   | exception ex -> traceln "mkdir %S -> %a" path Fmt.exn ex
 
+let try_read_dir dir path =
+  match Eio.Dir.read_dir dir path with
+  | names -> traceln "read_dir [ %a ] -> ok" Fmt.(list ~sep:Fmt.comma string) names
+  | exception ex -> traceln "read_dir %a" Fmt.exn ex
```
talex5, line 37:
```suggestion
  | names -> traceln "read_dir %S -> %a" path Fmt.Dump.(list string) names
  | exception ex -> traceln "read_dir %S -> %a" path Fmt.exn ex
```
Just for consistency with the others. `mkdir` only prints `ok` because there's nothing else to print.

lib_eio/eio.mli
```diff
@@ -1183,6 +1184,9 @@ module Dir : sig
   val with_open_dir : #t -> path -> (<t; Flow.close> -> 'a) -> 'a
   (** [with_open_dir] is like [open_dir], but calls [fn dir] with the new directory and closes
       it automatically when [fn] returns (if it hasn't already been closed by then). *)
+
+  val read_dir : #t -> path -> string list
+  (** [read_dir t path] reads directory entries for [t/path].*)
```
talex5, line 1189: (and then mention sorting here)

talex5 (commented):
Thanks, it seems to work!

> Also the readdir function reuses openat2 to do the necessary filesystem containment checking, but then opens a directory handle anyway which probably isn't particularly efficient.

Also, there's a possible race if a symlink appears between the check and the open. The libc and OCaml APIs for reading directories aren't very nice, and if we want to return types too then I guess we should just call `getdents64` directly and avoid the whole mess.

> What if a directory contains 10**12 files ([see](https://stackoverflow.com/questions/466521/how-many-files-can-i-put-in-a-directory)) and I don't want a list that big in memory? There should be an API to read the contents of the directory asynchronously just like we are able to read a file asynchronously

Yes, probably makes sense for the backend API to provide open_dir, etc, and for `read_dir` to be a wrapper around that.

patricoferris:
> Yes, probably makes sense for the backend API to provide open_dir, etc, and for read_dir to be a wrapper around that.

Should `Eio.Dir.read_dir` try to read all directory entries or just some amount of them and the user can call it multiple times ? Or should that be a completely separate function ?

lib_eio_linux/eio_stubs.c
```diff
@@ -40,3 +45,37 @@ CAMLprim value caml_eio_getrandom(value v_ba, value v_off, value v_len) {
...
+    unsigned short d_reclen; /* Size of this dirent */
+    unsigned char  d_type;   /* File type */
+    char           d_name[]; /* Filename (null-terminated) */
+};
+
+CAMLprim value caml_eio_getdents(value v_fd) {
+  CAMLparam1(v_fd);
+  CAMLlocal1(result);
+  char buf[DIRENT_BUF_SIZE];
+  struct linux_dirent64 *d;
```
talex5, line 61: Can't we just use the version from `dirent.h`?
```suggestion
  struct dirent64 *d;
```

lib_eio_linux/eio_stubs.c
```diff
@@ -40,3 +45,37 @@ CAMLprim value caml_eio_getrandom(value v_ba, value v_off, value v_len) {
...
+  caml_enter_blocking_section();
+  nread = syscall(SYS_getdents64, Int_val(v_fd), buf, DIRENT_BUF_SIZE);
+  caml_leave_blocking_section();
+  if (nread == -1) uerror("getdents", Nothing);
+
+  result = Val_int(0); /* The empty list */
+
+  for (pos = 0; pos < nread;) {
+    d = (struct linux_dirent64 *) (buf + pos);
+    CAMLlocal1 (r);
```
talex5, line 72: Not sure, but I suspect doing `CAMLlocal1` inside a loop is a bad idea (it's adding `r` to `caml__roots_`).

talex5 (commented):
> Should Eio.Dir.read_dir try to read all directory entries or just some amount of them and the user can call it multiple times ? Or should that be a completely separate function ?

Ideally, I think there should be two functions. Maybe the backend would provide something like:

```ocaml
method with_dir_entries : string -> (dir_entry Seq.t -> 'a) -> 'a
```

But I'm happy to merge it with just this read-everything function for now. Almost everyone will just use that.

lib_eio_linux/eio_stubs.c
patricoferris: Oops yes! Thanks

lib_eio_linux/eio_stubs.c
patricoferris: Ah, I didn't see this when looking for linux examples using `getdents64`, thank you for pointing it out :))

talex5 (approved):

PR #209, opened 2022-04-02
Use Fun.protect

favonia:
It seems the codebase is re-implementing `Fun.protect` at various places. It has the immediate advantage of preserving backtraces. The only danger of using `Fun.protect` is that, if the finalizer itself is raising an exception, that exception would be wrapped with `Fun.Finally_raised` and might be missed by the handlers which wish to capture it. I am not familiar with the codebase enough to assert that this would not happen, but these changes seem fine.

### DIFF
--- a/lib_eio/cancel.ml
+++ b/lib_eio/cancel.ml
@@ -112,9 +112,7 @@ let with_cc ~ctx:fiber ~parent ~protected fn =
   let deactivate = activate t ~parent in
   move_fiber_to t fiber;
   let cleanup () = move_fiber_to parent fiber; deactivate () in
-  match fn t with
-  | x            -> cleanup (); x
-  | exception ex -> cleanup (); raise ex
+  Fun.protect ~finally:cleanup @@ fun () -> fn t
 
 let protect fn =
   let ctx = Effect.perform Get_context in
--- a/lib_eio/stream.ml
+++ b/lib_eio/stream.ml
@@ -15,9 +15,7 @@ type 'a t = {
 
 let with_mutex t f =
   Mutex.lock t.mutex;
-  match f () with
-  | x -> Mutex.unlock t.mutex; x
-  | exception ex -> Mutex.unlock t.mutex; raise ex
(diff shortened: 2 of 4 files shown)


favonia:
The CI failure seems to be all about `ocaml-uring/uring.opam` being missing.

talex5:
> It seems the codebase is re-implementing Fun.protect at various places. It has the immediate advantage of preserving backtraces.

I've had the opposite experience. Fun.protect tends to give backtraces that show that Fun.protect was called, rather than the function you're interested in (possibly because it's often a tail-call). OCaml seems to preserve backtraces fine if you just re-raise.

e.g.

```ocaml
let bar () =
  raise Not_found

let foo () =
  try bar ()
  with ex ->
    print_endline "re-throw";
    raise ex

let () =
  foo ()
```
gives
```
re-throw                          
Fatal error: exception Not_found
Raised at Dune__exe__Main.bar in file "main.ml" (inlined), line 2, characters 2-17
Raised by primitive operation at Dune__exe__Main.foo in file "main.ml", line 5, characters 6-12
Re-raised at Dune__exe__Main.foo in file "main.ml", line 8, characters 4-12
Called from Dune__exe__Main in file "main.ml", line 11, characters 2-8
```

But:

```ocaml
let bar () =
  raise Not_found

let foo () =
  Fun.protect bar ~finally:(fun () -> print_endline "protect")

let () =
  foo ()
```
gives
```
protect                           
Fatal error: exception Not_found
Raised at Dune__exe__Main.bar in file "main.ml", line 2, characters 2-17
Called from Stdlib__Fun.protect in file "fun.ml", line 33, characters 8-15
Re-raised at Stdlib__Fun.protect in file "fun.ml", line 38, characters 6-52
Called from Dune__exe__Main in file "main.ml", line 8, characters 2-8
```
Here, we've lost the information that `bar` was called from `foo`.

favonia:
Ah, thank you. I guess there's no point in pursuing this PR, then.

Issue #211, opened 2022-04-22
Check EINTR is handled correctly


talex5:
System calls need to be retried if they return EINTR (unless this was due to them being cancelled; you don't always get `ECANCELED` in that case).

There is some handling already, e.g. here's eio_linux handling EINTR for rw ops:

https://github.com/ocaml-multicore/eio/blob/85841dc0b779a920bc4b1daca9172456e33988ee/lib_eio_linux/eio_linux.ml#L519-L527

Check if we need this in more places, and also check the eio_luv backend.

@hannesm notes in https://github.com/mirage/mirage-crypto/pull/155#issuecomment-1106412128 that `getrandom` needs this too.

haesbaert:
We could also just enforce `SA_RESTART`.
Not sure how you feel about eio calling `sigaction` though.

talex5:
Reading signal(7), it seems hard to avoid EINTR in all cases. In particular, it seems that `SIGCONT` can generate EINTR.

It's also a bit unclear to me what happens with signals and io_uring operations. I would expect that the `io_uring_enter` call would return EINTR rather than the individual operations, so maybe the retry logic above isn't even needed. That code has been there since the very first commit by @avsm; I don't know if it was needed for something.

talex5:
As far as I know this is working correctly everywhere, so I'll close this. We can reopen if we find any remaining places we missed.

PR #212, opened 2022-04-22
Handle EINTR error when calling `getrandom` system call

bikallem:
This PR handles EINTR error return code when `getrandom` is called via eio_linux or eio_luv packages.

Fixes part of https://github.com/ocaml-multicore/eio/issues/211
Required by https://github.com/mirage/mirage-crypto/pull/155

### DIFF
--- a/lib_eio_linux/eio_stubs.c
+++ b/lib_eio_linux/eio_stubs.c
@@ -2,6 +2,7 @@
 #include <sys/types.h>
 #include <sys/eventfd.h>
 #include <sys/random.h>
+#include <errno.h>
 
 #include <caml/mlvalues.h>
 #include <caml/memory.h>
@@ -33,10 +34,16 @@ CAMLprim value caml_eio_mkdirat(value v_fd, value v_path, value v_perm) {
 CAMLprim value caml_eio_getrandom(value v_ba, value v_off, value v_len) {
   CAMLparam1(v_ba);
   ssize_t ret;
-  void *buf = Caml_ba_data_val(v_ba) + Long_val(v_off);
-  caml_enter_blocking_section();
-  ret = getrandom(buf, Long_val(v_len), 0);
-  caml_leave_blocking_section();
-  if (ret == -1) uerror("getrandom", Nothing);
+  size_t off = (size_t)Long_val(v_off);
+  size_t len = (size_t)Long_val(v_len);
+  while (off < len) {
+    void *buf = Caml_ba_data_val(v_ba) + off;
+    caml_enter_blocking_section();
+    ret = getrandom(buf, len - off, 0);
+    caml_leave_blocking_section();
+    if (ret < 0 && errno == EINTR) continue;
+    else if (ret < 0) uerror("getrandom", Nothing);
+    off += ret;
+  }
   CAMLreturn(Val_long(ret));
 }
--- a/lib_eio_linux/eio_stubs.c
+++ b/lib_eio_linux/eio_stubs.c
@@ -34,16 +34,14 @@ CAMLprim value caml_eio_mkdirat(value v_fd, value v_path, value v_perm) {
 CAMLprim value caml_eio_getrandom(value v_ba, value v_off, value v_len) {
   CAMLparam1(v_ba);
   ssize_t ret;
-  size_t off = (size_t)Long_val(v_off);
-  size_t len = (size_t)Long_val(v_len);
-  while (off < len) {
+  ssize_t off = (ssize_t)Long_val(v_off);
+  ssize_t len = (ssize_t)Long_val(v_len);
+  do {
     void *buf = Caml_ba_data_val(v_ba) + off;
     caml_enter_blocking_section();
-    ret = getrandom(buf, len - off, 0);
+    ret = getrandom(buf, len, 0);
     caml_leave_blocking_section();
-    if (ret < 0 && errno == EINTR) continue;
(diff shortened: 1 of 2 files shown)


lib_eio_linux/eio_stubs.c
```diff
@@ -33,10 +34,16 @@ CAMLprim value caml_eio_mkdirat(value v_fd, value v_path, value v_perm) {
...
-  caml_leave_blocking_section();
-  if (ret == -1) uerror("getrandom", Nothing);
+  size_t off = (size_t)Long_val(v_off);
+  size_t len = (size_t)Long_val(v_len);
+  while (off < len) {
+    void *buf = Caml_ba_data_val(v_ba) + off;
+    caml_enter_blocking_section();
+    ret = getrandom(buf, len - off, 0);
+    caml_leave_blocking_section();
+    if (ret < 0 && errno == EINTR) continue;
```
haesbaert, line 44: Bikeshedding a bit, `I'd` suggest comparing explicitly with == -1, that's what caml_eio_mkdirat() above already does, just to keep the same style.

haesbaert: off and len could be ssize_t as well, avoids a signed vs unsigned in off += ret.

bikallem: is `EINTR` == -1 always true? I am mostly following mirage-crypto style since that is what prompted this PR.
https://github.com/mirage/mirage-crypto/blob/main/rng/unix/mc_getrandom_stubs.c#L24

haesbaert: I mean `ret`: ```if (ret == -1 && errno == EINTR) continue;```
manpage says:
`       On error, -1 is returned, and errno is set to indicate the error.`

lib_eio_linux/eio_stubs.c
```diff
@@ -33,10 +34,16 @@ CAMLprim value caml_eio_mkdirat(value v_fd, value v_path, value v_perm) {
...
   CAMLparam1(v_ba);
   ssize_t ret;
-  void *buf = Caml_ba_data_val(v_ba) + Long_val(v_off);
-  caml_enter_blocking_section();
-  ret = getrandom(buf, Long_val(v_len), 0);
-  caml_leave_blocking_section();
-  if (ret == -1) uerror("getrandom", Nothing);
+  size_t off = (size_t)Long_val(v_off);
+  size_t len = (size_t)Long_val(v_len);
+  while (off < len) {
```
talex5, line 39: Why if `off < len`? It's perfectly reasonable to ask for e.g. len=2 bytes at offset=10.

I suggest writing the loop as `do { ... } while (ret == -1 && errno == EINTR)`.

(and `Flow.read_exactly` will handle repeating the read until the buffer is full, so no need to handle that here)

talex5 (changes requested):

lib_eio_linux/eio_stubs.c
bikallem: Excellent suggestion! Done.

lib_eio_linux/eio_stubs.c
bikallem: Ah, yes. You are correct. Incorporated in the last commit.

lib_eio_linux/eio_stubs.c
```diff
@@ -34,16 +34,15 @@ CAMLprim value caml_eio_mkdirat(value v_fd, value v_path, value v_perm) {
...
+  ssize_t off = (ssize_t)Long_val(v_off);
+  ssize_t len = (ssize_t)Long_val(v_len);
+  do {
     void *buf = Caml_ba_data_val(v_ba) + off;
     caml_enter_blocking_section();
     ret = getrandom(buf, len - off, 0);
     caml_leave_blocking_section();
-    if (ret < 0 && errno == EINTR) continue;
-    else if (ret < 0) uerror("getrandom", Nothing);
     off += ret;
```
talex5, line 44: This shouldn't be here. Either `off` is never used again, or `ret` is negative.

bikallem: done

lib_eio_linux/eio_stubs.c
```diff
@@ -33,10 +34,15 @@ CAMLprim value caml_eio_mkdirat(value v_fd, value v_path, value v_perm) {
...
-  void *buf = Caml_ba_data_val(v_ba) + Long_val(v_off);
-  caml_enter_blocking_section();
-  ret = getrandom(buf, Long_val(v_len), 0);
-  caml_leave_blocking_section();
+  ssize_t off = (ssize_t)Long_val(v_off);
+  ssize_t len = (ssize_t)Long_val(v_len);
+  do {
+    void *buf = Caml_ba_data_val(v_ba) + off;
+    caml_enter_blocking_section();
+    ret = getrandom(buf, len - off, 0);
```
talex5, line 42: We just need the length here, don't we?
```suggestion
    ret = getrandom(buf, len, 0);
```

lib_eio_linux/eio_stubs.c
talex5: It's still there. Did git-push fail?

bikallem: Hmmm ... it not there.

talex5 (approved):

talex5:
(CI is failing due to https://github.com/ocurrent/ocurrent/issues/340)

PR #213, opened 2022-04-25
Eio_linux: handle sleeping tasks that are now due before paused fibers

TheLortex:
When using Eio I encountered an issue while porting a test in the mirage `arp` library. 

The failing code can be reduced to the following:
```ocaml
let rec loop () =
  Eio.Fiber.yield ();
  loop ()

let () = run @@ fun ~clock ->
  Fiber.first
    loop
    (fun () -> Eio.Time.sleep clock 0.1)
```

I expected this code to finish after `0.1s` of execution, but instead it does an infinite loop. This is because the scheduler prioritizes paused fibers before sleeping fibers that are now due. 

This PR adds a `Zzz.pop` before checking for paused fibers.

### DIFF
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -128,6 +128,7 @@ type io_job =
   (* When done, remove the cancel_fn from [Suspended.t] and call the callback (unless cancelled). *)
 
 type runnable =
+  | IO : runnable
   | Thread : 'a Suspended.t * 'a -> runnable
   | Failed_thread : 'a Suspended.t * exn -> runnable
 
@@ -432,16 +433,22 @@ let rec schedule ({run_q; sleep_q; mem_q; uring; _} as st) : [`Exit_scheduler] =
   (* This is not a fair scheduler *)
   (* Wakeup any paused fibers *)
   match Lf_queue.pop run_q with
+  | None -> assert false    (* We should always have an IO job, at least *)
   | Some Thread (k, v) -> Suspended.continue k v               (* We already have a runnable task *)
   | Some Failed_thread (k, ex) -> Suspended.discontinue k ex
-  | None ->
+  | Some IO -> (* Note: be sure to re-inject the IO task before continuing! *)
+    (* This is not a fair scheduler: timers always run before all other IO *)
     let now = Unix.gettimeofday () in
     match Zzz.pop ~now sleep_q with
-    | `Due k -> Suspended.continue k ()                   (* A sleeping task is now due *)
+    | `Due k ->
+      Lf_queue.push run_q IO;                   (* Re-inject IO job in the run queue *)
+      Suspended.continue k ()                   (* A sleeping task is now due *)
     | `Wait_until _ | `Nothing as next_due ->
       (* Handle any pending events before submitting. This is faster. *)
       match Uring.peek uring with
-      | Some { data = runnable; result } -> handle_complete st ~runnable result
+      | Some { data = runnable; result } ->
+        Lf_queue.push run_q IO;                   (* Re-inject IO job in the run queue *)
+        handle_complete st ~runnable result
       | None ->
         let num_jobs = Uring.submit uring in
         st.io_jobs <- st.io_jobs + num_jobs;
@@ -452,7 +459,10 @@ let rec schedule ({run_q; sleep_q; mem_q; uring; _} as st) : [`Exit_scheduler] =
         in
         Log.debug (fun l -> l "scheduler: %d sub / %d total, timeout %s" num_jobs st.io_jobs
                       (match timeout with None -> "inf" | Some v -> string_of_float v));
-        if timeout = None && st.io_jobs = 0 then (
+        if not (Lf_queue.is_empty st.run_q) then (
+          Lf_queue.push run_q IO;                   (* Re-inject IO job in the run queue *)
+          schedule st
+        ) else if timeout = None && st.io_jobs = 0 then (
           (* Nothing further can happen at this point.
              If there are no events in progress but also still no memory available, something has gone wrong! *)
           assert (Queue.length mem_q = 0);
@@ -469,6 +479,7 @@ let rec schedule ({run_q; sleep_q; mem_q; uring; _} as st) : [`Exit_scheduler] =
             let result = Uring.wait ?timeout uring in
             Ctf.note_resume system_thread;
             Atomic.set st.need_wakeup false;
+            Lf_queue.push run_q IO;                   (* Re-inject IO job in the run queue *)
             match result with
             | None ->
               (* Woken by a timeout, which is now due, or by a signal. *)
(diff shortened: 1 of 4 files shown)


talex5:
This does need fixing, but I think we need a more general solution. A fiber that keeps yielding will also prevent any other IO from happening, not just sleeps.

Perhaps we should keep a "check-for-IO" job on the run queue. When it gets to the head, we dispatch any IO (and add another check-for-IO job)?

lib_eio/utils/lf_queue.ml
```diff
@@ -137,6 +137,14 @@ let is_empty t =
     ~none:(fun () -> true)
     ~some:(fun _ -> false)
 
+let has_single_element t =
+  Node.fold (Atomic.get t.head.next)
+    ~none:(fun () -> true)
```
talex5, line 142: Should be `false`, from the name.

lib_eio_linux/eio_linux.ml
```diff
@@ -434,7 +435,11 @@ let rec schedule ({run_q; sleep_q; mem_q; uring; _} as st) : [`Exit_scheduler] =
   match Lf_queue.pop run_q with
   | Some Thread (k, v) -> Suspended.continue k v               (* We already have a runnable task *)
   | Some Failed_thread (k, ex) -> Suspended.discontinue k ex
-  | None ->
+  | Some IO as v
+  | (None as v) ->
```
talex5, line 439: I don't think the `None` case can happen.

lib_eio_linux/eio_linux.ml
```diff
@@ -452,16 +457,16 @@ let rec schedule ({run_q; sleep_q; mem_q; uring; _} as st) : [`Exit_scheduler] =
         in
         Log.debug (fun l -> l "scheduler: %d sub / %d total, timeout %s" num_jobs st.io_jobs
                       (match timeout with None -> "inf" | Some v -> string_of_float v));
-        if timeout = None && st.io_jobs = 0 then (
-          (* Nothing further can happen at this point.
+        if timeout = None && st.io_jobs = 0 && Lf_queue.has_single_element run_q then (
```
talex5, line 460: Why is this extra condition needed? If there are no IO jobs at all then we're not even monitoring the wakeup FD, so we must be shutting down at this point anyway.

talex5 (commented):
Thanks; this looks generally sensible!

But with this change, `make test_luv` now hangs, so we probably need a similar fix there.

Other notes:

- I think it would be simpler to delay adding the IO event until just before returning. Then you don't need `has_single_element`. I had a go at that here: https://github.com/ocaml-multicore/eio/compare/main...talex5:fairness?expand=1

- The fake test clock in `prelude.ml` says:
  > At the moment, the scheduler only checks for expired timers when the run-queue is
  > empty, so this is a convenient way to wait for the system to be idle.

  It still seems to be working, but might need to make it more robust, or at least update the comment to warn that it might not work now.

lib_eio_linux/eio_linux.ml
TheLortex: So I assumed `st.io_jobs` could be zero while having some outstanding tasks to run, in that case the scheduler would exit prematurely. But now I have seen that there is always at least one io job running for the eventfd mechanism.

TheLortex:
Thank you for your review.

> But with this change, make test_luv now hangs, so we probably need a similar fix there.

I have added a fix to the luv backend in the second commit. In particular I'm using `Luv.Async.send` to yield back to the event loop when the `IO` task is due.

> I think it would be simpler to delay adding the IO event until just before returning. 

Indeed that sounds better and reduces the diff. I have force-pushed my first commit with the suggested changes

> It still seems to be working, but might need to make it more robust, or at least update the comment to warn that it might not work now.

Do you have an idea on how to make it more robust ? Do you have tests that are waiting for the run queue to become empty ?

talex5:
> I have added a fix to the luv backend in the second commit. In particular I'm using Luv.Async.send to yield back to the event loop when the IO task is due.

This has a large effect on the benchmarks! e.g.

Before:
```
$ EIO_BACKEND=luv make bench
dune exec -- ./bench/bench_yield.exe
n_fibers, ns/iter, promoted/iter
       1,   92.68,        4.9931
       2,   91.56,        4.9955
       3,   91.59,        4.9979
```

After:
```
$ EIO_BACKEND=luv make bench
dune exec -- ./bench/bench_yield.exe
n_fibers, ns/iter, promoted/iter
       1,  968.06,        9.9792
       2,  535.77,        7.4967
       3,  395.65,        6.6500
```

io-uring is also affected, though not as much.
Before:
```
$ EIO_BACKEND=io-uring make bench
dune exec -- ./bench/bench_yield.exe
n_fibers, ns/iter, promoted/iter
       1,   79.11,        4.9939
       2,   77.59,        4.9954
       3,   78.49,        4.9969
```
After:
```
$ EIO_BACKEND=io-uring make bench
dune exec -- ./bench/bench_yield.exe
n_fibers, ns/iter, promoted/iter
       1,  165.01,        9.9854
       2,  118.29,        7.4761
       3,  106.06,        6.6621
```

TheLortex:
Thank you for performing the benchmarks. I expected a performance cost but not that high for the luv backend. A single task yielding is indeed the worst case, the run queue having 1 task and and 1 `IO` at any time, so the `IO` is performed once every two iterations. 

To address the performance cost globally, what we can do is remove the `IO` event from the run queue, and have instead a counter-based strategy:
- initially counter is at zero
- each run queue pop increases the counter
- when the counter is higher than a threshold (which value?), take the IO path instead of popping the next task
- when the IO path is taken set the counter to zero

This strategy has a negligible runtime cost when no busy yielding is happening. 
When busy yielding is happening, the cost will depend on on the number of tasks run between each IO lookup. The higher the threshold, the lower the overhead but it increases latency.

What do you think ?

lib_eio_linux/eio_linux.ml
```diff
@@ -453,7 +460,7 @@ let rec schedule ({run_q; sleep_q; mem_q; uring; _} as st) : [`Exit_scheduler] =
         Log.debug (fun l -> l "scheduler: %d sub / %d total, timeout %s" num_jobs st.io_jobs
                       (match timeout with None -> "inf" | Some v -> string_of_float v));
         if timeout = None && st.io_jobs = 0 then (
-          (* Nothing further can happen at this point.
+          (* Nothing further can happen at this point. The single element in the run queue is the IO job.
```
talex5, line 463:
```suggestion
          (* Nothing further can happen at this point.
```

talex5 (commented):
Thinking on about this, I don't think the performance change for Eio_linux is a problem: the yield benchmark previously wasn't checking for IO, but now it is, and that's correct. It's not surprising it's slightly slower.

I tested the luv changes with the HTTP benchmark (https://github.com/ocaml-multicore/retro-httpaf-bench, adjusted to use `Eio_luv`) and this PR didn't make any noticeable difference, so I think it's fine.

However, the tests are failing in CI (but they work for me locally). Any idea why that is?

TheLortex:
I have no idea why the CI failure is happening. I have also been able to reproduce the test using the Docker instructions of the CI job..

lib_eio_linux/eio_linux.ml
talex5: Sorry, I was wrong about this. We still need to run everything on the run queue even if we are shutting down. I think this is the cause of the CI failures. I've pushed a fix.

TheLortex: Indeed it fixes the CI, thanks for figuring out that it was the root of the issue!

talex5 (approved):

talex5:
Thanks!

PR #218, opened 2022-06-01
Eio_linux: fix read_dir

talex5:
It was ignoring the directory and always reading from env#cwd.

### DIFF
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -779,12 +779,12 @@ module Low_level = struct
       client, client_addr
     )
 
-    let open_dir ?dir ~sw path =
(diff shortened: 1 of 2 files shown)


talex5:
I also optimised it a bit. Testing with:

```ocaml
open Eio.Std

let () =
  Eio_main.run @@ fun env ->
  let t0 = Unix.gettimeofday () in
  let items = Eio.Dir.read_dir env#fs "/nix/store" in
  let t1 = Unix.gettimeofday () in
  traceln "Read %d items in %f seconds" (List.length items) (t1 -. t0)
```
I got:

- `Read 152441 items in 34.181577` seconds with the old code.
- `Read 152441 items in 1.219926` seconds with the new code.

/cc @patricoferris

PR #219, opened 2022-06-01
eio_linux: make read_dir faster

talex5:
This further improves the performance of directory reads by using a larger buffer and by performing all reads in a single systhread, rather than creating a new one for each chunk.

- `Read 152448 items in 1.188156 seconds` with the old code.
- `Read 152448 items in 0.217013 seconds` with the new code.

(continues #218)


patricoferris (approved):
Great! :)) Thanks for the fixes in #218 as well

PR #221, opened 2022-06-09
eio_linux: add support for integration with Async

talex5:
See https://github.com/talex5/async_eio


talex5:
(closing as very stale and there doesn't seem to be interest in this)

PR #224, opened 2022-06-15
Update to 5.0.0~alpha0

talex5:
Drop compatibility with 4.12+domains.

Before merging, this needs:

- https://github.com/ocurrent/ocaml-ci/pull/473
- merlin and utop working on 5.0.

Work-arounds for testing:
```
opam pin add lwt https://github.com/dra27/lwt.git#check-symbols-caml_unix
opam pin utop https://github.com/dra27/utop.git#lib-layout
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+Dockerfile
+_build
+.git
+**/*.swp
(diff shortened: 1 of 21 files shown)


talex5:
This is currently waiting for an odoc-parser release that's compatible with OCaml 5 (there's one in the alpha repository, but that doesn't help with ocaml-ci or opam-repo-ci).

talex5:
Now that eio 0.3 is out, I've added a pin-depends on odoc-parser, on the assumption that it will be officially released before eio 0.4.

talex5:
The lint-doc CI job is failing due to odoc not installing, but we can ignore that for now.

PR #256, opened 2022-07-19
Fiber-local storage

SquidDev:
I'm not super happy with the interface for this, so felt it was worth opening an issue and starting a bit of a discussion before spending any more time polishing this off.

Some comments:
 - I initially went for `Context` to describe the store of fibre-local variables, sort of mimicking Async's `Execution_context` and Python's `contextvars`s library. However, I'm not sure this is the best term - it definitely ends up confusing as eio also has `Fiber_context.t`.

 - The store of fibre-local variables is inherited from the currently running fibre when forking (rather than the passed switch). This is (at least in my minds) more intuitive, but does involve an extra `Effect.perform Get_context` call on each fork, which obviously comes with some performance cost.

### DIFF
--- a/lib_eio/core/eio__core.mli
+++ b/lib_eio/core/eio__core.mli
@@ -287,20 +287,34 @@ module Fiber : sig
       run concurrently in separate fibers.
       @param max_fibers Maximum number of fibers to run concurrently *)
 
-  (** {2 Fiber-local variables} *)
+  (** {2 Fiber-local variables}
+
+      Each fiber maintains a map of additional variables associated with it,
+      which can be used to store fiber-related state or context. This map is
+      propagated to any forked fibers. *)
 
   type 'a key
-  (** ['a key] is a key for a value of type ['a] stored within the fiber's context. *)
+  (** ['a key] is a fiber-local variable of type ['a]. *)
 
   val create_key : unit -> 'a key
-  (** [create_key ()] creates a new context key. *)
+  (** [create_key ()] creates a new fiber-local variable. *)
 
   val get : 'a key -> 'a option
-  (** [get key] reads [key] from the current context, returning its value or {!None} if it has not
-     been defined. *)
+  (** [get key] reads [key] from the map of fiber local variables, returning its
+      value or {!None} if it has not been bound. *)
 
   val with_binding : 'a key -> 'a -> (unit -> 'b) -> 'b
-  (** [with_binding key value fn] runs [fn] with the given key set to the assigned value. *)
+  (** [with_binding key value fn] runs [fn] with [key] bound to the provided
+      [value].
+
+      Whilst this binding only exists for the duration of this function {i on
+      this fiber}, it will be propagated to any forked fibers. If [fn] creates
+      fibers using an external switch, the bound value may be continue to be
+      used after this function returns. *)
+
+  val without_binding : 'a key -> (unit -> 'b) -> 'b
+  (** [with_binding key value fn] runs [fn] with any binding for [key] removed.
+      *)
 end
 
 (** @canonical Eio.Exn *)
--- a/lib_eio/core/eio__core.mli
+++ b/lib_eio/core/eio__core.mli
@@ -291,7 +291,11 @@ module Fiber : sig
 
       Each fiber maintains a map of additional variables associated with it,
       which can be used to store fiber-related state or context. This map is
-      propagated to any forked fibers. *)
+      propagated to any forked fibers.
+
+      While fiber-local variables can be useful, they can also make code much
+      harder to reason about, as they effectively act as another form of global
+      state. When possible, prefer passing arguments around explicitly. *)
 
   type 'a key
   (** ['a key] is a fiber-local variable of type ['a]. *)
--- a/lib_eio/core/eio__core.mli
+++ b/lib_eio/core/eio__core.mli
@@ -286,6 +286,21 @@ module Fiber : sig
   (** [iter f x] is like [List.iter f x] except that the invocations of [f] are
       run concurrently in separate fibers.
       @param max_fibers Maximum number of fibers to run concurrently *)
+
+  (** {2 Fiber-local variables} *)
+
+  type 'a key
+  (** ['a key] is a key for a value of type ['a] stored within the fiber's context. *)
+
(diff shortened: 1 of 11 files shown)


talex5 (commented):
Thanks - this looks good!

I'm not sure about the interface either. I've made a branch experimenting with some naming changes here: https://github.com/ocaml-multicore/eio/compare/main...talex5:eio:fiber-storage?expand=1

- I moved the basic get and set functions to `Fiber` (e.g. `Fiber.get key`), which avoids the word "context" and maybe makes it more obvious they're attached to fibers. It could do with an introduction explaining what fiber-local variables are for, and maybe discouraging over-use of them.
- Renamed `context` to `vars` in `Fiber_context`.
- Renamed `with_value` -> `with_binding`. We might want a `without_binding` too.
- I removed the convenience `Fiber_context.make`. I think it's a bit clearer to be explicit here. For example, `Fiber.any` really only needs to get them once.
- I removed the code to inherit across domains for now. I think this is a bit risky, because the values may not be thread-safe. `Domain` has a `split_from_parent` argument (https://github.com/ocaml/ocaml/blob/8dda88493c0e8bfec83d3d40129bc6e0fb7cef29/stdlib/domain.mli#L95). Possibly something like that will be useful, but we can add it later.
- For now, I removed the API to get and set the whole set at once (only domains needed it). If it comes back, it should probably go in `Private`.

Do you think this is better?

SquidDev:
> I moved the basic get and set functions to `Fiber`

Oh, that's a fantastic change. That makes things much nicer!

> I removed the code to inherit across domains for now. I think this is a bit risky, because the values may not be thread-safe. 
> 
> For now, I removed the API to get and set the whole set at once (only domains needed it). If it comes back, it should probably go in `Private`.

Ahh, that's a good point. And agreed - I added this going "_weelllll_, I can see when this might be useful", but probably worth deferring until there's more concrete use-cases.

Are you happy for me pull in your changes and then polish docs off, etc...?

talex5:
> Are you happy for me pull in your changes and then polish docs off, etc...?

That would be great - thanks!

SquidDev (commented):
I've fleshed out the docs a little. I'm afraid I found it much harder to write a succinct introduction to them, and I think what I've ended up with is a possibly a little _too_ sparse.

lib_eio/core/eio__core.mli
```diff
@@ -403,11 +436,13 @@ module Private : sig
   module Fiber_context : sig
     type t
 
+    type vars
```
talex5, line 439: Since this is abstract and only used as an input, it can probably be removed. I guess the `make` function doesn't need to be exposed either.

lib_eio/core/fiber.ml
```diff
@@ -128,7 +132,8 @@ let any fs =
           | [] -> await_cancel ()
           | [f] -> wrap f; []
           | f :: fs ->
-            let new_fiber = Cancel.Fiber_context.make ~cc in
+            let vars = Cancel.Fiber_context.get_vars () in
```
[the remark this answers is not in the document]
talex5: Yes please (I was just lazy).

lib_eio/domain_manager.ml
```diff
@@ -13,6 +13,7 @@ let run (t : #t) fn =
   let cancelled, set_cancelled = Promise.create () in
   Private.Fiber_context.set_cancel_fn ctx (Promise.resolve set_cancelled);
   (* If the spawning fiber is cancelled, [cancelled] gets set to the exception. *)
+
```
talex5, line 16:
```suggestion
```

tests/test_domains.md
```diff
@@ -189,3 +189,78 @@ Can't fork into another domain:
...
+  trace_key ();
+  Fiber.with_binding key 456 (fun () -> trace_key ());
+  trace_key ();;
++Key => 123
++Key => 456
++Key => 123
+- : unit = ()
+```
+
+Values are propagated when forking, or sending fibers to other domains.
```
talex5, line 239:
```suggestion
Values are propagated when forking, but not when spawning fibers in other domains (as the values may not be thread-safe).
```

lib_eio/core/eio__core.mli
```diff
@@ -286,6 +286,39 @@ module Fiber : sig
...
+
+  (** {2 Fiber-local variables}
+
+      Each fiber maintains a map of additional variables associated with it,
+      which can be used to store fiber-related state or context. This map is
+      propagated to any forked fibers.
+
+      While fiber-local variables can be useful, they can also make code much
+      harder to reason about, as they effectively act as another form of global
+      state. When possible, prefer passing arguments around explicitly. *)
```
talex5, line 298:
```suggestion
      state. When possible, prefer passing arguments around explicitly.
      
      Fiber-local variables are particularly useful for attaching extra information for debugging,
      such as a request ID that the log system can include in all logged messages. *)
```

lib_eio/core/eio__core.mli
```diff
@@ -286,6 +286,39 @@ module Fiber : sig
...
+
+  type 'a key
+  (** ['a key] is a fiber-local variable of type ['a]. *)
```
talex5, line 301:
```suggestion
  (** ['a key] is a fiber-local variable of type ['a].
      
      Since the key is required to get or set a variable,
      a library can keep its key private to control how the variable can be accessed. *)
```

talex5 (commented):
Thanks, this is looking good :-)

lib_eio/core/fiber.ml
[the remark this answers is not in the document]
SquidDev: I wasn't sure if it was worth checking if the list was empty here - I'd personally argue not, as `any []` should be pretty rare, but happy to handle it if you disagree!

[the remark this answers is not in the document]
talex5: It's fine I think. If we really cared about optimising the empty-list case, we'd do it at the start of the function anyway.

talex5 (approved):
This PR is still marked as a draft, but it looks ready to merge to me!

talex5:
Thanks!

SquidDev:
Thank you for all your help too! Looking forward to using this :)

PR #261, opened 2022-07-25
eio-linux: fix IPv6 support

talex5:
Use `PF_INET6` for IPv6 addresses, not `PF_INET`.

Reported by @haesbaert.


haesbaert (approved):
reads fine to me \o/

Issue #124, opened 2022-01-03
Decide how to represent pathnames


talex5:
At the moment, paths in Eio are strings, which are used in the context of some base directory. e.g.

```ocaml
Eio.Dir.with_open_in cwd "foo/input.dat" (fun flow -> ...)
```

This should work well on most systems, but we need to think about how to support Windows too.

Possibilities include:

- Using Unix path strings everywhere and providing a way to import and export Windows paths.
- Using an abstract type for paths.

The [Fpath](https://erratique.ch/software/fpath) library encodes lots of knowledge about Windows paths. It probably can't be used directly because its paths behave differently depending on the host platform (preventing e.g. a program running on Linux from creating paths that are to be used on Windows). The path rules should probably be per-filesystem instead.

- [ ] Find out what other languages do (e.g. look at [Python's pathlib](https://docs.python.org/3/library/pathlib.html), Rust, etc).

samwgoldman:
My 2¢ here is to leave paths as strings. Getting paths "right" for all use cases doesn't seem possible, and having a flexible low-level API somewhere, even if it opens up the possibility for errors, seems valuable to me.

For Flow, we have the need to share paths between platforms. These are relative paths, using the `/` path separator. We only support Windows systems that transparently support the `/` path separator, so we can just pass these paths into Windows APIs without transforming them.

However, we do build wrappers around the various paths, to ensure they are not mishandled. But this is an application-layer concern for us. We take the strings returned by system calls into the appropriate (unboxed) wrapper type and coerce back into strings as needed.

talex5:
Thanks, that makes sense. And I see that DOS has always supported `/` as a path separator anyway:

https://learn.microsoft.com/en-us/archive/blogs/larryosterman/why-is-the-dos-path-character

avsm:
Do you have a link to your path wrapper library, @samwgoldman?  It'd be interesting to contrast it with Fpath and Python's Pathlib. I agree that keeping strings at a low-level in Eio is a reasonable for Eio 1.0.

PR #266, opened 2022-07-28
Change Dir.t to be a pair of a dirfd and a path

talex5:
Instead of passing two arguments each time we want to refer to something, we now just pass the tuple.

This also avoids having to pass e.g. `cwd "."` - you can just use `cwd`.

`Dir.(/)` provides filename concatenation, which might make it easier to support Windows in future (see #124).

It makes the examples a bit longer, but also a bit clearer, I think.


talex5:
I've now split `Dir` into `Fs` and `Path`, since a `Dir.t` was no longer always a directory.

PR #274, opened 2022-08-04
Clarify OCaml version constraint

talex5:
`ocaml-base-compiler` is version `5.0.0~alpha1` but `ocaml` itself is just `5.0.0`.

(spotted by @kit-ty-kate in https://github.com/ocaml/opam-repository/pull/21935)

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -12,7 +12,7 @@
  (synopsis "Effect-based direct-style IO API for OCaml")
  (description "An effect-based IO API for multicore OCaml with fibers.")
  (depends
-  (ocaml (>= 5.0.0~alpha0))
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
could you also add the `"@runtest" {with-test & os != "macos"}` for eio-main too?

talex5:
> could you also add the "@runtest" {with-test & os != "macos"} for eio-main too?

I don't think so. That bit is generated by dune.

kit-ty-kate:
you can always stop generating the opam file by dune. It’s fairly useless and more bothersome than anything once first generated IMO

talex5:
Merging this part for now. I've opened #275 to track the test problem.

Issue #258, opened 2022-07-23
DNS utility functions


v0idpwn:
Hi! First off, thanks for the amazing work in eio!

I'd like to have an API for `uv_getaddrinfo` and `uv_getnameinfo`. Luv has bindings to these on their DNS module: https://github.com/aantron/luv/blob/0ed668e87ca55d65f4f84422b0dd3381701c71b0/src/DNS.ml

If this feature is welcome in `eio`, I'd gladly send a PR.

talex5:
Yes please! For uring, something like this could be used:

```ocaml
Eio_unix.run_in_systhread (fun () -> Unix.gethostbyaddr name)
```

There is a whole family of these `get*` functions in OCaml's `Unix` module. I'm not sure how many are useful. The types will need to be wrapped, because `Eio.Net` can't depend on `Unix` (it has to work on non-Unix platforms like MirageOS too).

talex5:
OK, thanks for confirming!

PR #278, opened 2022-08-11
Add support for getaddrinfo

bikallem:
Fixes part of https://github.com/ocaml-multicore/eio/issues/258

### DIFF
--- a/lib_eio/mock/net.ml
+++ b/lib_eio/mock/net.ml
@@ -5,18 +5,21 @@ type t = <
   on_listen : Eio.Net.listening_socket Handler.t;
   on_connect : <Eio.Net.stream_socket; Eio.Flow.close> Handler.t;
   on_datagram_socket : <Eio.Net.datagram_socket; Eio.Flow.close> Handler.t;
+  on_getaddrinfo : Eio.Net.Sockaddr.t list Handler.t;
 >
 
 let make label =
   let on_listen = Handler.make (`Raise (Failure "Mock listen handler not configured")) in
   let on_connect = Handler.make (`Raise (Failure "Mock connect handler not configured")) in
   let on_datagram_socket = Handler.make (`Raise (Failure "Mock datagram_socket handler not configured")) in
+  let on_getaddrinfo = Handler.make (`Raise (Failure "Mock getaddrinfo handler not configured")) in
   object
     inherit Eio.Net.t
 
     method on_listen = on_listen
     method on_connect = on_connect
     method on_datagram_socket = on_datagram_socket
+    method on_getaddrinfo = on_getaddrinfo
 
     method listen ~reuse_addr:_ ~reuse_port:_ ~backlog:_ ~sw addr =
       traceln "%s: listen on %a" label Eio.Net.Sockaddr.pp addr;
@@ -35,6 +38,10 @@ let make label =
       let socket = Handler.run on_datagram_socket in
       Switch.on_release sw (fun () -> Eio.Flow.close socket);
       socket
+
+    method getaddrinfo ~service node =
+      traceln "%s: getaddrinfo ~service:%s %s" label service node;
+      Handler.run on_getaddrinfo
   end
 
 let on_connect (t:t) actions =
@@ -48,6 +55,8 @@ let on_listen (t:t) actions =
 let on_datagram_socket (t:t) actions =
   let as_socket x = (x :> <Eio.Net.datagram_socket; Eio.Flow.close>) in
   Handler.seq t#on_datagram_socket (List.map (Action.map as_socket) actions)
+
+let on_getaddrinfo (t:t) actions = Handler.seq t#on_getaddrinfo actions
 
 type listening_socket = <
   Eio.Net.listening_socket;
--- a/lib_eio/net.mli
+++ b/lib_eio/net.mli
@@ -104,6 +104,7 @@ class virtual t : object
   method virtual listen : reuse_addr:bool -> reuse_port:bool -> backlog:int -> sw:Switch.t -> Sockaddr.stream -> listening_socket
   method virtual connect : sw:Switch.t -> Sockaddr.stream -> <stream_socket; Flow.close>
   method virtual datagram_socket : sw:Switch.t -> Sockaddr.datagram -> <datagram_socket; Flow.close>
+  method virtual getaddrinfo : service:string -> string -> Sockaddr.t list
 end
 
 (** {2 Out-bound Connections} *)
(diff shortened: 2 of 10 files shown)


lib_eio/net.mli
```diff
@@ -171,6 +172,12 @@ val recv : #datagram_socket -> Cstruct.t -> Sockaddr.datagram * int
     returned along with the sender address and port. If the [buf] is too small then excess bytes may be discarded
     depending on the type of the socket the message is received from. *)
 
+(** {2 DNS queries} *)
+
+val addrinfo: #t -> string -> ([`Tcp | `Udp] * Ipaddr.v4v6) list
```
talex5, line 177:
```suggestion
val addrinfo: #t -> string -> Sockaddr.t list
```
(or at least a compatible subset, if you want to exclude Unix-domain sockets)

bikallem:
@talex5 do you know if the `ocaml-ci` has ipv6 enabled? or how to enable it? The test for this PR seems to be failing as I expect Ipv6 address but the machine is not returning any.

lib_eio_luv/eio_luv.ml
```diff
@@ -407,6 +407,24 @@ module Low_level = struct
...
+              let st = match socktype' with `STREAM -> `Tcp | `DGRAM -> `Udp in
+              let ipaddr = Eio_unix.Ipaddr.of_unix (Unix.inet_addr_of_string host) in
+              (st, ipaddr)
+            )
+      | _ -> None
+    in
+    let request = Luv.DNS.Addr_info.Request.make () in
+    await_with_cancel ~request (fun loop -> Luv.DNS.getaddrinfo ~loop ~request ~node:hostname ())
+    |> Result.map (List.filter_map convert_addrinfo)
+    |> or_raise
```
talex5, line 426:
```suggestion
    |> or_raise
    |> List.filter_map convert_addrinfo
```
(just a bit simpler)

tests/network.md
```diff
@@ -485,3 +485,24 @@ EPIPE:
     (fun () -> Eio.Flow.shutdown a `Receive);;
 - : unit = ()
 ```
+
+## Addrinfo (Linx and Libuv)
```
talex5, line 489:
```suggestion
## Addrinfo (Linux and Libuv)
```

tests/network.md
```diff
@@ -485,3 +485,24 @@ EPIPE:
...
+```ocaml
+# Eio_main.run @@ fun env ->
+  Eio.Net.addrinfo env#net "localhost";;
+- : ([ `Tcp | `Udp ] * Eio.Net.Ipaddr.v4v6) list =
+[(`Tcp, "\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\001");
+ (`Udp, "\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\001");
+ (`Tcp, "\127\000\000\001"); (`Udp, "\127\000\000\001")]
+```
+
+<!-- $MDX set-EIO_BACKEND=luv -->
```
talex5, line 500: We need a proper fix for this (at the moment, I run `make test_luv` to run all the tests this way). Probably we should have a `tests/luv` directory with some dune magic to duplicate all the tests.

talex5 (commented):
Thanks! This looks reasonable to me.

I think we should call this `getaddrinfo` rather than `addrinfo` though, as that's the name everyone uses for it (including OCaml's Unix module). Also, it may access the network so the getting part is quite important.

> do you know if the ocaml-ci has ipv6 enabled? or how to enable it? The test for this PR seems to be failing as I expect Ipv6 address but the machine is not returning any.

What gets returned for `localhost` will depend on the system configuration, so it's probably best to test it using numeric addresses.

lib_eio/net.mli
bikallem: Ended up using `Sockaddr.t` as suggested.

lib_eio_luv/eio_luv.ml
bikallem: Done.

tests/network.md
bikallem: Removed the redundant line and luv test.

tests/network.md
bikallem: Removed luv test and using `make test_luv` instead.

bikallem:
@talex5 I have renamed `addrinfo` to `getaddrinfo` as suggested and removed the luv specific test from `network.md`.

SquidDev:
One thing possibly worth mentioning is that users may wish to override this method to point to a non-libc implementation (especially on Alpine, where musl's `getaddrinfo` implementation is intentionally limited).

Supporting this use case isn't a problem (capabilities make this very easy to do), but I do feel a bit weird about calling this function `getaddrinfo` when it no longer uses the POSIX function - I'd personally go for something like `resolve_addr`.

bikallem:
@SquidDev Thanks for the feedback. Don't really have any personal preference but I agree with @talex5 - `getaddrinfo` is the more common term for what we are doing here since both sets of backend (luv and unix) both name it as such.

talex5 (commented):
Either `getaddrinfo` or `resolve_addr` is fine with me. I don't think there's any problem with `getaddrinfo` calling an alternative implementation as long as it does roughly the same thing. Otherwise we'd have to rename everything else too (e.g. `unlink` and `mkdir` might not be calling the glibc function either, if it's a remote directory, or a directory inside a zip archive).

lib_eio/net.mli
```diff
@@ -171,6 +172,12 @@ val recv : #datagram_socket -> Cstruct.t -> Sockaddr.datagram * int
     returned along with the sender address and port. If the [buf] is too small then excess bytes may be discarded
     depending on the type of the socket the message is received from. *)
 
+(** {2 DNS queries} *)
+
+val getaddrinfo: #t -> string -> Sockaddr.t list
```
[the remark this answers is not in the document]
haesbaert: I'd like that !

lib_eio/mock/net.ml
```diff
@@ -5,18 +5,21 @@ type t = <
...
   on_connect : <Eio.Net.stream_socket; Eio.Flow.close> Handler.t;
   on_datagram_socket : <Eio.Net.datagram_socket; Eio.Flow.close> Handler.t;
+  on_getaddrinfo : Eio.Net.Sockaddr.t list Handler.t;
 >
 
 let make label =
   let on_listen = Handler.make (`Raise (Failure "Mock listen handler not configured")) in
   let on_connect = Handler.make (`Raise (Failure "Mock connect handler not configured")) in
   let on_datagram_socket = Handler.make (`Raise (Failure "Mock datagram_socket handler not configured")) in
+  let on_getaddrinfo = Handler.make (`Raise (Failure "Mock addrinfo handler not configured")) in
```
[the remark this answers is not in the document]
bikallem: Ack.

lib_eio/mock/net.ml
```diff
@@ -49,6 +56,9 @@ let on_datagram_socket (t:t) actions =
   let as_socket x = (x :> <Eio.Net.datagram_socket; Eio.Flow.close>) in
   Handler.seq t#on_datagram_socket (List.map (Action.map as_socket) actions)
 
+let on_getaddrinfo (t:t) actions =
+  Handler.seq t#on_getaddrinfo (List.map (Action.map Fun.id) actions)
```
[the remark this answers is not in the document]
bikallem: Ack.

lib_eio/net.mli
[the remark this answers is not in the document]
bikallem: Done. I have added the `service` parameter now.

talex5 (approved):
Looks good to me! I added one test for port numbers and squash-rebased it on `main`.

PR #279, opened 2022-08-14
Allow sharing of libuv poll handles

patricoferris:
Multiple poll handles for the same file descriptor are not allowed by libuv as @talex5 noted in https://github.com/ocaml-multicore/lwt_eio/issues/11

> http://docs.libuv.org/en/v1.x/poll.html says:
>
>> It is not okay to have multiple active poll handles for the same socket, this can cause libuv to busyloop or otherwise ? malfunction.

This can happen seemingly quite easily, for example:

```ocaml
let () =
  Eio_luv.run @@ fun _env ->
  Eio.Switch.run @@ fun sw ->
  let src, _dst = Eio_unix.socketpair ~sw () in
  let fd = Option.get @@ Eio_unix.FD.peek_opt src in
  Eio.Fiber.both
    (fun () -> Eio_unix.await_writable fd)
    (fun () -> Eio_unix.await_writable fd)
(* Eio_luv.Luv_error(EEXIST) (* file already exists *) *)
```

This PR makes it so we keep track of the various fibers that are waiting on read and write events and we enqueue them accordingly based on the events that the polling callback receives. This means the same handle must register for both kinds of events and then we decided how to handle them rather than having a dedicated `READABLE` and a dedicated `WRITABLE` poll handle.

So far this PR fixes both the simple example and the origin Irmin example that caused me to find this ([code here](https://github.com/ocaml-multicore/lwt_eio/issues/11#issue-1337047305)), however I haven't tried:

 - [ ] Awaiting across multiple domains
 - [ ] I believe the logic for when to stop a handle is currently incorrect, a fiber could be cancelled but this doesn't remove it from the `Lf_queue` and I'm using the `Lf_queue` size to check whether there are any fibers waiting on the handle -- perhaps an integer that we decrement/increment would suffice here?

### DIFF
--- a/lib_eio_luv/eio_luv.ml
+++ b/lib_eio_luv/eio_luv.ml
@@ -148,8 +148,9 @@ let get_loop () =
 let cancel_all_rw_waiters t fd =
   match Fd_map.find_opt fd t.fd_map with
   | Some v ->
-    apply_all v.read (fun k -> enqueue_and_remove t enqueue_failed_thread k (Failure "Closed file descriptor whilst polling"));
-    apply_all v.write (fun k -> enqueue_and_remove t enqueue_failed_thread k (Failure "Closed file descriptor whilst polling"));
+    let ex = Failure "Closed file descriptor whilst polling" in
+    apply_all v.read (fun k -> enqueue_and_remove t enqueue_failed_thread k ex);
+    apply_all v.write (fun k -> enqueue_and_remove t enqueue_failed_thread k ex);
     maybe_stop t v
   | None -> ()
 
@@ -190,7 +191,7 @@ module Low_level = struct
       mutable release_hook : Eio.Switch.hook;        (* Use this on close to remove switch's [on_release] hook. *)
       close_unix : bool;
       mutable fd : [`Open of 'a Luv.Handle.t | `Closed]
-    }
+    } constraint 'a = [< `Poll | `Stream of [< `Pipe | `TCP | `TTY ] | `UDP ]
 
     let get op = function
       | { fd = `Open fd; _ } -> fd
@@ -200,18 +201,6 @@ module Low_level = struct
       | { fd = `Open _; _ } -> true
       | { fd = `Closed; _ } -> false
 
-    let cancel_waiters t fd =
-      (* Luv doesn't expose a way to optionally get a file number for an ['a Luv.Handle.t] handle.
-         So we bypass the typechecker and [fileno] will raise EINVAL if a handle that doesn't support
-         file descriptors is used. *)
-      let fd = Obj.magic fd in
-      match Luv.Handle.fileno fd with
-        | Ok os_fd ->
-          let fd = Luv_unix.Os_fd.Fd.to_unix os_fd in
-          cancel_all_rw_waiters t fd
-        | Error `EINVAL -> ()
-        | Error e -> raise (Luv_error e)
-
     let close t =
       Ctf.label "close";
       let fd = get "close" t in
@@ -219,7 +208,11 @@ module Low_level = struct
       Eio.Switch.remove_hook t.release_hook;
       if t.close_unix then (
         enter_unchecked @@ fun t k ->
-        cancel_waiters t fd;
+        begin match Luv.Handle.fileno fd with
+          | Ok fd -> cancel_all_rw_waiters t (Luv_unix.Os_fd.Fd.to_unix fd)
+          | Error `EBADF -> ()  (* We don't have a Unix FD yet, so we can't be watching it. *)
+          | Error e -> raise (Luv_error e)
+        end;
         Luv.Handle.close fd (enqueue_thread t k)
       )
 
--- a/lib_eio_luv/eio_luv.ml
+++ b/lib_eio_luv/eio_luv.ml
@@ -81,7 +81,8 @@ type runnable =
   | IO
   | Thread of (unit -> unit)
 
-type 'a fd_event_waiters = {
+type fd_event_waiters = {
+  fd : Unix.file_descr;
   handle : Luv.Poll.t;
   read : unit Suspended.t Lwt_dllist.t;
   write : unit Suspended.t Lwt_dllist.t;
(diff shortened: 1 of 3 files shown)


talex5:
I had a look at lwt_luv to see how they handle it:

```ocaml
open Lwt.Infix

let () =
  let () = Lwt_engine.set (new Lwt_luv.engine) in
  let socket, _ = Lwt_unix.(socketpair PF_UNIX SOCK_STREAM 0) in
  Lwt_main.run begin
    let _a = Lwt_unix.wait_read socket in
    let rec aux () =
      Lwt_unix.wait_write socket >>= fun () ->
      Lwt_unix.write socket (Bytes.of_string "a") 0 1 >>= fun (_ : int) ->
      aux ()
    in
    aux ()
  end
```
The result is:
```
Fatal error: exception Failure(
"Could not register fd for write polling, this is probably a error in Lwt,
please open a issue on the repo. \n
Error message: EEXIST")
```
However, Lwt does handle the case of multiple readers OK (just not a mix of readers and writers):
https://github.com/ocsigen/lwt/blob/5.6.1/src/unix/lwt_unix.cppo.ml#L495

> I believe the logic for when to stop a handle is currently incorrect, a fiber could be cancelled but this doesn't remove it from the Lf_queue and I'm using the Lf_queue size to check whether there are any fibers waiting on the handle -- perhaps an integer that we decrement/increment would suffice here?

You probably want a `Lwt_dllist` here instead of an `Lf_queue`. There shouldn't be any need for anything lock-free here, since we run a separate luv event loop in each domain.

> This means the same handle must register for both kinds of events and then we decided how to handle them

I don't think this will work. http://docs.libuv.org/en/v1.x/poll.html#c.uv_poll_start says:

>  If one of the events UV_READABLE or UV_WRITABLE are set, the callback will be called again, as long as the given fd/socket remains readable or writable accordingly.

However, it also seems unnecessary:
> Note: Calling uv_poll_start() on a handle that is already active is fine. Doing so will update the events mask that is being watched for. 

It's not clear to me what happens to the callback if you call `start` multiple times. Possibly it ignores the second one, or disconnects the first. In any case, `stop` doesn't say which callback to stop.

patricoferris:
Yes, I'm not really sure what the actual behaviour is. From the source code, it would seem after calling `Poll.start` multiple times a new callback is installed each time https://github.com/libuv/libuv/blob/988f2bfc4defb9a85a536a3e645834c161143ee0/src/unix/poll.c#L152. In [this tutorial](http://docs.libuv.org/en/v1.x/guide/utilities.html) under "Setting up Polling", it would seem to suggest this is the case and they're happy enough because it is the same callback installed each time.

In https://github.com/ocaml-multicore/eio/pull/279/commits/b12c0d6b71afda6cfe534a75e29deb7ad2351d31 I've made it such that callbacks are the same and a cancelled fiber will remove itself from the list of continuations to enqueue so the `is_empty` check (in my mind) in `safe_to_stop_polling` is now correct.

In https://github.com/ocaml-multicore/eio/pull/279/commits/499963077745290812fa8cc4bfa37018e8e14850 I've added the extra logic to handle simultaneously awaiting for writability and readability on a FD. The bit mask for events is updated to whatever you pass to `Poll.start`, I had thought it might bitwise OR with whatever is already there, this is not the case.

talex5 (commented):
This looks good to me (apart from minor whitespace issues). I've put some suggested simplifications at https://github.com/ocaml-multicore/eio/commit/4985168bfd11660e03fe2d658c4048cbd6c914d0. Feel free to include those if they look sensible. I put `fd` in the `fd_event_waiters`. We have to store it anyway, and it simplifies things to have it here rather than captured in a closure. Sorry about the extra re-indentation!

Do we need to fail all waiters if `close` is called too?

patricoferris:
Yes your changes are very sensible :)) What tool do you use for indentation and whitespaces, ocp-indent?

> Do we need to fail all waiters if close is called too?

This seems reasonable, the docs say:

> The user should not close a file descriptor while it is being polled by an active poll handle. This can cause the handle to report an error, but it might also start polling another socket. However the fd can be safely closed immediately after a call to [uv_poll_stop()](http://docs.libuv.org/en/v1.x/poll.html#c.uv_poll_stop) or [uv_close()](http://docs.libuv.org/en/v1.x/handle.html#c.uv_close).

It seems reasonable to me to fail any suspended fibers, would this approach work on Linux?

talex5 (approved):
> What tool do you use for indentation and whitespaces, ocp-indent?

I use ocp-indent for indentation. For whitespace, `.git/hooks/pre-commit.sample` will detect trailing spaces (if you remember to enable it).

> It seems reasonable to me to fail any suspended fibers, would this approach work on Linux?

I'm not sure if it's worth doing. I think Linux only uses the FD to look up the file object when you submit the request, and it's quite happy to continue waiting for the file after the FD is closed. It's more a limitation of luv that files can't be used after their FD is closed. Maybe it's needed on Windows? Though it might be nice to be consistent.

I've pushed a couple more changes:

- One removes the use of `Obj.magic` by stashing the FD away at the start. Using Obj.magic with an external library like this seems pretty risky.
- I got it to update the poll mask in the case where one queue becomes empty but the other isn't. It's hard to test for this because it just uses a lot of CPU if you forget, but doesn't stop working. I also moved the code around a bit to group all the polling stuff together.

If that looks OK to you, I'm happy to merge it.

talex5:
> One removes the use of `Obj.magic` ...

Hmm, this doesn't work for all handles. I guess the FD isn't known initially in some cases.

talex5:
OK, fixed it by adding a constraint. It seems we only ever use handles types that may have FDs anyway. I also got it to ignore EBADF when getting the FD, since sockets don't get their FD set immediately (e.g. a failure after `TCP.init` and before `bind` wouldn't have any FD).

patricoferris:
> OK, fixed it by adding a constraint. It seems we only ever use handles types that may have FDs anyway.

Ah, I wasn't sure if that was the case. Definitely better than `Obj.magic` ! All looks good to me :))

talex5:
Thanks!

PR #285, opened 2022-08-21
Mark getaddrinfo tests as non-deterministic

patricoferris:
On macOS Catalina these tests return the same list only with the `` `UDP`` entry _before_ thee `` `TCP`` entry, so either they should be marked as non-deterministic (as in this PR) or we need a stable way to sort the list before returning it.

### DIFF
--- a/tests/network.md
+++ b/tests/network.md
@@ -488,6 +488,7 @@ EPIPE:
 
 ## Getaddrinfo
 
+<!-- $MDX non-deterministic=output -->
 ```ocaml
 # Eio_main.run @@ fun env ->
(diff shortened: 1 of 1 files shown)


talex5:
Maybe we should have `getaddrinfo_stream` and `getaddrinfo_datagram` to limit the results to a specific type?

The order of results from `getaddrinfo` is significant, so we shouldn't sort in the function itself. `man getaddrinfo` says:

> Normally, the application should try using the addresses in the order in which they are returned.  The sorting function used within getaddrinfo() is defined in RFC 3484; the order can be tweaked for a particular  system  by  editing /etc/gai.conf

However, we could sort the results in the tests.

PR #291, opened 2022-08-26
Add Eio.Net.getaddrinfo_{stream,datagram}

talex5:
These filter the results of `getaddrinfo` to ensure you get something of the correct type. This makes it possible to use the result directly with e.g. `connect`.

This is also used in the tests to avoid non-deterministic ordering of TCP vs UDP results, which fixes #285.

/cc @bikallem


bikallem:
Looks good. 👍

PR #296, opened 2022-08-30
net: add Net.with_tcp_connect

bikallem:
`Net.with_tcp_connect` takes a domain-name/port/service and attempts to
establish connection to it. IP addresses for the host are tried one
after the other while preferring/trying IPv6 protocol addresses first
if available.

We also introduce two new exceptions:
1. Invalid_host - raised when a connection couldn't be established
  for any of the addresses defined for a given host/port/service.
2. Unix_error - introduced to mimic Unix.Unix_error without depending
  on "unix" library.

Update: It seems `ocaml-ci` doesn't support connecting to the internet (www.example.com here). So I have annotated the test as non-deterministic command in mdx.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,10 @@ Eio replaces existing concurrency libraries such as Lwt
   * [Async](#async)
   * [Lwt](#lwt)
   * [Unix and System Threads](#unix-and-system-threads)
+* [Best Practices](#best-practices)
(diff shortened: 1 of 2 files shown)


bikallem:
Closed this PR in support of https://github.com/ocaml-multicore/eio/pull/302. I was pushing the changes to an incorrect branch.

PR #299, opened 2022-08-31
Add Best Practices section to README

talex5:
/cc @avsm


avsm:
Looks good to me! Especially the switch one is useful to have explicitly noted here.

Issue #297, opened 2022-08-30
Implement signature that expects covariant type parameter


anmonteiro:
Most signatures that abstract over Lwt / Async currently define an interface to be implemented by the I/O runtime.

Part of that interface is usually:

```ocaml
type +'a t
```

I took a stab at implementing one of these (Caqti, in particular) for my own use, and I was trying to use an `Eio.Promise.t` (which is itself defined as `type !'a t`). The OCaml compiler doesn't seem to like it:

```
Error: In this definition, expected parameter variances are not satisfied.
       The 1st type parameter was expected to be covariant,
       but it is injective invariant.
```

Do you have a recommended solution? Is my only hope to implement a wrapper of sorts, or alternatively can the type parameter for Eio.Promise.t be made covariant?

talex5:
It can be done using `Obj.magic` internally (which is what Lwt does: https://github.com/ocsigen/lwt/blob/01eb4583f1f3a782351621248c7cb705056fb63e/src/core/lwt.ml#L546).

I think this is a reasonable change to make. Do you want to submit a PR?

anmonteiro:
I opened an initial draft in https://github.com/ocaml-multicore/eio/pull/300

PR #82, opened 2021-10-08
Migrate to 4.12.0+domains effects implementation(syntax-free effects version)

bikallem:
This PR updates `eio` to the OCaml 5.0 supported effects implementation. The migration strategy used is a mechanical one, i.e. `effect` -> ` type _ eff += ..`, `match .. with effect .. ` to `match_with`. 

Note: The PR replicates (1to1) both passing and failing tests in the repo.

Edit: Not strictly necessary but this PR(https://github.com/ocaml-multicore/ocaml-uring/pulls) is needed to make `dune b` happy since `ocaml-uring` is a git submodule of this repo.

PR #300, opened 2022-08-31
Make the type parameter for Promise.t covariant

anmonteiro:
Fixes https://github.com/ocaml-multicore/eio/issues/297

### DIFF
--- a/lib_eio/core/eio__core.mli
+++ b/lib_eio/core/eio__core.mli
@@ -129,7 +129,7 @@ module Promise : sig
   type +'a t
   (** An ['a t] is a promise for a value of type ['a]. *)
 
-  type 'a u
+  type -'a u
   (** An ['a u] is a resolver for a promise of type ['a]. *)
 
   val create : ?label:string -> unit -> 'a t * 'a u
--- a/lib_eio/core/eio__core.mli
+++ b/lib_eio/core/eio__core.mli
@@ -126,10 +126,10 @@ module Promise : sig
           (fun () -> Promise.resolve resolver 42)
       ]} *)
 
-  type +'a t
+  type +!'a t
   (** An ['a t] is a promise for a value of type ['a]. *)
 
-  type -'a u
+  type -!'a u
   (** An ['a u] is a resolver for a promise of type ['a]. *)
 
   val create : ?label:string -> unit -> 'a t * 'a u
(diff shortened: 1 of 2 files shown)


talex5 (commented):
Looks good to me - how about doing the resolvers too?

lib_eio/core/eio__core.mli
```diff
@@ -126,7 +126,7 @@ module Promise : sig
           (fun () -> Promise.resolve resolver 42)
       ]} *)
 
-  type !'a t
+  type +'a t
   (** An ['a t] is a promise for a value of type ['a]. *)
 
   type 'a u
```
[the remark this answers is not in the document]
anmonteiro: done in [90abf80](https://github.com/ocaml-multicore/eio/pull/300/commits/90abf80879199b1dfd7fe9660f302f49a2c2eb1f)

lib_eio/core/eio__core.mli
```diff
@@ -126,10 +126,10 @@ module Promise : sig
           (fun () -> Promise.resolve resolver 42)
       ]} *)
 
-  type !'a t
+  type +'a t
```
talex5, line 129:
```suggestion
  type +!'a t
```

lib_eio/core/eio__core.mli
```diff
@@ -126,10 +126,10 @@ module Promise : sig
...
   (** An ['a t] is a promise for a value of type ['a]. *)
 
-  type 'a u
+  type -'a u
```
talex5, line 132:
```suggestion
  type -!'a u
```

talex5 (commented):
Looks good to me! We should probably re-add the injectivity annotations, though.

(I don't recall why we needed them - @bikallem added them in #82 - but they shouldn't do any harm)

lib_eio/core/eio__core.mli
bikallem: IIRC I added them then since compilation was failing in the case below:
> type _ eff += Fork : (unit -> 'a) -> 'a Promise.t eff

Does it not fail anymore? I believe we still need `a Promise.t` to be injective for it to be usable in the above use-case (i.e. non-existential GADT usage).

anmonteiro:
Re-added the injectivity annotations.

talex5 (approved):
Thanks!

PR #302, opened 2022-09-02
net: add Net.with_tcp_connect

bikallem:
`Net.with_tcp_connect` takes a domain-name/port/service and attempts to
establish connection to it. IP addresses for the host are tried one
after the other while preferring/trying IPv6 protocol addresses first
if available.

We also introduce two new exceptions:
1. Connection_failure - raised when a connection couldn't be established
  for any of the addresses defined for a given host/port/service.
2. Unix_error - introduced to mimic Unix.Unix_error without depending
  on "unix" library in the "lib_eio" which is (should?) be free from "unix" lib dependency if I understand the design/architecture of "eio". 

**Open Issues**
Ideally, `Eio.Net.Unix_error` is an implementation level exception which could be hidden from lib `eio` users. Question to maintainers - should this exn be hidden? At the moment there is `net.mli` file which prevents this from being hidden at the `eio` level.

### DIFF
--- a/lib_eio/net.ml
+++ b/lib_eio/net.ml
@@ -2,9 +2,6 @@ exception Connection_reset of exn
 (** This is a wrapper for EPIPE, ECONNRESET and similar errors.
     It indicates that the flow has failed, and data may have been lost. *)
 
-exception Unix_error of string * string * string
-(* Unix.Unix_error without depending on "unix" *)
-
 exception Connection_failure of exn
 
 module Ipaddr = struct
@@ -213,35 +210,23 @@ let getnameinfo (t:#t) sockaddr = t#getnameinfo sockaddr
 
 let close = Flow.close
 
-type 'a timeout = (#Time.clock as 'a) * float
-
-let rec try_connection connect_fn = function
-  | [] -> None
-  | addr::addrs -> (
-    match connect_fn addr with
-    | conn -> Some conn
-    | exception Unix_error(_,_,_)
-    | exception Time.Timeout -> try_connection connect_fn addrs)
-
-let with_tcp_connect ?timeout ~host ~service t f =
-  let addrs =
-    getaddrinfo_stream ~service t host
-    |> List.filter_map (function
-        | `Tcp _ as x -> Some x
-        | `Unix _ -> None)
+let with_tcp_connect ?(timeout=Time.Timeout.none) ~host ~service t f =
+  Switch.run @@ fun sw ->
+  let rec aux = function
+    | [] -> raise (Connection_failure (Failure (Fmt.str "No TCP addresses for %S" host)))
+    | addr :: addrs ->
+      match Time.Timeout.run_exn timeout (fun () -> connect ~sw t addr) with
+      | conn -> f conn
+      | exception (Time.Timeout | Connection_failure _) when addrs <> [] ->
+        aux addrs
+      | exception (Connection_failure _ as ex) ->
+        raise ex
+      | exception (Time.Timeout as ex) ->
+        raise (Connection_failure ex)
   in
-  Switch.run (fun sw ->
-    let conn =
-      match timeout with
-      | Some (clock, d) ->
-        let connect_fn addr = Time.with_timeout_exn clock d (fun () -> connect ~sw t addr) in
-        try_connection connect_fn addrs
-      | None ->
-        let connect_fn addr = connect ~sw t addr in
-        try_connection connect_fn addrs
-    in
-    match conn with
-    | Some conn -> f conn
-    | None ->
-      let exn = Failure (Printf.sprintf "Unable to connect to host:%s, service:%s" host service) in
-      raise @@ Connection_failure exn)
+  getaddrinfo_stream ~service t host
+  |> List.filter_map (function
+      | `Tcp _ as x -> Some x
+      | `Unix _ -> None
+    )
+  |> aux
--- a/lib_eio/net.ml
+++ b/lib_eio/net.ml
@@ -2,6 +2,10 @@ exception Connection_reset of exn
 (** This is a wrapper for EPIPE, ECONNRESET and similar errors.
     It indicates that the flow has failed, and data may have been lost. *)
 
+exception Unix_error of string * string * string
+(* Unix.Unix_error without depending on "unix" *)
+
+exception Connection_failure of exn
 
 module Ipaddr = struct
   type 'a t = string   (* = [Unix.inet_addr], but avoid a Unix dependency here *)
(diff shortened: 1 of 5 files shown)


lib_eio/net.ml
```diff
@@ -208,3 +212,26 @@ let getaddrinfo_datagram ?service t hostname =
...
+    | exception Unix_error(_,_,_)
+    | exception Time.Timeout -> try_connection env sw timeout host service addrs)
+
+let with_tcp_connect ?(timeout=0.1) ~host ~service env f =
+  let addrs =
+    getaddrinfo_stream ~service env#net host
+    |> List.filter_map (function
+        | `Tcp (x, port) -> Some (Ipaddr.fold ~v4:(fun ip -> `V4, ip, port) ~v6:(fun ip -> `V6, ip, port) x)
+        | `Unix _ -> None)
+    |> List.sort ( fun (a,_,_) (b,_,_) ->
```
talex5, line 230: `getaddrinfo_stream` returns the results in the preferred order. Sorting them defeats that.

lib_eio/net.mli
```diff
@@ -12,6 +12,8 @@
 *)
 
 exception Connection_reset of exn
+exception Unix_error of string * string * string
+exception Connection_failure of (string * string)
```
talex5, line 16: No need for an extra tuple here.
```suggestion
exception Connection_failure of string * string
```
But what are these strings, anyway?

tests/network.md
```diff
@@ -546,3 +546,19 @@ EPIPE:
   Eio.Net.getnameinfo env#net sockaddr;;
 - : string * string = ("localhost", "http")
 ```
+
+## with_tcp_connet
+
+<!-- $MDX non-deterministic=command -->
```
talex5, line 552: You can pass in a mock network and make this test actually run. We already tested that DNS lookups and connecting work, so we only need to test the logic of trying them in sequence here.

lib_eio/net.mli
```diff
@@ -113,7 +115,32 @@ end
...
-    The new socket will be closed when [sw] finishes, unless closed manually first. *)
+    The new socket will be closed when [sw] finishes, unless closed manually first.
+
+    @raise Unix_error if connection couldn't be established. *)
+
+val with_tcp_connect :
+  ?timeout:float ->
+  host:string ->
+  service:string ->
+  <clock: #Time.clock; net: #t; ..> ->
```
talex5, line 126: We should probably define `type timeout = Clock.t * float` or similar, and then do:
```suggestion
val with_tcp_connect :
  ?timeout:Timeout.t ->
  t ->
  host:string ->
  service:string ->
```

talex5 (commented):
> Ideally, Eio.Net.Unix_error is an implementation level exception which could be hidden from lib eio users. Question to maintainers - should this exn be hidden? At the moment there is net.mli file which prevents this from being hidden at the eio level.

It can't be hidden because other people need to use it (either to create custom networks that raise it, or custom connection logic that catches it). You could just use `Connection_failure` for this too.

The usual pattern in Eio is:
```ocaml
exception Connection_failure of exn
```
Then the full details (the `Unix_error` or whatever) are available for logging, etc.

lib_eio/net.mli
bikallem: `(host, port)`

lib_eio/net.ml
bikallem: IIUC `getaddrinfo_stream` doesn't sort the `Eio.Net.Ipaddr.t` in the correct order, i.e. `V6` address first. The following snippet illustrates this:
```
Eio_main.run @@ fun env ->
Eio.Net.getaddrinfo_stream ~service:"http" env#net "www.example.com"
|> List.iter (function
  | `Tcp (x, port) -> Eio.Net.Ipaddr.fold ~v4:(fun ip -> print_string "V4\n") ~v6:(fun ip -> print_string "V6\n") x
  | `Unix _ -> ());;
```
The above prints:
```
V4
V6
```
Additionally, `getaddrinfo_stream` doesn't filter out the `Unix _ ` bit. So the code takes care of that too.

tests/network.md
bikallem: It seems we can mock this just yet, since we also need to be able to mock `clock` due to - I think - usage of `Eio.Time.timeout` function.

lib_eio/net.mli
bikallem: If we make clock part of `timeout` and then make `timeout` an optional parameter, does this mean we don't want connection timeout? i.e. if a user didn't use `timeout` parameter, we have a default value of `0.1` for timeout seconds but what do we use as default clock value?

If no timeout is used while connecting to the a host, wouldn't that degrade user experience of the API?

lib_eio/net.mli
bikallem: The exception is `Connection_failure of exn` now.

lib_eio/net.ml
talex5: getaddrinfo(3) says:

> Normally, the application should try using the addresses in  the  order  in  which they are returned.

So if V4 addresses are coming first, then presumably that's what that system wants us to try first.

https://www.ietf.org/rfc/rfc3484.txt says:
> Well-behaved applications SHOULD iterate through the list of
>  addresses returned from getaddrinfo() until they find a working
>  address.

lib_eio/net.mli
talex5: Generally I prefer to avoid default timeouts and let the calling code handle that. We have no idea whether the network will require more or less than 0.1s to make a connection.

There's also the [Happy Eyeballs](https://datatracker.ietf.org/doc/html/rfc8305) protocol, which tries IPv4 and IPv6 in parallel. However, user applications can use a library (e.g. https://github.com/roburio/happy-eyeballs) for that.

lib_eio/net.ml
bikallem: Ah, thanks for the reference RFC. I see what you mean, now. Since, we are using `getaddrinfo` behind the scenes it makes sense to follow what the RFC says. Accordingly, I will remove the sorting functionality.

Just FYI, the mechanism adopted by the PR is more in line with the happy eyeballs RFC (https://www.rfc-editor.org/rfc/rfc8305). The RFC seems to suggest for the preference of IPv6 as opposed to IPv4 due to the former becoming more prevalent.

lib_eio/net.mli
bikallem: >There's also the [Happy Eyeballs](https://datatracker.ietf.org/doc/html/rfc8305) protocol, which tries IPv4 and >IPv6 in parallel. However, user applications can use a library (e.g. https://github.com/roburio/happy-eyeballs) >for that.

Yes, indeed this PR was in part influenced by the Happy Eyeballs RFC in terms of IP version selection and the timeout requirements.

bikallem: Done.

lib_eio/net.ml
bikallem: Removed the sorting. It now uses the order returned by `getaddrinfo`.

tests/network.md
talex5: I've added a mock clock now in #328.

talex5 (approved):
I rebased this on `main` and updated it to use the new `Eio.Time.Timeout` and `Eio_mock.Clock` modules. The tests no longer require a real network and so can always be run.

I also removed the `Net.Unix_error` exception (using `Connection_failure` instead) to make it more generic, and made `with_tcp_connect` pass along the (final) exception from the network, so you can see why it failed.

If you're happy with that, it's ready to merge.

bikallem:
Thanks. I did a pass-through and the changes look good to me. 👍

talex5:
Thanks!

PR #306, opened 2022-09-04
io-uring needs kernel version >= 5.10, not 5.11

### LINKED ISSUE #304: `eio_main` selects `luv` by default on WSL
What the title says. I tried with `io-uring` by setting the `EIO_BACKEND` variable and it seems to work fine.

nojb:
Fixes #304

### DIFF
--- a/lib_main/eio_main.linux.ml
+++ b/lib_main/eio_main.linux.ml
@@ -1,6 +1,6 @@
 let has_working_uring v =
   match String.split_on_char '.' v with
-  | "5" :: minor :: _ -> int_of_string minor >= 11
+  | "5" :: minor :: _ -> int_of_string minor >= 10
   | major :: _ -> int_of_string major > 5
   | [] -> false
 


talex5:
Linux 5.11 is needed for unlink (see #289).

If Windows is still on 5.10, we could bring the old fallback code back, and use the probing support (added in uring 0.4) to decide whether to use it.

nojb:
> Linux 5.11 is needed for unlink (see #289).

Thanks for the pointer.

> If Windows is still on 5.10, we could bring the old fallback code back, and use the probing support (added in uring 0.4) to decide whether to use it.

Yes, WSL is still on 5.10, but 5.15 should be released soon-ish (it is currently on "prerelease"), so I'm not sure it is worth the trouble to add a fallback.

But I would amend the error message for clarity: https://github.com/ocaml-multicore/eio/blob/355478d6d9ae96bfdb06055fcb2e3cd800013c4c/lib_main/eio_main.linux.ml#L29

Issue #305, opened 2022-09-04
What to do for `ftruncate`?


nojb:
Is this function missing from `eio`, or is one supposed to use the one in the `unix` library (via `eio_unix`)?

nojb:
Also: "seek".

talex5:
There are lots of missing functions - PRs welcome!

talex5:
`seek` and `truncate` were added in https://github.com/ocaml-multicore/eio/pull/626.

PR #307, opened 2022-09-04
Add API for seekable read/writes

nojb:
This partly addresses #305. Currently just a draft, meant for discussing the API.

The PR adds a method `#read_at_offset` to the object returned by `open_in` and `with_open_in` allowing to do a read at a given offset. The new method is exposed as `Eio.Fs.read_at_offset` and a helper function is added as well `Eio.Fs.read_exact_at_offset`. Not sure if this is the right level at which to expose this functionality.

I'm just getting started with this library and am not sure I have yet grokked the full picture, so all feedback is welcome!

### DIFF
--- a/lib_eio/fs.ml
+++ b/lib_eio/fs.ml
@@ -12,17 +12,6 @@ exception Already_exists of path * exn
 exception Not_found of path * exn
 exception Permission_denied of path * exn
 
-class virtual ro = object (_ : <Generic.t; Flow.source; ..>)
-  method virtual pread : file_offset:Optint.Int63.t -> Cstruct.t list -> int
-end
-
-class virtual rw = object (_ : <Generic.t; Flow.source; Flow.sink; ..>)
-  inherit ro
-  method virtual pwrite : file_offset:Optint.Int63.t -> Cstruct.t list -> int
-  method probe _ = None
-  method read_methods = []
-end
-
 (** When to create a new file. *)
 type create = [
   | `Never                            (** fail if the named file doesn't exist *)
@@ -35,12 +24,12 @@ type create = [
 (** Note: use the functions in {!Path} to access directories. *)
 class virtual dir = object (_ : #Generic.t)
   method probe _ = None
-  method virtual open_in : sw:Switch.t -> path -> <ro; Flow.close>
+  method virtual open_in : sw:Switch.t -> path -> <File.ro; Flow.close>
   method virtual open_out :
     sw:Switch.t ->
     append:bool ->
     create:create ->
-    path -> <rw; Flow.close>
+    path -> <File.rw; Flow.close>
   method virtual mkdir : perm:Unix_perm.t -> path -> unit
   method virtual open_dir : sw:Switch.t -> path -> dir_with_close
   method virtual read_dir : path -> string list
@@ -54,27 +43,3 @@ and virtual dir_with_close = object
   inherit dir
   method virtual close : unit
 end
-
-let pread (t : #ro) ~file_offset bufs =
-  let got = t#pread ~file_offset bufs in
-  assert (got > 0 && got <= Cstruct.lenv bufs);
-  got
-
-let rec pread_exact (t : #ro) ~file_offset bufs =
-  if Cstruct.lenv bufs > 0 then (
-    let got = t#pread ~file_offset bufs in
-    let file_offset = Optint.Int63.add file_offset (Optint.Int63.of_int got) in
-    pread_exact t ~file_offset (Cstruct.shiftv bufs got)
-  )
-
-let pwrite (t : #rw) ~file_offset bufs =
-  let got = t#pwrite ~file_offset bufs in
-  assert (got > 0 && got <= Cstruct.lenv bufs);
-  got
-
-let rec pwrite_exact (t : #rw) ~file_offset bufs =
-  if Cstruct.lenv bufs > 0 then (
-    let got = t#pwrite ~file_offset bufs in
-    let file_offset = Optint.Int63.add file_offset (Optint.Int63.of_int got) in
-    pwrite_exact t ~file_offset (Cstruct.shiftv bufs got)
-  )
--- a/lib_eio/fs.ml
+++ b/lib_eio/fs.ml
@@ -12,7 +12,13 @@ exception Already_exists of path * exn
 exception Not_found of path * exn
 exception Permission_denied of path * exn
 
+class virtual ro = object (_ : <Generic.t; Flow.source; ..>)
+  method virtual pread : file_offset:Optint.Int63.t -> Cstruct.t list -> int
+end
+
 class virtual rw = object (_ : <Generic.t; Flow.source; Flow.sink; ..>)
+  inherit ro
+  method virtual pwrite : file_offset:Optint.Int63.t -> Cstruct.t list -> int
   method probe _ = None
   method read_methods = []
 end
@@ -29,7 +35,7 @@ type create = [
 (** Note: use the functions in {!Path} to access directories. *)
 class virtual dir = object (_ : #Generic.t)
   method probe _ = None
-  method virtual open_in : sw:Switch.t -> path -> <Flow.source; Flow.close>
+  method virtual open_in : sw:Switch.t -> path -> <ro; Flow.close>
   method virtual open_out :
     sw:Switch.t ->
     append:bool ->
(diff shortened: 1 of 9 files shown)


lib_eio/fs.ml
```diff
@@ -12,6 +12,10 @@ exception Already_exists of path * exn
 exception Not_found of path * exn
 exception Permission_denied of path * exn
 
+class virtual read_at_offset = object
```
talex5, line 15: Might as well make a `ro` type for this and related operations. We already have `rw` type.
```suggestion
class virtual ro = object (_ : <Generic.t; Flow.source; ..>)
```

lib_eio/fs.ml
```diff
@@ -29,7 +33,7 @@ type create = [
 (** Note: use the functions in {!Path} to access directories. *)
 class virtual dir = object (_ : #Generic.t)
   method probe _ = None
-  method virtual open_in : sw:Switch.t -> path -> <Flow.source; Flow.close>
+  method virtual open_in : sw:Switch.t -> path -> <Flow.source; Flow.close; read_at_offset>
```
talex5, line 36:
```suggestion
  method virtual open_in : sw:Switch.t -> path -> <ro; Flow.close>
```

lib_eio/fs.ml
```diff
@@ -12,6 +12,10 @@ exception Already_exists of path * exn
 exception Not_found of path * exn
 exception Permission_denied of path * exn
 
+class virtual read_at_offset = object
+  method virtual read_at_offset : int64 -> Cstruct.t -> int
```
talex5, line 16: I don't particularly like the name, but POSIX and Lwt_unix both call this `pread`.

Elsewhere, we're using `Int63.t` as the type for offsets, to avoid an allocation.

lib_eio/fs.ml
```diff
@@ -12,6 +12,10 @@ exception Already_exists of path * exn
...
+end
+
 class virtual rw = object (_ : <Generic.t; Flow.source; Flow.sink; ..>)
```
talex5, line 19:
```suggestion
class virtual rw = object (_ : <ro; Flow.sink; ..>)
```

lib_eio/fs.ml
```diff
@@ -48,3 +52,15 @@ and virtual dir_with_close = object
   inherit dir
   method virtual close : unit
 end
+
+let read_at_offset (t : #read_at_offset) ofs buf =
```
talex5, line 56: Could just call this `read` and take an optional argument. `eio_linux.mli` uses `file_offset`, e.g.
```suggestion
let read (t : #ro) ?file_offset buf =
```

talex5 (commented):
Thanks!  I added some suggestions inline (I'm not quite sure what the best API would be either).

lib_eio/fs.ml
nojb: Thanks, done.

lib_eio/fs.ml
nojb: Fixed.

lib_eio/fs.ml
nojb: I renamed to `pread`, and switched to `Int63.t`.

lib_eio/fs.ml
nojb: Fixed.

lib_eio/fs.ml
nojb: Here I was in two minds because there already is a `Flow.read` which corresponds to the case of `file_offset = None`. So for the moment I renamed this to `pread`, with a *mandatory* `~file_offset` argument (to avoid duplicating the functionality of `Flow.read` which may be a bit confusing).

nojb:
Thanks for the review @talex5. I amended the PR roughly as suggested, and added an analogous "pwrite" API. Let me know what you think. Thanks!

talex5 (approved):
Looks good to me! Can be merged once squashed and no longer marked as draft.

Two things I wonder about:

- I see the linux and luv backends both accept a list of buffers. Maybe that should be exposed to users in the cross-platform API too?
- Both backend APIs use `write` and `write_single`, whereas the new cross-platform API uses `write_exact` and `write`. (OCaml itself has `Unix.single_write`, but I dislike that word order because it's hard to find when using tab-completion).

nojb:
> * I see the linux and luv backends both accept a list of buffers. Maybe that should be exposed to users in the cross-platform API too?

Makes sense, will amend.

> * Both backend APIs use `write` and `write_single`, whereas the new cross-platform API uses `write_exact` and `write`. (OCaml itself has `Unix.single_write`, but I dislike that word order because it's hard to find when using tab-completion).

Since `read` is already used by `flow` I kept the `p` prefix: `pread` and `pwrite`. Does `pread_single`/`pread` and `pwrite_single`/`pwrite` sound good for the new functions?

nojb:
> > Since read is already used by flow I kept the p prefix: pread and pwrite. Does pread_single/pread and pwrite_single/pwrite sound good for the new functions?
> 
> I'm a bit nervous about `pread`. Short reads are useful at the end of a file and it might be unexpected if that's an error. Perhaps it's best to be explicit and have `p{read,write}_{all,single}`.

Sorry, FWIW I just remembered why I had used `pread` and `pread_exact` originally: it matches the terminology in `flow`'s `read` and `read_exact`.

Anyway, I don't have a strong opinion either way: `pread`/`pread_exact` or `pread_single`/`pread_all`. Let me know which one strikes you as preferable...

talex5:
> Sorry, FWIW I just remembered why I had used `pread` and `pread_exact` originally: it matches the terminology in flow's `read` and `read_exact`.

That seems reasonable. Let's use `pread`/`pread_exact` for now then, and we can rename to `p?read_single` in both places in a future PR, if we decide that's better.

> Anyway, I don't have a strong opinion either way: pread/pread_exact or pread_single/pread_all. Let me know which one strikes you as preferable...

`pread_all` sounds like it might mean to read all of the file.

nojb:
> That seems reasonable.

Perfect, I squashed the PR, and exposed the possibility of passing a list of buffers, as suggested. This is ready for merging on my side. Thanks!

talex5 (commented):
I pushed a commit with the suggested changes.

I also moved the file operations out to a new `Eio.File` module, added some doc-strings, and renamed the write operations slightly.

If you're happy with that, it can be merged.

nojb:
> If you're happy with that, it can be merged.

LGTM, thanks.

lib_eio_luv/eio_luv.ml
```diff
@@ -548,6 +552,20 @@ let flow fd = object (_ : <source; sink; ..>)
...
+    match File.read ~file_offset fd bufs |> or_raise |> Unsigned.Size_t.to_int with
+    | 0 -> raise End_of_file
+    | got -> got
+
+  method pwrite ~file_offset bufs =
+    let bufs = List.map Cstruct.to_bigarray bufs in
+    let file_offset = Optint.Int63.to_int64 file_offset in
+    match File.write_single ~file_offset fd bufs |> or_raise |> Unsigned.Size_t.to_int with
+    | 0 -> raise End_of_file
+    | got -> got
```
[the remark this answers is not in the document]
nojb: Good catch.

lib_eio/fs.ml
```diff
@@ -12,7 +12,13 @@ exception Already_exists of path * exn
...
 
+class virtual ro = object (_ : <Generic.t; Flow.source; ..>)
+  method virtual pread : file_offset:Optint.Int63.t -> Cstruct.t list -> int
+end
+
 class virtual rw = object (_ : <Generic.t; Flow.source; Flow.sink; ..>)
+  inherit ro
+  method virtual pwrite : file_offset:Optint.Int63.t -> Cstruct.t list -> int
   method probe _ = None
   method read_methods = []
```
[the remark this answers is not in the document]
haesbaert: For consistency, `File.write` is binding the result with `or_raise` (t -> Luv.Buffer.t list -> unit), shouldn't we make `File.write_single` the same ?

haesbaert:
For consistency, `File.write` is binding the result with `or_raise (t -> Luv.Buffer.t list -> unit)`, shouldn't we make `File.write_single` the same ?

talex5:
> For consistency, File.write is binding the result with or_raise (t -> Luv.Buffer.t list -> unit), shouldn't we make File.write_single the same ?

Good point. I pushed a commit making `write` return a result, to match the rest of the API, and also exposed `write_single`.

PR #309, opened 2022-09-07
Buf_read: Add take_while1 and skip_while1

bikallem:
This PR adds support for `take_while1` and `skip_while1` parsers. 

The parsers are equivalent of `take_while` and `skip_while` respectively. However, they fail if at least one token is not consumed by the parser. The parsers implement an equivalent of "at least one or more" token parsing needs.


talex5 (approved):
Thanks!

PR #310, opened 2022-09-07
Use ocamlformat

samoht:
Just opening the PR to start the discussion :-)

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,0 +1,5 @@
+version = 0.24.1
+profile = conventional
+
+break-infix = fit-or-vertical
+parse-docstrings = true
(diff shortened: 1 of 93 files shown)


talex5:
If you're bored and want to spend more time with ocamlformat, I'd suggest helping out the ocaml-cohttp project, whose [CI](https://github.com/mirage/ocaml-cohttp/runs/8208141393) is (as usual) broken by an ocamlformat problem (and whose [.ocamlformat history](https://github.com/mirage/ocaml-cohttp/commits/master/.ocamlformat) tells a sad tale of frustration).

Anyway, Eio is deliberately not using it :-)

I would, however, be happy with a CI change to ensure indentation matches what ocp-indent would do, or that there is no trailing whitespace on lines in OCaml code. But removing all the formatting and force-resetting everything to the default AST is not wanted.

Issue #312, opened 2022-09-08
Network tests fail on libuv + macOS


samoht:
Running the tests on macOS fails with:

```diff
❯ dune test
File "tests/network.md", line 1, characters 0-0:
diff --git a/_build/default/tests/network.md b/_build/default/tests/.mdx/network.md.corrected
index d2a08c4c69..df05279a8d 100644
--- a/_build/default/tests/network.md
+++ b/_build/default/tests/.mdx/network.md.corrected
@@ -451,8 +451,7 @@ ECONNRESET:
     ignore (Eio.Flow.read a (Cstruct.create 1) : int);
     assert false
   with Eio.Net.Connection_reset _ -> traceln "Connection failed (good)";;
-+Connection failed (good)
-- : unit = ()
+Exception: End_of_file.
 \```

 EPIPE:
```

haesbaert:
As tested yesterday, this seems to be normal behaviour on most stacks except linux.
Some how linux sees the close before write.

```C
#include <sys/socket.h>

#include <err.h>
#include <stdio.h>
#include <unistd.h>

int
main(void)
{
	int sockets[2];
	ssize_t n;
	char buf[1];

	if (socketpair(AF_UNIX, SOCK_STREAM, 0, sockets) == -1)
		err(1, "socketpair");

	n = write(sockets[0], "1", 1);
	if (n == -1)
		err(1, "write");
	else if (n == 0)
		errx(1, "write EOF");

	sleep(1);
	close(sockets[1]);
	sleep(1);

	n = read(sockets[0], buf, sizeof(buf));
	if (n == -1)
		err(1, "read");
	else if (n == 0)
		errx(1, "read EOF");
	
	return (0);
}
```
```
# Linux
sam:tmp: cc -o socketpair socketpair.c -Wall
sam:tmp: ./socketpair
socketpair: read: Connection reset by peer

# FreeBSD 12
[haesbaert@shell ~]$ cc -o socketpair socketpair.c -Wall
[haesbaert@shell ~]$ ./socketpair
socketpair: read EOF

# OpenBSD 7.2
naiad:haesbaert: ./socketpair
socketpair: read EOF

# macos 12.5.1
❯ cc -o socketpair socketpair.c -Wall
❯ ./socketpair
socketpair: read EOF
```
So I think we should just relax the test and accept EOF as well.

PR #313, opened 2022-09-08
Account for stack differences in the socketpair test (issue #312)

haesbaert:
Linux returns ECONNRESET while every other stack tested returns EOF on this test.

I did not dig into the especification but Linux appears to be on the weird side.

In order to get a ECONNRESET, the write needs to hit the socket _after_ the close, but our program does:

WRITE(A)
CLOSE(B)
READ(A)

Linux never gives "A" side a chance to see a proper EOF, while the others do:

Linux        -> ECONNRESET
FreeBSD 12   -> EOF
OpenBSD 7.2  -> EOF
macos 12.5.1 -> EOF

We can fix this by properly draining B side of things before the close, but we can't reliably test ECONNRESET so turn this into an EOF test for now.

### DIFF
--- a/tests/network.md
+++ b/tests/network.md
@@ -450,7 +450,7 @@ ECONNRESET:
   try
     ignore (Eio.Flow.read a (Cstruct.create 1) : int);
     assert false
-  with Eio.Net.Connection_reset _ -> traceln "Connection failed (good)";;
+  with Eio.Net.Connection_reset _ | End_of_file -> traceln "Connection failed (good)";;
 +Connection failed (good)
 - : unit = ()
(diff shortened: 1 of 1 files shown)


talex5:
The test is specifically trying to test for this error, though (i.e. where data may have been lost). I found some more info at https://stackoverflow.com/questions/2974021/what-does-econnreset-mean-in-the-context-of-an-af-local-socket

I suggest performing the same operations as before, but adding `| End_of_file` to allow that outcome too for non-Linux systems.

talex5 (approved):
Thanks!

PR #315, opened 2022-09-11
Move list functions into a dedicated submodule

raphael-proust:
As the title says: this PR introduces a submodule of `Fiber` called `List` to collect the list-related functions.

- The name `Eio.Fiber.map` could suggest mapping fibers.
- Lists are centrals but Seqs are also becoming important so having List functions unqualified in the main namespace may not be the best.
- At some point there'll be contributions to add `rev_map`, `mapi`, `filteri`, etc. and the main namespace would get bloated quickly.

### DIFF
--- a/lib_eio/core/eio__core.mli
+++ b/lib_eio/core/eio__core.mli
@@ -265,8 +265,15 @@ module Fiber : sig
       The current task remains runnable, but goes to the back of the queue.
       Automatically calls {!check} just before resuming. *)
 
+  (** Concurrent list operations. *)
   module List : sig
-    (** {2 Concurrent list operations} *)
+    (** These functions behave like the ones in the standard library's [List]
+        module, except that multiple items can be processed concurrently.
+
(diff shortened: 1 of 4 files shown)


lib_eio/core/eio__core.mli
```diff
@@ -265,27 +265,29 @@ module Fiber : sig
       The current task remains runnable, but goes to the back of the queue.
       Automatically calls {!check} just before resuming. *)
 
-  (** {2 Concurrent list operations} *)
-
-  val filter : ?max_fibers:int -> ('a -> bool) -> 'a list -> 'a list
```
talex5, line 270: Would be nice to mark these as deprecated first, with a clue where they went, for one release at least.

talex5 (commented):
Seems reasonable to me.

lib_eio/core/eio__core.mli
raphael-proust: done

talex5 (approved):
Thanks! I just rebased and added a bit of text to the odoc.

PR #317, opened 2022-09-16
direct implementation of Flow.source_string

c-cube:
this loses the `Read_source_buffer` read method, but removes the intermediate allocation + copy of a cstruct.


talex5 (approved):
Looks good. I force-pushed a change to restore the missing tests, now checking `cstruct_source` directly.

PR #318, opened 2022-09-17
Implement Flow.write and write_from

haesbaert:
These are the corollary from Flow.read and read_into.

-- 

Not super sure about `write_from` naming, but it's the closest to the opposite of `read_into` I can think of.

### DIFF
--- a/lib_eio/flow.ml
+++ b/lib_eio/flow.ml
@@ -70,10 +70,13 @@ let string_source s : source =
       len
   end
 
-class virtual sink = object (_ : <Generic.t; ..>)
+class virtual sink = object (self : <Generic.t; ..>)
   method probe _ = None
   method virtual copy : 'a. (#source as 'a) -> unit
+  method write bufs = self#copy (cstruct_source bufs)
 end
+
+let write (t : #sink) (bufs : Cstruct.t list) = t#write bufs
 
 let copy (src : #source) (dst : #sink) = dst#copy src
 
@@ -91,10 +94,13 @@ let buffer_sink b =
           Buffer.add_string b (Cstruct.to_string ~len:got buf)
         done
       with End_of_file -> ()
+
+    method! write bufs =
+      List.iter (fun buf -> Buffer.add_bytes b (Cstruct.to_bytes buf)) bufs
   end
 
 class virtual two_way = object (_ : <source; sink; ..>)
-  method probe _ = None
+  inherit sink
   method read_methods = []
 
   method virtual shutdown : shutdown_command -> unit
--- a/lib_eio/flow.mli
+++ b/lib_eio/flow.mli
@@ -62,7 +62,16 @@ type read_method += Read_source_buffer of ((Cstruct.t list -> int) -> unit)
 class virtual sink : object
   inherit Generic.t
   method virtual copy : 'a. (#source as 'a) -> unit
+
+  method write : Cstruct.t list -> unit
+  (** The default implementation is [copy (cstruct_source ...)], but it can be overridden for speed. *)
 end
+
+val write : #sink -> Cstruct.t list -> unit
+(** [write dst bufs] writes all bytes from [bufs].
+
+    This is a low level API, consider using {!copy} if possible as it
+    may allow optimizations. *)
 
 val copy : #source -> #sink -> unit
 (** [copy src dst] copies data from [src] to [dst] until end-of-file. *)
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -1044,6 +1044,8 @@ let flow fd =
 
     method read_methods = []
 
+    method write bufs = Low_level.writev fd bufs
+
     method copy src =
       match get_fd_opt src with
       | Some src -> fast_copy_try_splice src fd
--- a/lib_eio_luv/eio_luv.ml
+++ b/lib_eio_luv/eio_luv.ml
@@ -597,6 +597,10 @@ let flow fd = object (_ : <source; sink; ..>)
 
   method read_methods = []
 
+  method write bufs =
+    let bufs = List.map Cstruct.to_bigarray bufs in
+    File.write fd bufs |> or_raise
+
   method copy src =
     let buf = Luv.Buffer.create 4096 in
     try
@@ -623,6 +627,10 @@ let socket sock = object
   method read_into buf =
     let buf = Cstruct.to_bigarray buf in
     Stream.read_into sock buf
+
+  method! write bufs =
+    let bufs = List.map Cstruct.to_bigarray bufs in
+    Stream.write sock bufs
 
   method copy src =
     let buf = Luv.Buffer.create 4096 in
--- a/tests/flow.md
+++ b/tests/flow.md
@@ -107,3 +107,14 @@ Copying from src using `Read_source_buffer`:
 +dst: wrote (rsb) ["bar"]
 - : unit = ()
 ```
+
+## write
+
+```ocaml
+# run @@ fun () ->
+  let dst = Eio_mock.Flow.make "dst" in
+  Eio_mock.Flow.on_copy_bytes dst [`Return 6];
+  Eio.Flow.write dst [Cstruct.of_string "foobar"];;
++dst: wrote "foobar"
+- : unit = ()
+```
--- a/tests/network.md
+++ b/tests/network.md
@@ -695,3 +695,33 @@ Both attempts time out:
 +mock time is now 20
 Exception: Eio__Net.Connection_failure Eio__Time.Timeout.
 ```
+
+## read/write on SOCK_DGRAM
+
+```ocaml
+# Eio_main.run @@ fun _ ->
+  Switch.run @@ fun sw ->
+  let a, b = Eio_unix.socketpair ~sw ~domain:Unix.PF_UNIX ~ty:Unix.SOCK_DGRAM () in
+  ignore (Eio_unix.FD.peek a : Unix.file_descr);
+  ignore (Eio_unix.FD.peek b : Unix.file_descr);
+  let l = [ "foo"; "bar"; "foobar"; "cellar door" ] in
+  let buf = Cstruct.create 32 in
+  let write bufs = Eio.Flow.write a (List.map Cstruct.of_string bufs) in
+  let read () =
+    let n = Eio.Flow.read b buf in
+    traceln "Got: %d bytes: %S" n Cstruct.(to_string (sub buf 0 n))
+  in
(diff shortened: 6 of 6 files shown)


lib_eio/flow.ml
```diff
@@ -60,8 +60,11 @@ let string_source s = cstruct_source [Cstruct.of_string s]
 class virtual sink = object (_ : <Generic.t; ..>)
   method probe _ = None
   method virtual copy : 'a. (#source as 'a) -> unit
+  method virtual write_from : Cstruct.t -> unit
```
talex5, line 63: "write" on its own is fine here ("write buffer" makes sense in English, whereas "read buffer" doesn't).

I think it would also make sense to expose the lowest-level API here, like we do with `Eio.File.pwrite`. And perhaps take a list like that does? e.g.

```suggestion
  method virtual write : Cstruct.t list -> int
```

lib_eio/flow.ml
```diff
@@ -60,8 +60,11 @@ let string_source s = cstruct_source [Cstruct.of_string s]
...
 end
 
+let write (t : #sink) (buf : Cstruct.t) = t#write_from buf
```
talex5, line 66: With the change above, we can provide both `write_all` and `write_single` here, like in `File`.

lib_eio/flow.ml
```diff
@@ -78,6 +81,9 @@ let buffer_sink b =
           Buffer.add_string b (Cstruct.to_string ~len:got buf)
         done
       with End_of_file -> ()
+
+    method write_from buf =
+      Buffer.add_bytes b (Cstruct.to_bytes buf) (* XXX slow *)
```
talex5, line 86: I don't think there's a faster way to do it given the `Buffer` API. We can just note in the mli that this might not be very efficient (in fact, it already points you in the direction of `Buf_read`).

lib_eio/flow.mli
```diff
@@ -62,8 +62,12 @@ type read_method += Read_source_buffer of ((Cstruct.t list -> int) -> unit)
 class virtual sink : object
   inherit Generic.t
   method virtual copy : 'a. (#source as 'a) -> unit
+  method virtual write_from : Cstruct.t -> unit
 end
 
+val write : #sink -> Cstruct.t -> unit
+(** [write dst buf] writes one or more bytes from [buf] *)
```
talex5, line 69: Might be good to note here that this is a low-level API and that using `copy` may allow optimisations in some cases.

lib_eio/flow.ml
haesbaert: My first version had both, ha, I'll add it back.

haesbaert: The problem is libuv doesn't give us a way to do `write_single`, it will queue the next write and attempt to write all of them.
```
      int uv_write(uv_write_t *req, uv_stream_t *handle, const uv_buf_t bufs[], unsigned int nbufs, uv_write_cb cb)
              Write data to stream. Buffers are written in order. Example:

                 void cb(uv_write_t* req, int status) {
                     /* Logic which handles the write result */
                 }

                 uv_buf_t a[] = {
                     { .base = "1", .len = 1 },
                     { .base = "2", .len = 1 }
                 };

                 uv_buf_t b[] = {
                     { .base = "3", .len = 1 },
                     { .base = "4", .len = 1 }
                 };

                 uv_write_t req1;
                 uv_write_t req2;

                 /* writes "1234" */
                 uv_write(&req1, stream, a, 2, cb);
                 uv_write(&req2, stream, b, 2, cb);

              NOTE:
                 The memory pointed to by the buffers must remain valid until the callback gets called.  This also holds for uv_write2().
```

And no, it doesn't return bytes written, it's just 0 or error :/

haesbaert: So the option is to either lie (write_single = write_all) in libuv, and god knows how many writes were done, or not provide a write_single.

lib_eio/flow.ml
c-cube: Given that `Buffer` doesn't expose useful functions for that, it might be a good idea to also expose a better implementation of it (`{mutable buf: bytes;  mutable len: int }` basically), if only to get a `string` out of a `Flow.source` easily.

lib_eio/flow.ml
c-cube: should `copy` have a default implementation in terms of `write_from`?

talex5: libuv is callback based. I think the `return 0` just means the write operation was scheduled. The callback gets the number of bytes written.

https://github.com/aantron/luv/blob/master/src/stream.mli says:

> The second argument passed to the callback is the number of bytes
> written. libuv has an internal queue of writes, in part to implement retry.
> This means that writes can be partial at the libuv (and Luv) API level, so
> it is possible to receive both an [Error] result, and for some data to have
> been successfully written.

However, for files it looks like `write` behaves normally (either returns an error or a byte count, but not both).

So, yeah, maybe we shouldn't provide `write_single`. But then we don't need a write method at all - `Flow.write flow buf` can just be `Flow.copy (cstruct_source [buf]) flow`.

Possibly we just need to add `exception Write_error of int * exn` to report how many bytes got written, in case anyone cares.

haesbaert: > libuv is callback based. I think the `return 0` just means the write operation was scheduled. The callback gets the number of bytes written.
> 

Even though it's callback based, the code tries to do a write before setting up an event, so we can reason about that only:
`uv_write()->uv_write2()->empty_queue->uv__write()->uv__try_write()->uv__writev()->writev(2)`

> https://github.com/aantron/luv/blob/master/src/stream.mli says:
> 
> > The second argument passed to the callback is the number of bytes
> > written. libuv has an internal queue of writes, in part to implement retry.
> > This means that writes can be partial at the libuv (and Luv) API level, so
> > it is possible to receive both an [Error] result, and for some data to have
> > been successfully written.

This is a bit bad worded, what he means is that you can get an error _and_ you also get back how many bytes you successfully wrote before from previous writes, but the callback is only called once. There is no way to distinguish on success if the requested amount occurred in one or N calls.

With this I stand by my previous assessment (humbly), luv does not give us an API to do something in one `write` or give us control back in-between writes. In a sense they really treat a stream as a stream.

In fact, in order to retrieve the amount of partial bytes on write, he must be reaching into what libuv calls "private", cause the callback for write is not like read:
```
     typedef void (*uv_write_cb)(uv_write_t *req, int status)
              Callback called after data was written on a stream. status will be 0 in case of success, < 0 otherwise.
       typedef void (*uv_read_cb)(uv_stream_t *stream, ssize_t nread, const uv_buf_t *buf)
              Callback called when data was read on a stream.
```
http://docs.libuv.org/en/v1.x/stream.html#c.uv_write_cb

I can dig up luv later but he is probably getting from:
```
struct uv_write_s { /* haesbaert: this is write_cb_t without the useless typedef */
  UV_REQ_FIELDS
  uv_write_cb cb;
  uv_stream_t* send_handle; /* TODO: make private and unix-only in v2.x. */
  uv_stream_t* handle;
  UV_WRITE_PRIVATE_FIELDS
};
#define UV_WRITE_PRIVATE_FIELDS   
  void* queue[2]; 
  unsigned int write_index; /* haesbaert: this guy is likely the second argument to stream */
  uv_buf_t* bufs;
  unsigned int nbufs; 
  int error;
  uv_buf_t bufsml[4];

> 
> However, for files it looks like `write` behaves normally (either returns an error or a byte count, but not both).
> 
> So, yeah, maybe we shouldn't provide `write_single`. But then we don't need a write method at all - `Flow.write flow buf` can just be `Flow.copy (cstruct_source [buf]) flow`.
> 
> Possibly we just need to add `exception Write_error of int * exn` to report how many bytes got written, in case anyone cares.

I'd be against that, I like the idea of `write_all` + `write_single`, one means "I don't care how" and the other means "I need it to be in one".

Comment (haesbaert) [reply lib_eio/flow.ml:66]: Leme rephrase that, I'm fine with not having single + all, but having the Write_error is not necessary imho.

Comment (talex5) [reply lib_eio/flow.ml:66]: Yeah, no need for `Write_error` (or maybe `Partial_write`?) in this PR. I've never needed it, so we'll wait and see if anyone else does.

Note: `fallback_copy` won't be used if the source is a `cstruct_source` (it'll use `copy_with_rsb` instead, which behaves the way you want already). Though the luv backend doesn't do this optimisation yet.

I think it's fine to add a `write` method as an optimisation for this (would be interesting to know what the speed difference is).

File: lib_eio/flow.ml (hunk 3 above)
Comment (talex5) [reply lib_eio/flow.ml:86]: @avsm is thinking about changing Cstruct to be backed by `bytes` rather than bigarray. If that happens, `Eio.Buf_read` will work for this case (see https://github.com/ocaml-multicore/eio/issues/140).

Comment (haesbaert) [reply lib_eio/flow.ml:86]: > @avsm is thinking about changing Cstruct to be backed by `bytes` rather than bigarray. If that happens, `Eio.Buf_read` will work for this case (see #140).

Aye, I checked that, I've been thinking of implementing something like `Mbuf.t` which could be backed by a chain of Bytes or Cstruct, a bit inspired by how mbuf(9) works. With that we can guarantee that the backing Bytes is always > wosize so it goes directly in the major heap, and also, if needed, swap to Cstruct transparently in the future if things change in the GC.

It would be something like "allocate a fixed bytes size, say 256, start writing at say offset 63 so future prepends are free, appending a few words would just write to the existing chunk, appending a longer chunk would link it to the chain". Maybe have a big (>1500) and little ( < 512 ).

We could then make full usage of writing a backing method in Eio to turn that into the iovecs.

I don't think Cstruct or Bytes directly is a good way for people to build packets.

Comment (talex5) [reply lib_eio/flow.ml:86]: Note: we already have `Buf_write`, which is similar (though it doesn't allocate room for prepending). It does allow appending cstructs without copying, which results in using `writev` automatically.

File: lib_eio/mock/eio_mock.mli [hunk 5]
```diff
@@ -106,6 +107,9 @@ module Flow : sig
   val on_read : t -> string Handler.actions -> unit
   (** [on_read t actions] configures the values to return from the mock's [read] function. *)
 
+  val on_write : t -> unit Handler.actions -> unit
```
talex5, line 110: I don't think we need this. We can just use the default behaviour of calling `copy`.

talex5 (commented):
Looks good. How about taking a list (`writev`), since both backends already support that? I guess it's needed for datagram sockets, where writing a vector of two buffers isn't the same as doing two separate writes.

haesbaert:
> I squashed and rebased the commits, and also marked `write` as non-virtual in the interface.
> 
> If you're happy with that, it's ready for merging.

yep :smile:

PR #320, opened 2022-09-20
Add Eio.Time.Timeout

talex5:
This allows passing (optional) timeouts as arguments easily. For example, it should simplify the code in #302.

/cc @bikallem


bikallem:
Indeed, this PR should simplify #302 and general timeout usage in `eio`.

PR #324, opened 2022-09-23
Condition.broadcast must interlock as well

haesbaert:
There's a race where broadcast doesn't interlock and a in-flight waiter will have a lost wakeup:
```
CPU0 at:
    Eio.mutex_lock mutex;
    check_condition -> maybe abort
    Mutex.lock t.mutex; (inlined await for visibility)
B:  Eio_mutex.unlock mutex;
---> HERE
    match Waiters.await ~mutex:(Some t.mutex) t.waiters t.id with
    | ()           -> Eio_mutex.lock mutex
    | exception ex -> Eio_mutex.lock mutex; raise ex

CPU1 at:
    Eio.Mutex.use_rw ~protect:false mutex (fun () -> foo := true)
---> HERE
    Waiters.wake_all t.waiters (); (inlined broadcast for visibility)

1 - CPU0:HERE already tested the condition, it's false so it's ready to sleep,
    it interlocked with the runtime Mutex and released the Eio one.
2 - CPU1 sets the condition just after CPU0:B and now drains the queue and
    wakes the waiters, but CPU0 didn't make the waiting list yet, so it loses
    the wakeup.
3 - CPU0 waits forever.
```
The fix is to lock t.mutex, we don't need the full dance: LOCK(eio_mutex)->LOCK(mutex)->UNLOCK(eio_mutex), just locking the inner mutex is enough.

disclaimer: It's late and I might have missed something.

### DIFF
--- a/lib_eio/condition.ml
+++ b/lib_eio/condition.ml
@@ -22,4 +22,6 @@ let await_no_mutex t =
   Waiters.await ~mutex:(Some t.mutex) t.waiters t.id
 
 let broadcast t =
-  Waiters.wake_all t.waiters ()
+  Mutex.lock t.mutex;
+  Waiters.wake_all t.waiters ();
+  Mutex.unlock t.mutex


talex5 (approved):
The docs for `Waiters.wake_all` say:
> If [t] is shared between domains, the caller must hold the mutex while calling this.

So, yeah, the old code is obviously wrong.

Thanks!

PR #325, opened 2022-09-24
Expose `Eio.Fiber.Not_first`

anmonteiro:
Useful to match against in e.g.:

```ocaml
Fiber.first
 (fun () ->
   try 
     `Ok (operation_that_can_switch_fibers)
   with
   | Eio.Cancel.Cancelled Fiber.Not_first (* right now need to use `Eio__core__Fiber.Not_first` *) ->
     `Canceled)
 (fun () ->
    Promise.await some_promise;
   `Ok ())

```

### DIFF
--- a/lib_eio/core/eio__core.mli
+++ b/lib_eio/core/eio__core.mli
@@ -173,6 +173,10 @@ end
 
 (** @canonical Eio.Fiber *)
 module Fiber : sig
+  exception Not_first
+  (** Exception attached to [Cancel.Cancelled] when an operation is cancelled
+      in one of the racing functions described below. *)
+
   (** Within a domain, only one fiber can be running at a time.
       A fiber runs until it performs an IO operation (directly or indirectly).
       At that point, it may be suspended and the next fiber on the run queue runs. *)


talex5:
You should just match on `Eio.Cancel.Cancelled _`. You don't need to know why it was cancelled (it's just some extra info to help with debugging).

But there's no reason to catch it in this case anyway. Just have the second branch return `` `Cancelled`` directly:
```ocaml
Fiber.first
 (fun () -> `Ok (operation_that_can_switch_fibers))
 (fun () -> Promise.await some_promise; `Cancelled)
```

anmonteiro:
> You should just match on `Eio.Cancel.Cancelled _`. You don't need to know why it was cancelled (it's just some extra info to help with debugging).

makes sense, I'll do this.




> But there's no reason to catch it in this case anyway. Just have the second branch return `` `Cancelled `` directly:
> 
> ```ocaml
> Fiber.first
>  (fun () -> `Ok (operation_that_can_switch_fibers))
>  (fun () -> Promise.await some_promise; `Cancelled)
> ```

Doesn't this throw on the outside though?

I think the example I wanted to illustrate is:

```ocaml
Fiber.first
 (fun () -> try `Ok (operation_that_can_switch_fibers) with | Cancelled _ -> (* doesn't matter *))
 (fun () -> Promise.await some_promise; `Cancelled)
```

I thought that in my case if I don't add the `try ... with` it'd throw to the caller of `Fiber.first`, but now that I wrote this sentence I'm doubting myself.

anmonteiro:
It does indeed look like EIO will catch `Cancelled` for me in the fiber, and my mental model was wrong about `Fiber.any`, etc. Closing.

Issue #319, opened 2022-09-20
Hang in eio_linux tests


talex5:
@haesbaert [wrote](https://github.com/ocaml-multicore/eio/pull/311#issuecomment-1242692739):
> it seems to hang on lib_eio_linux/tests/test.ml on the very first test

We should try to find out what's causing this. I wrote:

> What happens if you just run `./_build/default/lib_eio_linux/tests/test.exe` in a loop? It doesn't hang for me.

haesbaert:
Sorry I had missed the original message, yes it hangs for me, rather fast:
```
RUN 579
Testing `eio_linux'.
This run has ID `26FSRR2T'.

  ...           io          0   copy.
  ```
  
with
```
sam:eio: local i=0;while true; do echo RUN $i; i=$((i+1)); ./_build/default/lib_eio_linux/tests/test.exe;done
```

**Comment by haesbaert** [comment]
I reduced the program that hangs as much as I could to this, I think my initial impression is correct, somehow the EOF is never seen by the reader.

```ocaml
let test_copy () =
  Eio_linux.run ~queue_depth:10 @@ fun _stdenv ->
  Eio.Switch.run @@ fun sw ->
  let from_pipe, to_pipe = Eio_linux.pipe sw in
  let buffer = Cstruct.create 20 in
  Eio.Flow.copy (Eio.Flow.string_source "a") to_pipe;
  Eio.Flow.close to_pipe;
  let () = 
    try 
      while true do
        ignore (Eio.Flow.read from_pipe buffer)
      done
  with End_of_file -> ()
  in
  Eio.Flow.close from_pipe
```
strace here: https://gist.github.com/haesbaert/437fd9e30e4568cc3f5ba95f0387d63a

Writer is FD6, which is actually closed during the hang (by looking at /proc/foo), FD5 (reader) is still opened and we are blocked in `io_uring_wait_cqe()`.

The pattern I see is that if the `close` happens before a `readv` is queued, _sometimes_ the `readv` will never see EOF. If the `readv` is submitted before the `close`, it always sees the EOF. I've tested this in two machines with slightly different kernels 5.18 vs 5.19, with released as well as current code base for eio and uring, behaviour is the same.

My only theory of why you can't trigger the bug is because on your tests the writer/reader dance terminates always in the order where the close only happens after the reader is queued, the order really depends on which CQE comes back first on the Fiber.both() tests. This program above should always trigger the bad case, I can hang it in < 5 seconds.

[...]

PR #327, opened 2022-09-26
Fix pipe hang (issue #319)

haesbaert:
Turns out the kernel needs to handle blocking file descriptors differently, it needs to do thread-pooling and it's a different code path (also slower).

It also seems that this is bugged in some cases. This makes sense as I believe most people don't test with blocking FDs. I believe we should make all our FDs non blocking as a precaution.

There is a program listed in issue #319 that replicates this. I could hang it in < 5 seconds, and it's now running for 10minutes with the fix, so kaboom.

### DIFF
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -953,7 +953,7 @@ let fast_copy src dst =
   with End_of_file -> ()
 
 (* Try a fast copy using splice. If the FDs don't support that, switch to copying. *)
-let fast_copy_try_splice src dst =
+let _fast_copy_try_splice src dst =
   try
     while true do
       let _ : int = Low_level.splice src ~dst ~len:max_int in
@@ -961,7 +961,11 @@ let fast_copy_try_splice src dst =
     done
   with
   | End_of_file -> ()
+  | Unix.Unix_error (Unix.EAGAIN, "splice", _) -> fast_copy src dst
   | Unix.Unix_error (Unix.EINVAL, "splice", _) -> fast_copy src dst
+
+(* XXX workaround for issue #319, PR #327 *)
+let fast_copy_try_splice src dst = fast_copy src dst
 
 (* Copy using the [Read_source_buffer] optimisation.
    Avoids a copy if the source already has the data. *)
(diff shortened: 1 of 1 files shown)


haesbaert:
There is a second hang in the `direct_copy` test, that is unrelated and I'm gonna have a look at it. I had it sporadically already and assumed would go away once this was fixed.

haesbaert:
> There is a second hang in the `direct_copy` test, I'm not sure it is related now, but I could reproduce it before as well, this fixes the first one at least.

It's also likely I'm just hiding the original bug by changing some timing.

talex5:
Is there an upstream bug report for this somewhere (e.g. on https://github.com/axboe/liburing)?

haesbaert:
> Is there an upstream bug report for this somewhere (e.g. on https://github.com/axboe/liburing)?

I have to check deeper, I found a bunch of stuff related to bugs with pipes, also the commit for adding splice support for nonblocking in pipes is from June this year, it's all very recent (https://lore.kernel.org/all/20220607080619.513187-5-hao.xu@linux.dev/)

Basically setting the pipes to nonblocking fix the first case, but the second case makes the splice call return EAGAIN, and restarting/handling is not enough. it ends up hanging at some point.

I need to isolate things better before opening something upstream, but it all seems a bit dicey and immature, I'm writing C equivalents to make sure we're not missing something.

haesbaert:
> 

As noted in https://github.com/ocaml-multicore/eio/issues/319#issuecomment-1258087688, I can reproduce the behaviour in that C program.

anmonteiro:
> I believe we should make all our FDs non blocking as a precaution.

After reading through this PR and comments, I did this for my application, which solved hangs I was seeing under `eio_linux` for [eio-ssl](https://github.com/anmonteiro/eio-ssl)

bikallem:
Probably outside of the scope of this PR, but shouldn't we make the sockets non blocking by default too - if they are not already? I was under the impression that the uring and luv backends both set sockets fd to be non-blocking by default. Is this not the case?

haesbaert:
I've ran some more tests today and this is where we are:
 - The fix has been commited to mainline linux: https://github.com/torvalds/linux/commit/46a525e199e4037516f7e498c18f065b09df32ac
 - We can't expect users to be on kernel 6.1.
 - The fix we can do now is use pipes as non-blocking _and_ disable splice, I couldn't reproduce the bug with this.
 - The kernel fix also fixes the splice case, so it's the same bug, so in the future we can use it.
 - There are more things that need to be changed in uring but I wouldn't address that now, for now I'm concerned with just providing something that won't hang.

talex5 (approved):
I rebased it on `main` to fix the CI error.

haesbaert:
> Is this ready to merge, or does it need splice to be disabled first?

Apologies, I should have been more clear, it's not ready, it was more like checking if this would be ok. I've also wanted to hunt the kernel versions where the patch has been applied and maybe test conditionally for disabling it, but I didn't get to it.

haesbaert:
This should be good to go, I updated the commit message as well.

PR #329, opened 2022-09-27
Qualify `Effect.Unhandled`

anmonteiro:
https://github.com/ocaml/ocaml/pull/11423 is landing in the next pre-release of OCaml 5, which moves the `Unhandled` exception to the `Effect` module.


talex5 (approved):
Thanks. Good to merge once the next OCaml alpha is out.

talex5:
The README instructions need updating too.

talex5:
Fixed in #346.

Issue #126, opened 2022-01-03
API for spawning sub-processes


talex5:
Eio needs to provide a way to create and manage sub-processes (like `Lwt_process`).

- Must work in programs using multiple domains. Note that `Unix.fork` cannot be used there.
- API should try to prevent bugs with argument splitting (e.g. avoid error-prone `Unix.system` style API by default). But also needs to work on Windows, where I think you can't provide an argv.
- Child processes should be attached to a switch by default to avoid leaking them, but need a way to spawn detached sub-processes too.
- Must allow passing pipes, etc, to child processes.
- Must prevent accidentally leaking FDs to child processes. All FDs in Eio are already opened with close-on-exec set for this reason.
- Should provide a simple way to check or report the process's exit status.

Possible sources of inspiration include:

- [Lwt_process](http://ocsigen.org/lwt/5.2.0/api/Lwt_process)
- [Bos.Cmd](https://erratique.ch/software/bos/doc/Bos/Cmd/index.html)
- [Feather](https://github.com/charlesetc/feather)
- [Python's subprocess](https://docs.python.org/3/library/subprocess.html#module-subprocess)

Current status:

- [x] #359
- [x] #461
- [x] #464 
- [x] #472
- [ ] #473

patricoferris:
Just a heads up that I've started looking into this as I'm going to need it for some applications I'm building. Hopefully will have a draft PR in the next week or two to iron out the details :))

benmandrew:
Polite bump; this would be extremely useful for converting a lot of the ocurrent ecosystem to Eio. The work in https://github.com/ocurrent/obuilder/pull/113 looks great so far!

talex5:
The current PR (submitted 2 days ago) is here: https://github.com/ocaml-multicore/eio/pull/473

talex5:
Implemented in #499.

PR #330, opened 2022-09-29
Add subprocess support

patricoferris:
This PR is primarily for generating a discussion around how best to support spawning subprocesses in Eio. In particular for issue #126. The current implementation makes the following decisions.

 - Child I/O by default is mapped to `/dev/null`
 - There's a lot of file-descriptor mangling related to child I/O which definitely needs double-checked.

One unfortunate side-effect is that adding subprocesses bypasses all capability-style security iiuc as someone could easily `cd ..` out of the current working directory of the process. Perhaps processes should live in a separate package?

Note: if it ever lands, it would probably be good to use `io-uring-spawn`: https://www.phoronix.com/news/Linux-LPC2022-io_uring_spawn

### DIFF
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -180,7 +180,7 @@ module Stdenv : sig
     stdout : Flow.sink;
     stderr : Flow.sink;
     net : Net.t;
-    process : Process.t;
+    process : Process.mgr;
     domain_mgr : Domain_manager.t;
     clock : Time.clock;
     fs : Fs.dir Path.t;
@@ -227,7 +227,9 @@ module Stdenv : sig
       To use this, see {!Domain_manager}.
   *)
 
-  val process : <process : #Process.t as 'a; ..> -> 'a
+  val process : <process : #Process.mgr as 'a; ..> -> 'a
+  (** [process t] allows spawning subprocesses. This is a powerful capability
+      and should be used with care. *)
 
   val domain_mgr : <domain_mgr : #Domain_manager.t as 'a; ..> -> 'a
   (** [domain_mgr t] allows running code on other cores. *)
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -84,6 +84,9 @@ module Buf_write = Buf_write
 (** Networking. *)
 module Net = Net
 
+(** Subprocesses *)
+module Process = Process
+
 (** Parallel computation across multiple CPU cores. *)
 module Domain_manager : sig
   class virtual t : object
@@ -177,6 +180,7 @@ module Stdenv : sig
     stdout : Flow.sink;
     stderr : Flow.sink;
     net : Net.t;
+    process : Process.t;
     domain_mgr : Domain_manager.t;
     clock : Time.clock;
     fs : Fs.dir Path.t;
@@ -222,6 +226,8 @@ module Stdenv : sig
 
       To use this, see {!Domain_manager}.
   *)
+
+  val process : <process : #Process.t as 'a; ..> -> 'a
 
   val domain_mgr : <domain_mgr : #Domain_manager.t as 'a; ..> -> 'a
   (** [domain_mgr t] allows running code on other cores. *)
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -228,8 +228,7 @@ module Stdenv : sig
   *)
 
   val process : <process : #Process.mgr as 'a; ..> -> 'a
-  (** [process t] allows spawning subprocesses. This is a powerful capability
-      and should be used with care. *)
+  (** [process t] allows spawning subprocesses. *)
 
   val domain_mgr : <domain_mgr : #Domain_manager.t as 'a; ..> -> 'a
   (** [domain_mgr t] allows running code on other cores. *)
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -180,7 +180,7 @@ module Stdenv : sig
     stdout : Flow.sink;
     stderr : Flow.sink;
     net : Net.t;
-    process : Process.mgr;
+    process_mgr : Process.mgr;
     domain_mgr : Domain_manager.t;
     clock : Time.clock;
     fs : Fs.dir Path.t;
@@ -227,8 +227,8 @@ module Stdenv : sig
       To use this, see {!Domain_manager}.
   *)
 
-  val process : <process : #Process.mgr as 'a; ..> -> 'a
-  (** [process t] allows spawning subprocesses. *)
+  val process_mgr : <process_mgr : #Process.mgr as 'a; ..> -> 'a
+  (** [process_mgr t] allows spawning subprocesses. *)
 
   val domain_mgr : <domain_mgr : #Domain_manager.t as 'a; ..> -> 'a
   (** [domain_mgr t] allows running code on other cores. *)
--- a/lib_eio/process.mli
+++ b/lib_eio/process.mli
@@ -2,24 +2,24 @@ type status = Exited of int | Signaled of int | Stopped of int
 
 val pp_status : Format.formatter -> status -> unit
 
-class virtual process : object
+class virtual t : object
     inherit Generic.t
     method virtual pid : int
     method virtual status : status
     method virtual stop : unit
 end
 
-val pid : #process -> int
+val pid : #t -> int
 (** The process ID *)
 
-val status : #process -> status
+val status : #t -> status
 (** Checks the status of the subprocess, this will block waiting for the subprocess
     to terminate or be stopped by a signal. *)
 
-val stop : #process -> unit
+val stop : #t -> unit
 (** Stop a running subprocess *)
 
-class virtual t : object
+class virtual mgr : object
     inherit Generic.t
     method virtual spawn : 
         sw:Switch.t -> 
@@ -29,7 +29,7 @@ class virtual t : object
         ?stdin:Flow.source ->
         string ->
         string list ->
-        process
+        t
     
     method virtual spawn_detached :
         ?cwd:string ->
@@ -38,13 +38,14 @@ class virtual t : object
         ?stdin:Flow.source ->
         string ->
         string list ->
-        process
+        t
 end
+(** A process manager capable of spawning new processes. *)
 
-val spawn : sw:Switch.t -> #t -> ?cwd:string -> ?stderr:Flow.sink -> ?stdout:Flow.sink -> ?stdin:Flow.source -> string -> string list -> process
+val spawn : sw:Switch.t -> #mgr -> ?cwd:string -> ?stderr:Flow.sink -> ?stdout:Flow.sink -> ?stdin:Flow.source -> string -> string list -> t
 (** [spawn ~sw t cmd] creates a new subprocess that is connected to the
     switch [sw]. The standard input and outputs redirect to nothing by default. *)
 
-val spawn_detached : #t -> ?cwd:string -> ?stderr:Flow.sink -> ?stdout:Flow.sink -> ?stdin:Flow.source -> string -> string list -> process
+val spawn_detached : #mgr -> ?cwd:string -> ?stderr:Flow.sink -> ?stdout:Flow.sink -> ?stdin:Flow.source -> string -> string list -> t
 (** [spawn_detached t cmd] is like {! spawn} but the new subprocess is not
     attached to any switch. *)
--- a/lib_eio/process.mli
+++ b/lib_eio/process.mli
@@ -1,0 +1,50 @@
+type status = Exited of int | Signaled of int | Stopped of int
+
+val pp_status : Format.formatter -> status -> unit
+
+class virtual process : object
+    inherit Generic.t
+    method virtual pid : int
+    method virtual status : status
+    method virtual stop : unit
+end
+
+val pid : #process -> int
+(** The process ID *)
+
+val status : #process -> status
+(** Checks the status of the subprocess, this will block waiting for the subprocess
+    to terminate or be stopped by a signal. *)
+
+val stop : #process -> unit
+(** Stop a running subprocess *)
+
+class virtual t : object
+    inherit Generic.t
+    method virtual spawn : 
+        sw:Switch.t -> 
+        ?cwd:string ->
+        ?stderr:Flow.sink ->
+        ?stdout:Flow.sink -> 
+        ?stdin:Flow.source ->
+        string ->
+        string list ->
+        process
+    
+    method virtual spawn_detached :
+        ?cwd:string ->
+        ?stderr:Flow.sink ->
+        ?stdout:Flow.sink -> 
+        ?stdin:Flow.source ->
+        string ->
+        string list ->
+        process
+end
+
+val spawn : sw:Switch.t -> #t -> ?cwd:string -> ?stderr:Flow.sink -> ?stdout:Flow.sink -> ?stdin:Flow.source -> string -> string list -> process
+(** [spawn ~sw t cmd] creates a new subprocess that is connected to the
+    switch [sw]. The standard input and outputs redirect to nothing by default. *)
+
+val spawn_detached : #t -> ?cwd:string -> ?stderr:Flow.sink -> ?stdout:Flow.sink -> ?stdin:Flow.source -> string -> string list -> process
+(** [spawn_detached t cmd] is like {! spawn} but the new subprocess is not
+    attached to any switch. *)
--- a/lib_eio/process.mli
+++ b/lib_eio/process.mli
@@ -24,28 +24,28 @@ class virtual mgr : object
     method virtual spawn : 
         sw:Switch.t -> 
         ?cwd:Fs.dir Path.t ->
-        ?stderr:Flow.sink ->
-        ?stdout:Flow.sink -> 
-        ?stdin:Flow.source ->
+        stdin:Flow.source ->
+        stdout:Flow.sink -> 
+        stderr:Flow.sink ->
         string ->
         string list ->
         t
     
     method virtual spawn_detached :
         ?cwd:Fs.dir Path.t ->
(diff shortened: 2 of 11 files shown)


c-cube:
I think processes shouldn't live in another package :-). They're a cornerstone of any unix-y IO library, and having them tightly integrated with `Flow` is really nice! 

If you worry about capability security, maybe this capability should be easy to disable (in an irreversible way)? But I personally think capabilities in OCaml are more about avoiding mistakes, than true security, since there's basically no sandboxing possible.

lib_eio_linux/eio_linux.ml
```diff
@@ -1160,11 +1160,87 @@ let net = object
   method getnameinfo = Eio_unix.getnameinfo
 end
 
+let pid_to_process close pid = object
+  inherit Eio.Process.process
+  method pid = pid
+  method status = 
+    match Eio_unix.run_in_systhread @@ fun () -> let v = Unix.waitpid [] pid in close (); v with
```
talex5, line 1167: You might need to handle `EINTR` here.

lib_eio_linux/eio_linux.ml
```diff
@@ -1160,11 +1160,87 @@ let net = object
   method getnameinfo = Eio_unix.getnameinfo
 end
 
+let pid_to_process close pid = object
+  inherit Eio.Process.process
+  method pid = pid
+  method status =
```
talex5, line 1166: We need to reap the child even if the caller doesn't call `status`. So status should probably just return a promise that gets resolved automatically when it finishes (this is also what Lwt does).

lib_eio_linux/eio_linux.ml
```diff
@@ -1160,11 +1160,87 @@ let net = object
...
+let open_null flags = lazy (Unix.openfile "/dev/null" flags 0o666)
+
+let dev_null_in = open_null [ Unix.O_RDONLY; Unix.O_CLOEXEC ]
+
+let dev_null_out = open_null [ Unix.O_WRONLY; Unix.O_CLOEXEC ]
+
+let process = object
+  inherit Eio.Process.t
+
+  (* TODO: Is Sys.chdir cwd domain-safe, ? *)
```
talex5, line 1199: No, and neither is forcing a promise. You'll probably need some C code for this bit.

lib_eio_linux/eio_linux.ml
```diff
@@ -1160,11 +1160,87 @@ let net = object
...
+        (Array.of_list args) 
+        stdin stdout stderr
+    in
+    let close () =
+      Unix.close stdin;
+      Unix.close stdout;
+      Unix.close stderr
+    in
+    let process = pid_to_process close pid in
+    let cleanup () =
```
talex5, line 1216: Should kill it here I think. We know that no fiber is waiting for the result, or the switch wouldn't be getting released.

talex5 (commented):
> Child I/O by default is mapped to /dev/null

Might be better to make these non-optional and force the caller to decide what to do. Sending errors to /dev/null by default sounds frustrating.

> One unfortunate side-effect is that adding subprocesses bypasses all capability-style security iiuc as someone could easily cd .. out of the current working directory of the process.

This is OK; it just means that the process manager is a powerful capability and you need to audit code that gets it carefully. Normally, you'd wrap it quickly in something more limited (e.g. that runs a single trusted command, or runs it inside bubblewrap, etc).

You could take a `Path.t` for the cwd. That would allow running a sub-process in a directory that wasn't in the parent's default mount namespace (and would fail if it wasn't backed by an FD).

An API like that would also be necessary on FreeBSD in `cap_enter` mode, since (presumably) there you wouldn't have access to `chdir` (only to `fchdir`).

I suggest renaming `Process.t` to `Process.mgr`, and `Process.process` to `Process.t`.

BTW, I've been idly wondering if we can (ab)use format strings for shell commands. e.g.

```ocaml
# Process.expand "rm -- %s" "My file";;
["rm"; "--"; "My file"]
```
Might be a bit of fun to get that working! Might also confuse people, though.

lib_eio_linux/eio_linux.ml
patricoferris: Handled in [91c7c4a](https://github.com/ocaml-multicore/eio/pull/330/commits/91c7c4a9299e8639d313de7eb590722fe2b19baf)

lib_eio_linux/eio_linux.ml
patricoferris: Hopefully done in https://github.com/ocaml-multicore/eio/pull/330/commits/0127acc27b8db6dbb4ae3b29ae3546597f4e98d1 -- it tries to stop the process and catches `ESRCH` if it is already gone, would it be better to remove the hook if `status` is called instead?

lib_eio_linux/eio_linux.ml
patricoferris: `status` now returns a promise :))

lib_eio_linux/eio_linux.ml
patricoferris: Cleaning up now stops the process 👍

lib_eio/eio.mli
```diff
@@ -223,6 +227,10 @@ module Stdenv : sig
       To use this, see {!Domain_manager}.
   *)
 
+  val process : <process : #Process.mgr as 'a; ..> -> 'a
+  (** [process t] allows spawning subprocesses. This is a powerful capability
```
anmonteiro, line 231: I believe the 2nd sentence is a bit unnecessary here.

lib_eio/process.mli
```diff
@@ -0,0 +1,51 @@
...
+        stdout:Flow.sink -> 
+        stderr:Flow.sink ->
+        string ->
+        string list ->
+        t
+end
+(** A process manager capable of spawning new processes. *)
+
+val spawn : sw:Switch.t -> #mgr -> ?cwd:Fs.dir Path.t -> stdin:Flow.source -> stdout:Flow.sink -> stderr:Flow.sink -> string -> string list -> t
+(** [spawn ~sw t cmd] creates a new subprocess that is connected to the
```
anmonteiro, line 46: Can we call this `mgr`, since `t` refers to a "subprocess"?
```suggestion
(** [spawn ~sw mgr ?cwd ~stdin ~stdout ~stderr cmd args] creates a new subprocess that is connected to the
```

lib_eio/process.mli
```diff
@@ -0,0 +1,51 @@
...
+    switch [sw]. The standard input and outputs redirect to nothing by default. *)
+
+val spawn_detached : #mgr -> ?cwd:Fs.dir Path.t -> stdin:Flow.source -> stdout:Flow.sink -> stderr:Flow.sink -> string -> string list -> t
+(** [spawn_detached t cmd] is like {! spawn} but the new subprocess is not
```
anmonteiro, line 50:
```suggestion
(** [spawn_detached mgr ?cwd ~stdin ~stdout ~stderr cmd args] is like {! spawn} but the new subprocess is not
```

lib_eio/process.mli
```diff
@@ -0,0 +1,51 @@
...
+        stderr:Flow.sink ->
+        string ->
+        string list ->
+        t
+end
+(** A process manager capable of spawning new processes. *)
+
+val spawn : sw:Switch.t -> #mgr -> ?cwd:Fs.dir Path.t -> stdin:Flow.source -> stdout:Flow.sink -> stderr:Flow.sink -> string -> string list -> t
+(** [spawn ~sw t cmd] creates a new subprocess that is connected to the
+    switch [sw]. The standard input and outputs redirect to nothing by default. *)
```
anmonteiro, line 47: I believe this sentence is wrong. These arguments are required. 

```suggestion
    switch [sw]. *)
```

anmonteiro (commented):
A few minor comments that jumped at me while reading the PR. Thanks for your work, @patricoferris

lib_eio/process.mli
patricoferris: Yep, much clearer :)) Fixed in https://github.com/ocaml-multicore/eio/pull/330/commits/b02b37b3a96429b1e5d920ebc1d67803e8a31892

patricoferris:
Thanks @anmonteiro, all seemed sensible and I added the changes in https://github.com/ocaml-multicore/eio/pull/330/commits/b02b37b3a96429b1e5d920ebc1d67803e8a31892

I think I've addressed most comments there is just the relevant C code for being able to change directories for processes in the presence of multiple domains, I'll look into this soon.

lib_eio/process.mli
```diff
@@ -0,0 +1,54 @@
...
+(** A process manager capable of spawning new processes. *)
+
+val spawn : sw:Switch.t -> #mgr -> ?cwd:Fs.dir Path.t -> stdin:Flow.source -> stdout:Flow.sink -> stderr:Flow.sink -> string -> string list -> t
+(** [spawn ~sw mgr ?cwd ~stdin ~stdout ~stderr cmd args] creates a new subprocess that is connected to the
+    switch [sw]. A process will be stopped when the switch is released.
+    
+    You must provide a standard input and outputs that are backed by file descriptors and
+    [cwd] will optionally change the current working directory of the process.*)
+
+val spawn_detached : #mgr -> ?cwd:Fs.dir Path.t -> stdin:Flow.source -> stdout:Flow.sink -> stderr:Flow.sink -> string -> string list -> t
```
talex5, line 52: What's the intended use of this? Is it the same as a double-fork?

lib_eio/unix/eio_unix.ml
```diff
@@ -17,6 +17,7 @@ module Private = struct
     | Get_system_clock : Eio.Time.clock Effect.t
     | Socket_of_fd : Eio.Switch.t * bool * Unix.file_descr -> socket Effect.t
     | Socketpair : Eio.Switch.t * Unix.socket_domain * Unix.socket_type * int -> (socket * socket) Effect.t
+    | Pipe : Eio.Switch.t -> (<Eio.Flow.source; Eio.Flow.close; unix_fd> * <Eio.Flow.sink; Eio.Flow.close; unix_fd>) Effect.t
```
talex5, line 20: This looks useful. I suggest putting it in a separate PR so we can review it by itself.

lib_eio_linux/eio_linux.ml
```diff
@@ -1162,11 +1162,90 @@ let net = object
...
+external sig_to_host_sig : int -> int = "caml_signal_to_posix_signal"
+
+let pid_to_process close pid = object
+  inherit Eio.Process.t
+  method pid = pid
+  val mutable needs_close = true
+  method needs_close = needs_close
+  method set_needs_close b = needs_close <- b
+  method status =
+    let p, r = Promise.create () in
```
talex5, line 1178: You need to create the promise (and do the wait) outside of `status`, which should just await the promise that already exists. Otherwise we'll try to reap the child once for each time the user calls `status`, whereas we actually need to do it exactly once.

lib_eio_linux/eio_linux.ml
```diff
@@ -1162,11 +1162,90 @@ let net = object
...
+    let run () =
+      let v = wait_for_process [] pid in 
+      if needs_close then (close (); needs_close <- false);
+      match v with
+      | _, WEXITED i -> Promise.resolve r  (Eio.Process.Exited (sig_to_host_sig i))
```
talex5, line 1183:
```suggestion
      | _, WEXITED i -> Promise.resolve r  (Eio.Process.Exited i)
```

lib_eio_linux/eio_linux.ml
```diff
@@ -1162,11 +1162,90 @@ let net = object
...
+    p
+  
+  method stop = 
+    Unix.kill pid Sys.sigkill
+end
+
+let get_fd_or_err flow =
+  match get_fd_opt flow with
+  | Some fd -> 
+    Unix.dup ~cloexec:false (FD.to_unix `Peek fd)
```
talex5, line 1197: Setting an FD to be not close-on-exec needs to be done in the child process (in C). Otherwise, another domain might fork a process at the same time and inherit it.

lib_eio_linux/eio_linux.ml
```diff
@@ -1162,11 +1162,90 @@ let net = object
...
+    in
+    let close () =
+      Unix.close stdin;
+      Unix.close stdout;
+      Unix.close stderr
+    in
+    let process = pid_to_process close pid in
+    let cleanup () =
+      (* Catch if the process is already finished when trying to stop it. *)
+      (try process#stop with Unix.Unix_error (Unix.ESRCH, _, _) -> ());
```
talex5, line 1223: Using `pidfd_open(2)` would likely solve these races, and would avoid the need for a systhread to do the waitpid.

lib_eio_linux/eio_linux.ml
talex5: Not sure about whether we want an exception or not. I'm tempted to put the `waitpid` in its own fiber inside the switch, in which case it would stop the switch from exiting (unless there was an error). Then forking a process behaves like forking a fiber, keeping the switch alive.

lib_eio_linux/eio_stubs.c
```diff
@@ -99,3 +99,10 @@ CAMLprim value caml_eio_getdents(value v_fd) {
 
   CAMLreturn(result);
 }
+
+CAMLextern int caml_convert_signal_number(int);
+
+CAMLprim value caml_signal_to_posix_signal(value v_signo)
```
talex5, line 105: What is this for?

lib_eio/process.mli
patricoferris: I thought this was what was needed for this point in the original issue:

> Child processes should be attached to a switch by default to avoid leaking them, but need a way to spawn detached sub-processes too.

Perhaps I misunderstood? Looking at https://stackoverflow.com/questions/881388/what-is-the-reason-for-performing-a-double-fork-when-creating-a-daemon it looks like I have, perhaps this could be added in a follow-up PR?

lib_eio_linux/eio_stubs.c
patricoferris: I should have made that clearer in the code sorry. The problem is that the OCaml signal numbers don't match the POSIX ones (or at least the ones here: https://man7.org/linux/man-pages/man7/signal.7.html). That's why earlier commits were failing with:

```
Full test results in `/src/_build/default/lib_eio_linux/tests/_build/_tests/eio_linux'.
--
355 | Test Successful in 0.003s. 5 tests run.
356 | File "tests/process.md", line 1, characters 0-0:
357 | diff --git a/_build/default/tests/process.md b/_build/default/tests/.mdx/process.md.corrected
358 | index ade46dd..1eb6ef1 100644
359 | --- a/_build/default/tests/process.md
360 | +++ b/_build/default/tests/.mdx/process.md.corrected
361 | @@ -44,7 +44,7 @@ Stopping a subprocess works and checking the status waits and reports correctly
362 | let t = spawn ~sw "sleep" [ "sleep"; "10" ] in
363 | Process.stop t;
364 | Promise.await (Process.status t);;
365 | -- : Process.status = Eio.Process.Signaled 9
366 | +- : Process.status = Eio.Process.Signaled (-7)
367 | ```
368 |  
369 | A switch will stop a process when it is released.
370 | @@ -59,7 +59,7 @@ A switch will stop a process when it is released.
371 | in
372 | run ();
373 | Promise.await @@ Process.status (Option.get !proc);;
374 | -- : Process.status = Eio.Process.Signaled 9
375 | +- : Process.status = Eio.Process.Signaled (-7)
```

as I ran the tests using Luv (which doesn't use OCaml signal numbers). So I was converting them to be the same. Do you think there's a better approach?

lib_eio/unix/eio_unix.ml
patricoferris: Done in #350 :))

lib_eio/eio.mli
```diff
@@ -223,6 +227,9 @@ module Stdenv : sig
       To use this, see {!Domain_manager}.
   *)
 
+  val process : <process : #Process.mgr as 'a; ..> -> 'a
```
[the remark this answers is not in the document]
patricoferris: Yep done in https://github.com/ocaml-multicore/eio/pull/330/commits/1d0f3def3087405fb55968a0089e22c1f3221be4 although now there's a slight discrepancy with having `Eio.Process` instead of `Eio.Process_mgr` but I thought it would be better to keep `Eio.Process` like `Lwt_process`.

lib_eio_linux/eio_linux.ml
patricoferris: Yes! Thanks, done in https://github.com/ocaml-multicore/eio/pull/330/commits/4a37efbf581251ad8021f1252509230211ba58c0

talex5: That only creates the promise outside. It's still doing the wait once for each call. It should just be:
```ocaml
method await_exit = Promise.await exit_status
```

lib_eio_linux/eio_stubs.c
talex5: I don't think there is a standard set of numbers. Even the Linux man-page you linked uses different numbers for different architectures. Seems tempting to use a polymorphic variant here instead (e.g. ``Eio.Process.Signaled `TERM``.

talex5:
One way to make this PR more manageable might be to first split out a PR adding subprocess support to `Eio_linux.Low_level` (or to eio-luv), then add the cross-platform support in a later PR.

lib_eio_linux/eio_stubs.c
haesbaert: I had started this when I was playing with signals:
https://github.com/haesbaert/eio/blob/signal/lib_eio/include/discover.ml
Basically exports the SIGFOO and also tries to compile a program with the non standard SIGBARS that get exported as an optional. It might make sense to point out also that the Stdlib signal calls work with both the ocaml values (the negative ones) and the "real ones".

patricoferris:
Closing in favour of #473

Issue #267, opened 2022-07-29
Consider renaming Stream


rizo:
I think the [Stream](https://github.com/ocaml-multicore/eio/blob/2aadabf4249a0fb14f0f70a06542103fc07ff08f/lib_eio/stream.mli) module in Eio is not a stream in a traditional functional sense.

In particular it look more similar to abstractions provided by some other libraries/runtimes under different names:
- [Async.Pipe](https://github.com/monadbobo/ocaml-core/blob/9c1c06e7a1af7e15b6019a325d7dbdbd4cdb4020/base/async/core/lib/pipe.ml#L69)
- [Lwt_pipe](https://github.com/c-cube/lwt-pipe/blob/a1296c8875cec581a855a44b6658386fca6fb16c/src/Lwt_pipe.ml#L19)
- [Crystal/Channel](https://crystal-lang.org/api/1.5.0/Channel.html)
- [Go/chan](https://github.com/golang/go/blob/5c8ec89cb53025bc76b242b0d2410bf5060b697e/src/runtime/chan.go#L33)

"Streams" in functional languages tend to be related to one-directional element-wise "stream processing".

talex5:
It's named after http://ocsigen.org/lwt/5.5.0/api/Lwt_stream.

c-cube:
I think `Lwt_stream` is a bit controversial in design, so maybe it's not the best example? :)

rizo:
I see. In a way, this actually reinforces my point because `Lwt_stream` has some confusing semantics discussed in https://github.com/ocsigen/lwt/issues/250.

haesbaert:
I also found the name confusing, the wikipedia for Stream kinda backs it up though: https://en.wikipedia.org/wiki/Stream_(computing)#

I see Eio's stream as a mpsafe-queue, as in multiprocessor safe queue.
I'd be happy with anything that hints into it being a queue: Eio_queue, Mqueue, Mdqueue, Mpqueue, Dsqueue (domain safe queue), but then when capacity is 0 it's not really a queue and it's more like L4's rendezvous synchronous message passing.

avsm:
Calling it a Queue does seem clearer (especially that it's a domain-safe FIFO structure).  Rust's [concurrency-queue](https://crates.io/crates/concurrent-queue) separates out unbounded from bounded ones.

[...]

PR #334, opened 2022-09-30
Mbox implementation

haesbaert:
Mboxes is an asynchronous/reliable/mpsc/domain-safe/single-value message passing mechanism.

The mailbox holds one value only, senders block if a value is present, while readers block while a value is not present.

- asynchronous since senders don't block (no rendezvous).
- reliable since messages cannot be lost (contrary to FFQ for example).
- MPSC since you can go nuts.
- domain-safe since any domain can do any combinatino of read/write across domains.
- single-value since the buffer contains only one value.

This is not intended for high performance, it's more of a inter-domain control message passing, but no one will arrest you if you use it in the hot path.

Readers and Writers take turn and they're bounded, there is no thundering storm on any side.

This diff also includes a tests/racers directory, the idea is to add programs that try to race some part of the library.

racers/race_mbox.ml attempts to sends brazilions of messages making sure the program doesn't hang, that messages are not lost and things are dandy. Main strategy is having random sleeps on certain points and count sent vs received messages, it's naive but it works ! I ran it for ~8 hours.

### DIFF
--- a/lib_eio/mbox.ml
+++ b/lib_eio/mbox.ml
@@ -1,0 +1,50 @@
+type 'a t = {
+  writers: unit Waiters.t;
+  readers: unit Waiters.t;
+  mutex: Mutex.t;
+  id: Ctf.id;
+  value: 'a option ref;
+}
+
+let create () = {
+  writers = Waiters.create ();
+  readers = Waiters.create ();
+  mutex = Mutex.create ();
+  id = Ctf.mint_id ();
+  value = ref None;
+}
+
+let send (t:'a t) (m : 'a) : unit =
+  let sent = ref false in
+  let wait () =
+    Mutex.lock t.mutex;
+    match !(t.value) with
+    | Some _ ->
+      Waiters.await ~mutex:(Some t.mutex) t.writers t.id
+    | None ->
+      t.value := Some m;
+      Waiters.wake_one' t.readers ();
+      sent := true;
+      Mutex.unlock t.mutex
+  in
+  while !sent = false do
+    wait ()
+  done
+
+let recv t =
+  let received = ref None in
+  let wait () =
+    Mutex.lock t.mutex;
+    match !(t.value) with
+    | Some v ->
+      t.value := None;
+      received := Some v;
+      Waiters.wake_one' t.writers ();
+      Mutex.unlock t.mutex
+    | None ->
+      Waiters.await ~mutex:(Some t.mutex) t.readers t.id
+  in
+  while !received = None do
+    wait ()
+  done;
+  Option.get !received
--- a/lib_eio/mbox.ml
+++ b/lib_eio/mbox.ml
@@ -24,8 +24,8 @@ let send (t:'a t) (m : 'a) : unit =
     | None ->
       t.value := Some m;
       Waiters.wake_one' t.readers ();
-      sent := true;
-      Mutex.unlock t.mutex
+      Mutex.unlock t.mutex;
+      sent := true
   in
   while !sent = false do
     wait ()
@@ -38,9 +38,9 @@ let recv t =
     match !(t.value) with
     | Some v ->
       t.value := None;
-      received := Some v;
       Waiters.wake_one' t.writers ();
-      Mutex.unlock t.mutex
+      Mutex.unlock t.mutex;
+      received := Some v
     | None ->
       Waiters.await ~mutex:(Some t.mutex) t.readers t.id
   in
--- a/lib_eio/mbox.ml
+++ b/lib_eio/mbox.ml
@@ -23,7 +23,7 @@ let send (t:'a t) (m : 'a) : unit =
       Waiters.await ~mutex:(Some t.mutex) t.writers t.id
     | None ->
       t.value := Some m;
-      Waiters.wake_one' t.readers ();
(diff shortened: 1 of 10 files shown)


lib_eio/mbox.ml
```diff
@@ -0,0 +1,50 @@
...
+  let wait () =
+    Mutex.lock t.mutex;
+    match !(t.value) with
+    | Some _ ->
+      Waiters.await ~mutex:(Some t.mutex) t.writers t.id
+    | None ->
+      t.value := Some m;
+      Waiters.wake_one' t.readers ();
+      sent := true;
+      Mutex.unlock t.mutex
```
c-cube, line 28: wouldn't you want to unlock before waking up the reader? (same below for writer)

haesbaert: Not really, could be before the `sent := true` obviously, readers are in a wait-queue protected by the same mutex as writers.
`Waiters.await` passes the lock-down and that protects the internal `Lwt_dllist`, both wait-queues are under the same lock.

haesbaert: I pushed a diff to do the `sent := true` and `received := true` outside the lock, but the `wakeup_one` call needs to be locked.

c-cube: Ok, makes sense. I suppose that `wakeup_one` will not _immediately_ wake the fiber, it will be queued? otherwise the worry was that it'd try to lock the mutex just before it's released…

haesbaert: > Ok, makes sense. I suppose that `wakeup_one` will not _immediately_ wake the fiber, it will be queued? otherwise the worry was that it'd try to lock the mutex just before it's released…

Correct, the `wakeup_one` resolves the `Suspended` effect created in `Waiters.await`, enqueues it into the `run-queue`, so next time the current `Fiber` blocks, it will pop the next one from the `run-queue` (at least that's as far as my understanding of the code goes at this time, I haven't really dived into the scheduler).

haesbaert: Doing some more digging, I believe `wakeup_one` should be different, so that we don't end up calling the `enqueue` callback while holding the mutex, something like:
```ocaml
Mutex.lock t.mutex;
let waker = Waiters.pop_next_waiter t in
Mutex.unlock t.mutex;
Waiters.wake waker
```

c-cube: There is also the risk of a deadlock, possibly, if the reader this just poped from `waiters` turns out to have been cancelled while it slept? I think? I'm new to all this. With threads I've often started with `Condition.signal` only to move to `Condition.broadcast`.

haesbaert: > There is also the risk of a deadlock, possibly, if the reader this just poped from `waiters` turns out to have been cancelled while it slept? I think? I'm new to all this. With threads I've often started with `Condition.signal` only to move to `Condition.broadcast`.

Leme make sure I understood:
```ocaml
Mutex.lock t.mutex;
let waker = Waiters.pop_next_waiter t in
Mutex.unlock t.mutex;
(* at this point you mean the waker has been canceled *)
Waiters.wake waker
```

So I end up not waking the next one in queue, which in turn deadlocks ?

haesbaert: I pushed a change, my `wakeup_one'` is broken, I tried to describe the possible deadlock in the commit:
https://github.com/ocaml-multicore/eio/pull/334/commits/161579fbf982ead4e1b601d80cecc617b9dbf281

Thanks a lot for this ! I really appreciate you making me ponder about it !

c-cube:
It's tricky stuff 😁

lib_eio/mbox.ml
c-cube: :+1:

talex5:
Note that a Stream of capacity 1 already behaves as an mbox:

https://github.com/ocaml-multicore/eio/blob/e9ad2b005ed27307e0af7b49327ab4ca32cd0719/lib_eio/stream.mli#L19-L24

Would be interesting to compare the performance with that.

haesbaert:
> Note that a Stream of capacity 1 already behaves as an mbox:
> 

I completely missed that, why the name Stream though ? Stream in my head implies a single producer single consumer with no boundary between items. Eio's `Stream` is pretty much a domain-safe queue, it's also way more elegant than my Mbox.

haesbaert:
> > Note that a Stream of capacity 1 already behaves as an mbox:
> 
> I completely missed that, why the name Stream though ? Stream in my head implies a single producer single consumer with no boundary between items. Eio's `Stream` is pretty much a domain-safe queue, it's also way more elegant than my Mbox.

Answering my own question, a quick search shows that my understanding of stream is wrong

talex5:
> I completely missed that, why the name Stream though?

I think it comes from OCaml originally. The stdlib's (now deprecated) [Stream](https://v2.ocaml.org/api/Stream.html) module has discreet items, and I guess [Lwt_stream](http://ocsigen.org/lwt/5.5.0/api/Lwt_stream) copied the name. Then Eio copied that name. It's convenient for people converting Lwt code, but maybe another name would be better. See also #267.

haesbaert:
I'm dropping this until it makes more sense, will run some benchmarks at some point and maybe bring the racer program to Stream, but it makes no sense to let this hanging here for now.

PR #338, opened 2022-10-04
Add Time.Mono for monotonic clocks

talex5:
Based on PR #308 by @bikallem. I didn't add Ptime here, but it can be added using the same pattern later.

It differs from #308 in a few ways:
- This change is mostly backwards-compatible. We can deprecate the old float functions in a later release, once people have moved over.
- The `add_seconds` and `to_seconds` operations on clocks are gone. Instead, it provides a separate submodule for Mtime. This allows the API to work directly with Mtime types instead of converting via floats.
- `Time.Timeout` now works on monotonic clocks (only). It's not useful to use a real-time clock for a timeout, so I think it's fine to skip support for that.

### DIFF
--- a/lib_eio/dune
+++ b/lib_eio/dune
@@ -2,4 +2,4 @@
   (name eio)
   (public_name eio)
   (flags (:standard -open Eio__core -open Eio__core.Private))
-  (libraries eio__core cstruct lwt-dllist fmt bigstringaf optint))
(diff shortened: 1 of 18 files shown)


talex5:
@bikallem any objection to merging this?

bikallem:
Thanks. It seems my notifications from this PR was buried among the ci ones when I did `Select All` and `Done` in my github notifications. Apologies. Yes, no objections.

PR #339, opened 2022-10-05
Expose File.stat and improve Path.load

haesbaert:
Path.load was doubling the buffer and trying to fit enough until it saw an EOF. This would cause many unecessary large allocations and my poor 16GB ram laptop couldn't even load a 5GB file without getting OOMed.

This diff makes load create an initial buffer of the size of the file and load exactly that amount of bytes in it.

Also, using take_all here isn't a good idea since it will resize even if initial_size (capacity) is adequate, that's because it does:
  ensure t.len(0) (0+1), which ends up allocating another bigarray.

### DIFF
--- a/lib_eio/file.ml
+++ b/lib_eio/file.ml
@@ -1,8 +1,47 @@
+(** Tranditional Unix permissions. *)
+module Unix_perm = struct
+  type t = int
+  (** This is the same as {!Unix.file_perm}, but avoids a dependency on [Unix]. *)
+end
+
+(** Portable file stats. *)
+module Stat = struct
+
+  (** Kind of file from st_mode. **)
+  type kind = [
+    | `Unknown
+    | `Fifo
+    | `Character_special
+    | `Directory
+    | `Block_device
+    | `Regular_file
+    | `Symbolic_link
+    | `Socket
+  ]
+
+  (** Like stat(2). *)
+  type t = {
+    dev : Int64.t;
+    ino : Int64.t;
+    kind : kind;
+    perm : Unix_perm.t;
+    nlink : Int64.t;
+    uid : Int64.t;
+    gid : Int64.t;
+    rdev : Int64.t;
+    size : Optint.Int63.t;
+    atime : float;
+    mtime : float;
+    ctime : float;
+  }
+end
+
 (** A file opened for reading. *)
 class virtual ro = object (_ : <Generic.t; Flow.source; ..>)
   method probe _ = None
   method read_methods = []
   method virtual pread : file_offset:Optint.Int63.t -> Cstruct.t list -> int
+  method virtual stat : Stat.t
 end
 
 (** A file opened for reading and writing. *)
@@ -10,6 +49,12 @@ class virtual rw = object (_ : <Generic.t; Flow.source; Flow.sink; ..>)
   inherit ro
   method virtual pwrite : file_offset:Optint.Int63.t -> Cstruct.t list -> int
 end
+
+(** [stat t] returns the {!Stat.t} record associated with [t]. *)
+let stat (t : #ro) = t#stat
+
+(** [size t] returns the size of [t]. *)
+let size t = (stat t).size
 
 (** [pread t ~file_offset bufs] performs a single read of [t] at [file_offset] into [bufs].
 
--- a/lib_eio/fs.ml
+++ b/lib_eio/fs.ml
@@ -1,23 +1,21 @@
 (** Defines types used by file-systems. *)
 
 (** Tranditional Unix permissions. *)
-module Unix_perm = struct
-  type t = int
-  (** This is the same as {!Unix.file_perm}, but avoids a dependency on [Unix]. *)
-end
+module Unix_perm = File.Unix_perm [@@deprecated "Moved to File.Unix_perm"]
 
 type path = string
 
 exception Already_exists of path * exn
 exception Not_found of path * exn
 exception Permission_denied of path * exn
+exception File_too_large of path * exn
 
 (** When to create a new file. *)
 type create = [
-  | `Never                            (** fail if the named file doesn't exist *)
-  | `If_missing of Unix_perm.t        (** create if file doesn't already exist *)
-  | `Or_truncate of Unix_perm.t       (** any existing file is truncated to zero length *)
-  | `Exclusive of Unix_perm.t         (** always create; fail if the file already exists *)
+  | `Never                                 (** fail if the named file doesn't exist *)
+  | `If_missing of File.Unix_perm.t        (** create if file doesn't already exist *)
+  | `Or_truncate of File.Unix_perm.t       (** any existing file is truncated to zero length *)
+  | `Exclusive of File.Unix_perm.t         (** always create; fail if the file already exists *)
 ]
 (** If a new file is created, the given permissions are used for it. *)
 
@@ -30,7 +28,7 @@ class virtual dir = object (_ : #Generic.t)
     append:bool ->
     create:create ->
     path -> <File.rw; Flow.close>
-  method virtual mkdir : perm:Unix_perm.t -> path -> unit
+  method virtual mkdir : perm:File.Unix_perm.t -> path -> unit
   method virtual open_dir : sw:Switch.t -> path -> dir_with_close
   method virtual read_dir : path -> string list
   method virtual unlink : path -> unit
--- a/lib_eio/path.ml
+++ b/lib_eio/path.ml
@@ -30,10 +30,20 @@ let with_lines path fn =
   let buf = Buf_read.of_flow flow ~max_size:max_int in
   fn (Buf_read.lines buf)
 
-let load path =
-  with_open_in path @@ fun flow ->
-  let buf = Buf_read.of_flow flow ~max_size:max_int in
-  Buf_read.take_all buf
+let load (t, path) =
+  with_open_in (t, path) @@ fun flow ->
+  let size = File.size flow in
+  if Optint.Int63.(compare size (of_int Sys.max_string_length)) = 1 then
+    raise (Fs.File_too_large
+             (path, Invalid_argument "can't represent a string that long"));
+  let buf = Cstruct.create (Optint.Int63.to_int size) in
+  let rec loop buf got =
+    match Flow.single_read flow buf with
+    | n -> loop (Cstruct.shift buf n) (n + got)
+    | exception End_of_file -> got
+  in
+  let got = loop buf 0 in
+  Cstruct.to_string ~len:got buf
 
 let save ?append ~create path data =
   with_open_out ?append ~create path @@ fun flow ->
--- a/lib_eio/path.mli
+++ b/lib_eio/path.mli
@@ -88,7 +88,7 @@ val with_open_out :
 
 (** {1 Directories} *)
 
-val mkdir : perm:Unix_perm.t -> _ t -> unit
+val mkdir : perm:File.Unix_perm.t -> _ t -> unit
 (** [mkdir ~perm t] creates a new directory [t] with permissions [perm]. *)
 
 val open_dir : sw:Switch.t -> _ t -> <dir; Flow.close> t
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -110,6 +110,34 @@ module FD = struct
     match t.fd with
     | `Open fd -> Fmt.pf f "%d" (Obj.magic fd : int)
     | `Closed -> Fmt.string f "(closed)"
+
+  let fstat t =
+    (* todo: use uring  *)
+    let ust = Unix.LargeFile.fstat (get_exn "fstat" t) in
+    let st_kind : Eio.File.Stat.kind =
+      match ust.st_kind with
+      | Unix.S_REG  -> `Regular_file
+      | Unix.S_DIR  -> `Directory
+      | Unix.S_CHR  -> `Character_special
+      | Unix.S_BLK  -> `Block_device
+      | Unix.S_LNK  -> `Symbolic_link
+      | Unix.S_FIFO -> `Fifo
+      | Unix.S_SOCK -> `Socket
+    in
+    Eio.File.Stat.{
+      dev     = ust.st_dev   |> Int64.of_int;
+      ino     = ust.st_ino   |> Int64.of_int;
+      kind    = st_kind;
+      perm    = ust.st_perm;
+      nlink   = ust.st_nlink |> Int64.of_int;
+      uid     = ust.st_uid   |> Int64.of_int;
+      gid     = ust.st_gid   |> Int64.of_int;
+      rdev    = ust.st_rdev  |> Int64.of_int;
+      size    = ust.st_size  |> Optint.Int63.of_int64;
+      atime   = ust.st_atime;
+      mtime   = ust.st_mtime;
+      ctime   = ust.st_ctime;
+    }
 end
 
 type _ Eio.Generic.ty += FD : FD.t Eio.Generic.ty
@@ -816,9 +844,7 @@ module Low_level = struct
     | Cwd -> openat2 ~sw ?seekable ~access ~flags ~perm ~resolve:Uring.Resolve.beneath path
     | Fs -> openat2 ~sw ?seekable ~access ~flags ~perm ~resolve:Uring.Resolve.empty path
 
-  let fstat fd =
-    (* todo: use uring *)
-    Unix.fstat (FD.get_exn "fstat" fd)
+  let fstat fd = FD.fstat fd
 
   external eio_mkdirat : Unix.file_descr -> string -> Unix.file_perm -> unit = "caml_eio_mkdirat"
 
@@ -1032,6 +1058,8 @@ let flow fd =
   object (_ : <source; sink; ..>)
     method fd = fd
     method close = FD.close fd
+
+    method stat = FD.fstat fd
 
     method probe : type a. a Eio.Generic.ty -> a option = function
       | FD -> Some fd
--- a/lib_eio_linux/tests/eurcp_lib.ml
+++ b/lib_eio_linux/tests/eurcp_lib.ml
@@ -31,7 +31,7 @@ let run_cp block_size queue_depth infile outfile () =
   let open Unix in
   let infd = U.openfile ~sw infile [O_RDONLY] 0 in
   let outfd = U.openfile ~sw outfile [O_WRONLY; O_CREAT; O_TRUNC] 0o644 in
-  let insize = U.fstat infd |> fun {st_size; _} -> Int63.of_int st_size in
+  let insize = (U.fstat infd).size in
   Logs.debug (fun l -> l "eurcp: %s -> %s size %a queue %d bs %d"
                  infile
                  outfile
--- a/lib_eio_luv/eio_luv.ml
+++ b/lib_eio_luv/eio_luv.ml
@@ -249,6 +249,33 @@ let get_loop () =
   enter_unchecked @@ fun t k ->
   Suspended.continue k t.loop
 
+let unix_fstat fd =
+  let ust = Unix.LargeFile.fstat fd in
+  let st_kind : Eio.File.Stat.kind =
+    match ust.st_kind with
+    | Unix.S_REG  -> `Regular_file
+    | Unix.S_DIR  -> `Directory
+    | Unix.S_CHR  -> `Character_special
+    | Unix.S_BLK  -> `Block_device
+    | Unix.S_LNK  -> `Symbolic_link
+    | Unix.S_FIFO -> `Fifo
+    | Unix.S_SOCK -> `Socket
+  in
+  Eio.File.Stat.{
+    dev     = ust.st_dev   |> Int64.of_int;
+    ino     = ust.st_ino   |> Int64.of_int;
+    kind    = st_kind;
+    perm    = ust.st_perm;
+    nlink   = ust.st_nlink |> Int64.of_int;
+    uid     = ust.st_uid   |> Int64.of_int;
+    gid     = ust.st_gid   |> Int64.of_int;
+    rdev    = ust.st_rdev  |> Int64.of_int;
+    size    = ust.st_size  |> Optint.Int63.of_int64;
+    atime   = ust.st_atime;
+    mtime   = ust.st_mtime;
+    ctime   = ust.st_ctime;
+  }
+
 module Low_level = struct
   type 'a or_error = ('a, Luv.Error.t) result
 
@@ -376,6 +403,10 @@ module Low_level = struct
       let t = of_luv_no_hook ~close_unix fd in
       t.release_hook <- Switch.on_release_cancellable sw (fun () -> ensure_closed t);
       t
+
+    let fstat fd =
+      let request = Luv.File.Request.make () in
+      await_with_cancel ~request (fun loop -> Luv.File.fstat ~loop ~request (to_luv fd))
 
     let open_ ~sw ?mode path flags =
       let request = Luv.File.Request.make () in
@@ -597,6 +628,8 @@ let flow fd = object (_ : <source; sink; ..>)
   method fd = fd
   method close = Low_level.File.close fd
   method unix_fd op = File.to_unix op fd
+
+  method stat = unix_fstat (File.to_unix `Peek fd)
 
   method probe : type a. a Eio.Generic.ty -> a option = function
     | FD -> Some fd
--- a/tests/fs.md
+++ b/tests/fs.md
@@ -64,6 +64,10 @@ let try_rmdir path =
 let chdir path =
   traceln "chdir %S" path;
   Unix.chdir path
+
+let assert_kind path kind =
+  Path.with_open_in path @@ fun file ->
+  assert ((Eio.File.stat file).kind = kind)
 ```
 
 # Basic test cases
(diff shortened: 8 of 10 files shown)


lib_eio/path.ml
```diff
@@ -32,8 +33,9 @@ let with_lines path fn =
 
 let load path =
   with_open_in path @@ fun flow ->
-  let buf = Buf_read.of_flow flow ~max_size:max_int in
-  Buf_read.take_all buf
+  let initial_size = size path in
+  let buf = Buf_read.of_flow flow ~initial_size ~max_size:max_int in
+  Buf_read.take initial_size buf
```
talex5, line 38: Seems overly complicated. How about just `read_exact`?

lib_eio/path.mli
```diff
@@ -40,6 +40,9 @@ val ( / ) : 'a t -> string -> 'a t
 val pp : _ t Fmt.t
 (** [pp] formats a [_ t] as "<label:path>", suitable for logging. *)
 
+val size : _ t -> int
```
talex5, line 43: This will be too small on 32-bit systems. We're using int63 for file offsets, so should probably use that here too.

lib_eio_linux/eio_linux.ml
```diff
@@ -1205,6 +1205,11 @@ class dir ~label (fd : dir_fd) = object
     | Dir_fd -> Some fd
     | _ -> None
 
+  method size path =
+    wrap_errors path @@ fun () ->
+    let st = Unix.stat path in
```
talex5, line 1210: This needs to be relative to `fd`.

talex5 (commented):
Thanks, this does need improving.

I'm not sure `size` is the right method, though. Perhaps we need a more general `stat` operation?

The Linux `statx` system call takes a mask saying what information you want. Maybe we could do something like that?

lib_eio_linux/eio_linux.ml
haesbaert: We could, but then it becomes a pain as you need to open just to get its size, but no strong feelings.

talex5: Well, it needs to be correct and give you the size of the file you asked for, not some other file with the same name!

But `fstatat` and `statx` both take the FD as an argument anyway.

haesbaert: ah yes I see what you mean, ofc, so I openat (resolve_beneath) -> fstat -> close ?

lib_eio/path.mli
haesbaert: I was wondering, because we already take an `int` as `maxsize` in `Buf_read`, so are we actually supporting 32bit ?

talex5: You can't load a 4GB file into memory in one go on a 32-bit system, so it's fine for the maximum buffer size to be an `int`. But you should still be able to find out how big it is and then load it in chunks.

(`size` returning an `int` would be OK for `Path.load`, but not necessarily for other uses)

lib_eio_linux/eio_linux.ml
talex5: Yes, I think that's best. It could be optimised for the common case where `fd` is `Fs`. Or, if we had linked uring requests, it could all be done in one go too.

haesbaert:
Let me know what you think, I've added two versions of `load`, the problem with using `read_exact` is that if the file shrinks mid-read we just throw an EOF, the second issue is that due to the integer overflow in `luv`, `read_exact` would end up always giving an EOF.
So I've added both versions, one that copes with getting an early EOF, and one that doesn't.
I've fixed the Uring fstat to use the Large version as well.
If you don't want to expose `size` as you mentioned before, we have to decide what to put and how to export `stat`, since we would need to see what Luv offers in windows and so on.

talex5 (commented):
> the problem with using read_exact is that if the file shrinks mid-read we just throw an EOF

OK, that seems like a good reason to use the `Buf_read`!

> the second issue is that due to the integer overflow in luv, read_exact would end up always giving an EOF.

If the low-level read operation doesn't work correctly, then neither will Buf_read built on top of it. `Eio_luv` should just limit the buffer size to something it can handle. Callers need to handle short reads anyway (and I doubt to OS is going to write more than 2GB in one go anyway).

> I've fixed the Uring fstat to use the Large version as well.

Thanks!

> If you don't want to expose size as you mentioned before, we have to decide what to put and how to export stat, since we would need to see what Luv offers in windows and so on.

Yes, it needs a bit of thinking about. But I'd prefer to avoid adding a new method that everyone has to implement, when we know we're going to get rid of it again soon.

haesbaert:
This is ready for review, let me know what you think.

lib_eio/path.ml
```diff
@@ -30,10 +31,22 @@ let with_lines path fn =
...
+  if Optint.Int63.(compare st.st_size (of_int mymax)) = 1 then
+    raise (Fs.File_too_large
+             (path, Invalid_argument "can't represent a string that long"));
+  with_open_in (t, path) @@ fun flow ->
+  let buf = Cstruct.create (Optint.Int63.to_int st.st_size) in
+  let rec loop buf got =
+    try
+      let n = Flow.read flow buf in
+      loop (Cstruct.shift buf n) (n + got)
+    with End_of_file -> got
```
talex5, line 46:
```suggestion
    match Flow.read flow buf with
    | got -> loop (Cstruct.shift buf n) (n + got)
    | exception End_of_file -> got
```
Otherwise, it's not tail-recursive.

lib_eio_linux/eio_linux.ml
```diff
@@ -1207,6 +1208,42 @@ class dir ~label (fd : dir_fd) = object
     | Dir_fd -> Some fd
     | _ -> None
 
+  method stat path =
+    Switch.run @@ fun sw ->
+    let fd = Low_level.openat ~sw ~seekable:false fd path
+          ~access:`R
+          ~flags:Uring.Open_flags.(path + cloexec + nofollow)
```
talex5, line 1215: So `nofollow` makes this behave as `lstat`? Needs a comment I think, and maybe a separate function. Otherwise, `File.load` on a symlink will use the size of the link instead of the size of the target, won't it? Could do with a test-case, anyway.

lib_eio_luv/eio_luv.ml
```diff
@@ -81,6 +81,17 @@ let rec cstructv_to_luv = function
     let x1, x2 = Cstruct.split x max_luv_buffer_size in
     Cstruct.to_bigarray x1 :: cstructv_to_luv (x2 :: xs)
 
+let float_of_timespec (ts : Luv.File.Stat.timespec) =
+  let s : float = Signed.Long.to_int64  ts.sec |> Int64.to_float in
+  let nsec : float = Signed.Long.to_int64  ts.nsec |> Int64.to_float in
```
talex5, line 86:
```suggestion
  let s : float = Signed.Long.to_int64 ts.sec |> Int64.to_float in
  let nsec : float = Signed.Long.to_int64 ts.nsec |> Int64.to_float in
```

lib_eio_luv/eio_luv.ml
```diff
@@ -81,6 +81,17 @@ let rec cstructv_to_luv = function
...
+
+  let n = nsec /. 1_000_000_000.0 in
+  let t = s +. n in
+  if t = (s +. 1.0) then
+    Float.next_after t s
```
talex5, line 91: What's this for? Needs a comment.

lib_eio_luv/eio_luv.ml
```diff
@@ -455,6 +470,28 @@ module Low_level = struct
         t.fd <- `Closed;
         Eio.Switch.remove_hook t.release_hook;
         fd
+
+    let kind_map =
+      (* we only need 12, but 13 maps directly *)
+      let km = Array.init 13 (fun _ -> `Unknown) in
+      Array.set km 0o1  `Fifo;
```
talex5, line 477: Where do these numbers come from?

lib_eio_luv/eio_luv.ml
```diff
@@ -455,6 +470,28 @@ module Low_level = struct
...
+      Array.set km 0o10 `Regular_file;
+      Array.set km 0o12 `Symbolic_link;
+      Array.set km 0o14 `Socket;
+      km
+
+    let kind_of_mode (mode : int) : Eio.Fs.kind =
+      if Sys.os_type <> "Unix" then
+        `Unknown
+      else
+        try Array.get kind_map (mode lsr 12)
```
talex5, line 490: Should this be masked with `S_IFMT`?

lib_eio_luv/eio_luv.ml
```diff
@@ -455,6 +470,28 @@ module Low_level = struct
...
+      Array.set km 0o4  `Directory;
+      Array.set km 0o6  `Block_device;
+      Array.set km 0o10 `Regular_file;
+      Array.set km 0o12 `Symbolic_link;
+      Array.set km 0o14 `Socket;
+      km
+
+    let kind_of_mode (mode : int) : Eio.Fs.kind =
+      if Sys.os_type <> "Unix" then
+        `Unknown
```
talex5, line 488: What happens on Windows?

lib_eio_luv/eio_luv.ml
```diff
@@ -953,6 +990,29 @@ class dir ~label (dir_path : string) = object (self)
       | exception Eio.Fs.Permission_denied (dir, ex) ->
         raise (Eio.Fs.Permission_denied (Filename.concat dir leaf, ex))
 
+  method stat path =
+    Switch.run @@ fun sw ->
+    let fd = File.open_ ~sw (self#resolve path) [`NOFOLLOW; `RDONLY] |> or_raise_path path in
+    let lst : Luv.File.Stat.t = File.fstat fd |> or_raise_path path in
+    File.close fd;
+    let mode = (Obj.magic lst.mode : int) in
```
talex5, line 998: Is there an issue open on Luv to track this? We'll want to be notified when it's fixed.

tests/fs.md
```diff
@@ -495,3 +499,25 @@ Unconfined:
 +rename <fs:../foo> to <fs:foo> -> ok
 - : unit = ()
 ```
+
+# Stat
```
talex5, line 503: With `make test_luv`, I get:
```
+|Exception: Eio.Fs.Not_found ("stat_link", _)
```

lib_eio/path.ml
```diff
@@ -30,10 +31,22 @@ let with_lines path fn =
   let buf = Buf_read.of_flow flow ~max_size:max_int in
   fn (Buf_read.lines buf)
 
-let load path =
-  with_open_in path @@ fun flow ->
-  let buf = Buf_read.of_flow flow ~max_size:max_int in
-  Buf_read.take_all buf
+let load ((t:#Fs.dir), path) =
+  let st = t#stat path in
+  let mymax = min Sys.max_string_length Sys.max_array_length in
```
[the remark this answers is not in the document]
haesbaert: Because we have to store it in a Cstruct.t before the converting it to a string, and max_array_length < max_string_length, at least on amd64.

lib_eio/path.ml
haesbaert: good catch ! I only looked at the call and ignored it was wrapped in a try catch.

lib_eio_linux/eio_linux.ml
haesbaert: Pretty much, I added this late when I was writing the symbolic link test. Now thinking about it, we should probably make `stat` get an optional argument for `nofollow`, `load` would want it to `follow` while direct calls would probably want it to `nofollow`. Also, should we restrict `load` just to a resulting `Regular_file` ? Can't think of any more uses.

lib_eio_luv/eio_luv.ml
haesbaert: Not needed but I'll add it for clarity, POSIX defines those 12 bits and we never look outside them.

lib_eio_luv/eio_luv.ml
haesbaert: No, I'll open one.

tests/fs.md
haesbaert: oopsy, will fix it.

lib_eio_luv/eio_luv.ml
haesbaert: No clue :man_shrugging:, so I chose to be conservative until we could test it on windows.

lib_eio_luv/eio_luv.ml
haesbaert: I'll clarify in a comment, but I was mostly trying to avoid having a `discovery.ml` for `S_IF*`, the values are actually standarized and I've checked linux/OpenBSD/FreeBSD/OS X. The array of 13 is that we can store say `S_IFSOCK (0o140000) at km.(0o14) instead of 0o13`.

lib_eio_luv/eio_luv.ml
haesbaert: I basically copied the algorithm from the runtime (they do it in C though), I didn't reason much about it.

tests/fs.md
haesbaert: heh, `File.realpath` resolves the symlink and returns ENOENT, fix is not trivial, but since we don't do it in u-ring and therefore behaves differently, this is a bug. We're not using realpath(3) in Uring.

lib_eio/path.ml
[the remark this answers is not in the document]
talex5: As far as I know, that doesn't apply to bigarrays like cstruct.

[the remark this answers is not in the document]
talex5: Also, it occurs to me that, since we're opening the file anyway, it would be simpler to use `File.stat` here, not `Path.stat`.

lib_eio_luv/eio_luv.ml
haesbaert: https://github.com/aantron/luv/issues/134

lib_eio_luv/eio_luv.ml
haesbaert: I've checked the actual posix specification nowand just the lower 3 bits are guarenteed to have those values, which means I had to add the discover infrastructure.

lib_eio_luv/eio_luv.ml
haesbaert: I've removed this in favor of an association list

lib_eio/path.ml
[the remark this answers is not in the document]
haesbaert: I did check and bigarray is constrained by the that limit, but I might have been mistaken I'll check again.

I move it to `File`, problem is I can't call it one unix domain socket, I've added the method to `File.ro`, to have it in socket it would have to be deeper, I'm not sure what is preferred here ?

haesbaert:
This is a bit in WIP, I think I've addressed all points but I'm not too sure about File.stat

lib_eio/file.ml
```diff
@@ -1,8 +1,47 @@
+(** Tranditional Unix permissions. *)
+module Unix_perm = struct
+  type t = int
+  (** This is the same as {!Unix.file_perm}, but avoids a dependency on [Unix]. *)
```
avsm, line 4: i'm not so keen on this in the portable `eio` core, as we may just have very simple filesystem backends.  It would be nice to not have to dummy out `Stat.t` with fake values for uid/gid, for example.

haesbaert: But this was in `Path`, `Path` is already part of the portable core, or did I misunderstand something ?

I'm unsure where to put the Stat, I was thinking on adding a new effect on `Eio_unix.Private`:
    `| Stat : <unix_fd> -> Stat.t Effect.t`

Does that make sense, technically all we need is a fd, or should I add to `Eio_unix.unix_fd` directly ?

talex5: There's no problem with having the concept of "Unix permission" itself in the core - even non-Unix systems will often want to access Unix filesystems, tarballs, etc.

A more interesting question is whether we should add type parameters to `dir` to cover different types of file-system. e.g. our `dir` could become `Unix_perm.t dir`. However, that could be annoying, and any FS that can be mounted on Linux already has some suitable mapping.

Linux's `statx(2)` already allows you to do a stat that might or might not return some values (e.g. you can say you don't care about the UID and it is free to leave that field blank). So ideally we want a way to indicate which fields are provided even there.

haesbaert:
I had to step back a bit on this diff, I've rebased and force-pushed, from the new commit message:
I apologize for wasting everyone's time and radically changing it, but the whole luv bits were too insane for my taste, see details below.
I might be doing the whole interface/OO thing wrong and would appreciate some pointers here.
```
    Expose `stat` and improve Path.load

    `stat` can be called on files and flows in order to discover what is the backing
    descriptor. Some notes:
      - Use Unix.Largestat instead of Unix.stat
      - Add a portable stat for the user
      - Moved Path.Unix_perm into File.Unix_perm to avoid cyclical ref

    A previous attempt of this diff tried to make proper use of Luv.File.Stat but
    there were multiple issues: `mode` was not properly exported so I had to add a
    discovery.ml; it only works for a subset of file types and we had to manually
    convert timestamps to float; all doable but an insufferable amount of code to end
    up with less functionality than just calling into Unix.Largestat.stat.

    There are still some kinks:
      - Luv will always follow a link, so we can't really stat on a symbolic link,
        this is because we rely on realpath(3) which resolves the link.
      - We can't open named pipes in with luv atm, so we also can't stat it (#357).
      - I did test all file-modes, but didn't add all to the test since they are
        impractical (opening /dev/tty for Character_special...)
```

haesbaert:
> > I apologize for wasting everyone's time and radically changing it, but the whole luv bits were too insane for my taste, see details below.
> 
> Sounds good. Do we really need to add it to the socket types, though? It doesn't seem useful for anything.

Frankly, can't see much uses, and can only remember one time I relied on it, I've added it more for the sake of correctness/completeness, and to use it as an excuse to learn some ocaml OO. I can gut it if it's the case.

talex5:
> Frankly, can't see much uses, and can only remember one time I relied on it, I've added it more for the sake of correctness/completeness, and to use it as an excuse to learn some ocaml OO. I can gut it if it's the case.

Yes please. It will be a pain for e.g. Mirage having to support fstat.

haesbaert:
> > Frankly, can't see much uses, and can only remember one time I relied on it, I've added it more for the sake of correctness/completeness, and to use it as an excuse to learn some ocaml OO. I can gut it if it's the case.
> 
> Yes please. It will be a pain for e.g. Mirage having to support fstat.

Ack, leme know if this is ok and then I'll squash and merge.

lib_eio/path.ml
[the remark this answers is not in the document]
talex5: > I did check and bigarray is constrained by the that limit, but I might have been mistaken I'll check again.

Could you post a link to the code. It seems very surprising to me that a "big array" is the same size as a normal one.

In particular, https://v2.ocaml.org/api/Bigarray.html says:

> Bigarrays are not limited in size, unlike OCaml arrays.

lib_eio/file.ml
```diff
@@ -11,6 +53,12 @@ class virtual rw = object (_ : <Generic.t; Flow.source; Flow.sink; ..>)
   method virtual pwrite : file_offset:Optint.Int63.t -> Cstruct.t list -> int
 end
 
+(**  [stat t] returns the {!Stat.t} record associated with [t]. *)
```
talex5, line 56:
```suggestion
(** [stat t] returns the {!Stat.t} record associated with [t]. *)
```

lib_eio/fs.ml
```diff
@@ -1,23 +1,18 @@
 (** Defines types used by file-systems. *)
 
-(** Tranditional Unix permissions. *)
-module Unix_perm = struct
```
talex5, line 4: We could put a (deprecated) alias here, to avoid breaking existing code.

lib_eio/path.ml
```diff
@@ -30,10 +30,21 @@ let with_lines path fn =
   let buf = Buf_read.of_flow flow ~max_size:max_int in
   fn (Buf_read.lines buf)
 
-let load path =
-  with_open_in path @@ fun flow ->
-  let buf = Buf_read.of_flow flow ~max_size:max_int in
-  Buf_read.take_all buf
+let load ((t:#Fs.dir), path) =
+  with_open_in (t, path) @@ fun flow ->
```
talex5, line 34:
```suggestion
let load t =
  with_open_in t @@ fun flow ->
```

lib_eio_linux/eio_linux.ml
```diff
@@ -36,6 +36,7 @@ let system_thread = Ctf.mint_id ()
 let wrap_errors path fn =
   try fn () with
   | Unix.Unix_error(Unix.EEXIST, _, _) as ex -> raise @@ Eio.Fs.Already_exists (path, ex)
+  | Unix.Unix_error(Unix.E2BIG, _, _)  as ex -> raise @@ Eio.Fs.File_too_large (path, ex)
```
talex5, line 39: Do we need this? I'd guess that `E2BIG` doesn't always mean "file too large".

lib_eio_linux/eio_linux.ml
```diff
@@ -110,6 +111,36 @@ module FD = struct
...
+      ino     = ust.st_ino   |> Int64.of_int;
+      kind    = st_kind;
+      perm    = ust.st_perm;
+      nlink   = ust.st_nlink |> Int64.of_int;
+      uid     = ust.st_uid   |> Int64.of_int;
+      gid     = ust.st_gid   |> Int64.of_int;
+      rdev    = ust.st_rdev  |> Int64.of_int;
+      size    = ust.st_size  |> Optint.Int63.of_int64;
+      blksize = 512          |> Int64.of_int;         (* XXX *)
+      blocks  = Int64.(div ust.st_size (of_int 512)); (* XXX *)
```
talex5, line 139: What's the plan for dealing with these?

lib_eio_linux/tests/eurcp_lib.ml
```diff
@@ -31,7 +31,7 @@ let run_cp block_size queue_depth infile outfile () =
   let open Unix in
   let infd = U.openfile ~sw infile [O_RDONLY] 0 in
   let outfd = U.openfile ~sw outfile [O_WRONLY; O_CREAT; O_TRUNC] 0o644 in
-  let insize = U.fstat infd |> fun {st_size; _} -> Int63.of_int st_size in
+  let insize = U.fstat infd |> fun {size; _} -> size in
```
talex5, line 34:
```suggestion
  let insize = (U.fstat infd).size in
```

tests/fs.md
talex5: I guess we should just do the `realpath` on the parent, then? It's fine to lstat a symlink even if we don't have access to its target.

lib_eio_linux/eio_linux.ml
haesbaert: I think we should just zap them, I don't see us adding stubs just to fetch those. When I added the Luv would expose the real ones at least.

tests/fs.md
haesbaert: Oh that's smart, I hadn't considered it.

lib_eio_linux/eio_linux.ml
haesbaert: zapped, no idea what I was thinking

tests/fs.md
haesbaert: I won't address the `realpath` on the parent here, it's outside of the scope imho.

lib_eio/path.ml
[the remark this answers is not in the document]
haesbaert: > > I did check and bigarray is constrained by the that limit, but I might have been mistaken I'll check again.
> 
> Could you post a link to the code. It seems very surprising to me that a "big array" is the same size as a normal one.
> 
> In particular, https://v2.ocaml.org/api/Bigarray.html says:
> 
> > Bigarrays are not limited in size, unlike OCaml arrays.

You're absolutely correct, I think my confusion came from the fact that Bigarray was using Array underneath and I assumed the same, but it's not:

`caml_make_vect` (Array.create) checks for `Max_wosize` where Bigarray's `caml_ba_create` does not. I've removed the test.

lib_eio/path.ml
talex5: This isn't resolved. Keeping it as one will also fix the problem that the string in the error loses the dir tag, since you can then use `Path.pp`.

(we might need a bit of variable renaming in this module; lots of functions here are using `t` for the directory and `path` for a full path, which is now a bit backwards)

tests/fs.md
talex5: Yes, now `Path.stat` is gone - we can worry about that later :-) PR title needs updating too...

talex5 (approved):
This looks pretty good now! (Would be nice to get the dir tag in the error message though). Can be merged once rebased / squashed.

lib_eio/path.ml
haesbaert: I'm a bit confused, we don't really include the dir in the error string for any of the Eio.Fs exceptions:
```
let open_in ~sw ((t:#Fs.dir), path) = t#open_in ~sw path

  method open_in ~sw path =
    let fd = File.open_ ~sw (self#resolve path) [`NOFOLLOW; `RDONLY] |> or_raise_path path in
    (flow fd :> <Eio.File.ro; Eio.Flow.close>)                                        ^^^^
```
```
utop # let _ = Eio_luv.run (fun env -> let cwd = Eio.Stdenv.cwd env in Eio.Switch.run @@ fun sw -> Eio.Path.load Eio.Path.(cwd / "not_there"));;
Exception: Eio__Fs.Not_found ("not_there", Eio_luv.Luv_error 877502405).

let _ = Eio_linux.run (fun env -> let cwd = Eio.Stdenv.cwd env in Eio.Switch.run @@ fun sw -> Eio.Path.load Eio.Path.(cwd / "not_there"));;
Exception:
Eio__Fs.Not_found ("not_there", Unix.Unix_error (Unix.ENOENT, "openat2", "")).
```

talex5: OK, let's sort that out later then! It just seemed a shame to be splitting the path up only for the purpose of producing a slightly less informative error message.

PR #343, opened 2022-10-10
Split the reads into no more than 2^32-1 for luv

### LINKED ISSUE #335: Eio.Path.load loads only first 4gb on big files
I'm using the Luv backend and when trying to load a file bigger than 4gb's the result string always ends up being the first 4gb's of the file.

talex5:
Luv uses a 32 bit int for buffer sizes and wraps if the value passed is too big. In particular, a request for to read 4GB of data is interpreted as a request for 0 bytes.

@haesbaert this is based on your [commit](https://github.com/ocaml-multicore/eio/commit/e2f38d35493f961d3efdc05f37a5a131a0e5fd03), except:
- For vectors, we split long buffers into multiple buffers instead of truncating them (which could skip data).
- For datagram sends and receives, overlong buffers are an error (instead of truncating).
- The conversations are used in more places.

Fixes #335.

### DIFF
--- a/lib_eio_luv/eio_luv.ml
+++ b/lib_eio_luv/eio_luv.ml
@@ -56,6 +56,31 @@ let or_raise_path path = function
   | Ok x -> x
   | Error e -> raise (wrap_error ~path e)
 
+(* Luv can't handle buffers with more than 2^32-1 bytes, limit it to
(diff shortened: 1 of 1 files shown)


haesbaert:
LGTM, I hadn't checked all the necessary places before, but that looks correct.

PR #350, opened 2022-10-19
Add pipe to eio_unix

patricoferris:
This adds `Eio_unix.pipe` to create a connected pair of flows similar to `socketpair`, this is part of #330

### DIFF
--- a/lib_eio_luv/eio_luv.ml
+++ b/lib_eio_luv/eio_luv.ml
@@ -596,6 +596,7 @@ let get_fd_opt t = Eio.Generic.probe t FD
 let flow fd = object (_ : <source; sink; ..>)
   method fd = fd
   method close = Low_level.File.close fd
+  method unix_fd op = File.to_unix op fd
 
   method probe : type a. a Eio.Generic.ty -> a option = function
     | FD -> Some fd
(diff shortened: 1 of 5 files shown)


lib_eio_luv/eio_luv.ml
```diff
@@ -1155,6 +1156,14 @@ let rec run : type a. (_ -> a) -> a = fun main ->
             with Luv_error _ as ex ->
               discontinue k ex
           )
+          | Eio_unix.Private.Pipe sw -> Some (fun k ->
+            let r, w = Unix.pipe ~cloexec:true () in
```
haesbaert, line 1160: libuv has pipes, can we use that ?
https://github.com/aantron/luv/blob/0ed668e87ca55d65f4f84422b0dd3381701c71b0/src/pipe.mli#L40

patricoferris: Thanks for the speedy review! :))

Yes, I think so but I had to specify no read and write flags (see [uv_pipe](http://docs.libuv.org/en/v1.x/pipe.html#c.uv_pipe)) so it ended up being

```ocaml
let r, w = Luv.Pipe.pipe ~read_flags:[] ~write_flags:[] () |> or_raise in
```

Do you think that's okay? (They're opened with cloexec `true`).

haesbaert: I think it's fine, the only flag it accepts is `nonblock`.

tests/flow.md
```diff
@@ -118,3 +118,25 @@ Copying from src using `Read_source_buffer`:
...
+Writing to and reading from a pipe.
+
+```ocaml
+# Eio_main.run @@ fun env ->
+  Switch.run @@ fun sw ->
+  let r, w = Eio_unix.pipe sw in
+  Eio.Fiber.both
+    (fun () ->
+      let buf = Cstruct.create 32 in
+      let r = Eio.Flow.read r buf in
```
talex5, line 133: In theory, this is not reliable (though it seems unlikely to fail). Should use `read_exact` here, or `Buf_read`.

patricoferris: Indeed thanks! I've rebased to tidy up the history and used `read_exact`.

talex5:
Thanks!

PR #352, opened 2022-10-20
Expose getaddrinfo errors. Was: Return empty list on EAI_ errors for getaddrinfo@luv (#351)

### LINKED ISSUE #351: Getaddrinfo raises error on eio_luv
It seems that luv raises an exception instead of returning an empty list. 

```ocaml
open Eio

let lookup net hostname =
   match Net.getaddrinfo_stream net hostname with
   | [] -> ()
   | _ -> assert false

let () =
  let hostname = "blahblahblah.tarides.com" in
  (assert (Unix.getaddrinfo hostname "" [] = []));
  Eio_main.run (fun env -> lookup env#net hostname)
```

This returns:

```ocaml
Fatal error: exception Eio_luv.Luv_error(EAI_NONAME) (* unknown node or service *)
```

I can't currently run this properly on Linux with Uring but I'm assuming it returns `[]` because `Unix.getaddrinfo` does.

haesbaert:
Unix.getaddrinfo used in u-ring ignores gar_errno and returns an empty list, this makes the luv backend follow a similar behaviour.

Should fix #351 where we would get EAI_NONAME (or others) if resolution failed, this is a bit defensive as we just handle expected EAI errors.


haesbaert:
cc @patricoferris

lib_eio_luv/eio_luv.ml
```diff
@@ -575,10 +575,15 @@ module Low_level = struct
...
-    |> List.filter_map to_eio_sockaddr_t
-
+    match (await_with_cancel ~request
+             (fun loop -> Luv.DNS.getaddrinfo ~loop ~request ~service ~node ()))
+    with
+    | Ok nl -> List.filter_map to_eio_sockaddr_t nl
+    | Error `EAI_ADDRFAMILY | Error `EAI_AGAIN | Error `EAI_BADFLAGS | Error `EAI_BADHINTS
+    | Error `EAI_CANCELED   | Error `EAI_FAIL  | Error `EAI_FAMILY   | Error `EAI_MEMORY
+    | Error `EAI_NODATA     | Error `EAI_NONAME| Error `EAI_OVERFLOW | Error `EAI_PROTOCOL
+    | Error `EAI_SERVICE    | Error `EAI_SOCKTYPE -> []
```
talex5, line 585: This list seems excessive (even if it does match the stdlib version). In particular, `EAI_AGAIN`, `EAI_BADFLAGS`, `EAI_MEMORY` and `EAI_SOCKTYPE` seem like they should be reported as errors, rather than claiming there are no addresses. Maybe `EAI_FAIL` too.

talex5:
> I'll differ to @talex5's knowledge of which EAI_ errors to catch and which to no :))

I have no special knowledge - just reading the `getaddrinfo` man-page :-)

lib_eio_luv/eio_luv.ml
haesbaert: I thought about it too, just some notes:
EAI_FAIL and EAI_AGAIN can be the result of a an actual DNS reply (would have to dig deeper, but those should return an empty list). EAI_SOCKTYPE, EAI_BADFLAGS, EAI_FAMILY we could, since they are mostly programming errors from our side.
I don't have a strong opinion, but I'd err on the side of just emulating the same behaviour of `Unix.getaddrinfo` which ignores everything.

talex5: I wonder if we should just copy the `getaddrinfo` API and not allow an empty result. The wrapper function could enforce this:

```ocaml
let getaddrinfo ?(service="") (t:#t) hostname =
  match t#getaddrinfo ~service hostname with
  | [] -> raise (Dns_error (Failure "No addresses for ..."))
  | xs -> xs
```

I guess most people getting an empty list will want to report an error, and it's a bit silly you have to guess what the error was when we just threw the real error away. e.g. in `with_tcp_connect` we currently do:

```ocaml
    | [] -> raise (Connection_failure (Failure (Fmt.str "No TCP addresses for %S" host)))
```
Then we could say if that's because the DNS server is busy, the entry doesn't exist, doesn't have an IP address, etc.

haesbaert: Also `Dns_error` is a bit imprecise since `getaddrinfo` is not only about DNS (`AI_NUMERICHOST`and so on)

haesbaert: Thinking a bit more about this, I think `connect` should take a `Sockaddr.stream list`, and loop over the candidates until it succeeds connecting to one. This is virtually what everyone does with `getaddrinfo` + `connect` and `bind`.
In this case `connect` itself could raise something, `getaddrinfo` could still return just an empty list, and the user would still have to handle an exception.

talex5: Yes, the error code was just an example. We can use a better exception name and contents.

> I think connect should take a Sockaddr.stream list, and loop over the candidates until it succeeds connecting to one.

`with_tcp_connect` already handles looping over the candidates; I'm not sure there's much benefit to changing `connect` too. In fact, that would just mean having to do something for the extra case of an empty list.

> In this case connect itself could raise something, getaddrinfo could still return just an empty list, and the user would still have to handle an exception.

I don't see how that helps. We'd still be losing the detailed error from `getaddrinfo` and replacing it with a guess.

haesbaert: can I get a ruling here ?

talex5: > Should we just raise all errors from `getaddrinfo` on both cases?

Yes, that makes sense to me.

( Having `connect` take a `Sockaddr.stream list` doesn't seem useful to me. You can't give a good error if there are no addresses - you don't know what the DNS error was, what the failing hostname was, or even whether DNS was used at all! )

haesbaert:
Some corrections to my previous statement, getaddrinfo(3) *cannot* return an empty list and no error, it's either something or an error. I checked libc.

This PR should be ready for reviewing, some notes:
- A lot of the stub is just copied from ocaml's stdlib
- The values chosen for the undefined `EAI_*` values are taken from libuv (since it works for them it works for us), they're added for completion sakeness and a future guard. I'm sure once we have some other backend those will popup and we will move the stub, or maybe glibc adds them in the future, who knows.
- The strings for `getaddrinfo_error_to_string` were copied verbatim from libc.

haesbaert:
Tests are zefixed, marking this as ready for review

lib_eio_linux/getaddrinfo_stubs.c
```diff
@@ -0,0 +1,187 @@
...
+/*                                                                        */
+/*             Xavier Leroy, projet Cristal, INRIA Rocquencourt,          */
+/*             Christiano Haesbaert, Tarides                              */
+/*   Copyright 2004 Institut National de Recherche en Informatique et     */
+/*     en Automatique.                                                    */
+/*   Copyright 2022 Tarides                                               */
+/*                                                                        */
+/*   All rights reserved.  This file is distributed under the terms of    */
+/*   the GNU Lesser General Public License version 2.1, with the          */
+/*   special exception on linking described in the file LICENSE.          */
```
talex5, line 13: I think we want to avoid LGPL code in Eio (it's currently BSD and ISC only). Though they might relicense this bit if asked.

talex5 (commented):
Looks sensible.

Note that I'm changing the error handling quite a lot in #378, which will impact this PR a bit.

lib_eio_linux/getaddrinfo_stubs.c
haesbaert: @gasche @xavierleroy (sorry for the ping).
I've copied significant part of `otherlibs/unix/getaddrinfo.c` into EIO (https://github.com/ocaml-multicore/eio/blob/61b737d5be683b9be611844b16c7e73b5bece09a/lib_eio_linux/getaddrinfo_stubs.c). The original file is LGPLed and we would like to relicense our new file to ISC.
Do you permit us to relicense the bits we copied out from LGPL to ISC in EIO ?

lib_eio_linux/getaddrinfo_stubs.c
```diff
@@ -0,0 +1,187 @@
...
+#include <errno.h>
+#include <netdb.h>
+
+#include <caml/mlvalues.h>
+#include <caml/memory.h>
+#include <caml/alloc.h>
+#include <caml/unixsupport.h>
+#include <caml/socketaddr.h>
+
+static value caml_unix_cst_to_constr(int n, int *tbl, int size, int deflt)
```
dra27, line 29: You could avoid this duplication by borrowing the `extern` declaration in `cst2constr.h`

lib_eio_linux/getaddrinfo_stubs.c
```diff
@@ -0,0 +1,187 @@
...
+  retcode = getaddrinfo(node, serv, &hints, &res);
+  caml_leave_blocking_section();
+  if (node != NULL) caml_stat_free(node);
+  if (serv != NULL) caml_stat_free(serv);
+  /* Convert result */
+  vres = Val_emptylist;
+  if (retcode == 0) {
+    for (r = res; r != NULL; r = r->ai_next) {
+      v = caml_alloc_small(2, Tag_cons);
+      Field(v, 0) = convert_addrinfo(r);
```
dra27, line 166: This violates GC rule 5 - the `convert_addrinfo` call needs to be done before the `caml_alloc_small` and stored in a registered local (`convert_addrinfo` allocates).

dra27 (commented):
Drive-by comments on the borrowed C code!

lib_eio_linux/getaddrinfo_stubs.c
haesbaert: thanks for spotting this, fixed it now

lib_eio_linux/getaddrinfo_stubs.c
haesbaert: fixed, somehow my brain remembered the function being static in first place.

lib_eio_linux/getaddrinfo_stubs.c
xavierleroy: No.  I only use copyleft licenses for my free software.  Please consider using the LGPL with OCaml linking exception for your library, or do not reuse any of my code.

haesbaert: > No. I only use copyleft licenses for my free software. Please consider using the LGPL with OCaml linking exception for your library, or do not reuse any of my code.

Thank you, will do one of the two as suggested.

talex5:
I updated this code (with fresh C stubs) and it's now merged as #866 - thanks!

PR #353, opened 2022-10-20
Rename `Flow.read` to `Flow.single_read`

talex5:
This is a low-level function and it is easy to use it incorrectly by ignoring the possibility of short reads. The name `Flow.read` made it sound like a suitable default way to read from a flow.

See discussion in https://github.com/ocaml-multicore/eio/pull/344.

/cc @haesbaert


haesbaert:
me gusta !

haesbaert (approved):

PR #355, opened 2022-10-24
Add Eio_unix.pipe workaround, zap Eio_linux.pipe

haesbaert:
Make sure we workaround uring pipe bug from #319 in Eio_unix.pipe. Zap Eio_linux.pipe since it's only used in tests.

cc @patricoferris

### DIFF
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -1292,15 +1292,6 @@ let stdenv ~run_event_loop =
     method debug = Eio.Private.Debug.v
   end
 
-let pipe sw =
-  let r, w = Unix.pipe () in
-  (* XXX workaround for issue #319, PR #327 *)
-  Unix.set_nonblock r;
-  Unix.set_nonblock w;
-  let r = source (FD.of_unix ~sw ~seekable:false ~close_unix:true r) in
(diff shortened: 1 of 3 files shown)


lib_eio_linux/eio_linux.ml
```diff
@@ -1455,6 +1446,9 @@ let rec run : type a.
             )
           | Eio_unix.Private.Pipe sw -> Some (fun k ->
               let r, w = Unix.pipe ~cloexec:true () in
+              (* XXX workaround for issue #319, PR #327 *)
```
talex5, line 1449: It's not exactly a work-around, though. We'd prefer them non-blocking even without the bug, right? So there's nothing to fix, and no need for `XXX` here.

haesbaert: Agreed, but it's the only case where we're setting a FD to nonblocking, I'd keep the issue mention at least.

talex5: The mention of the reason is good, but get rid of the `XXX`. Otherwise people will wonder in future what they're supposed to be fixing here.

haesbaert: fixed

talex5 (approved):
The test code is a bit ugly, but it doesn't really matter. We could use the new `Flow.write` now, but I guess we're still missing a high-level API for `readv`.

PR #359, opened 2022-10-29
eio_luv: add low-level process support

patricoferris:
This PR adds low-level process support to Eio_luv as per the suggestion in #330. A few questions:

1. Because this is the low-level API is it okay to just return `(int * int64)` from the callback function as the exit status (see https://aantron.github.io/luv/luv/Luv/Process/index.html#val-spawn for an explanation) ?
2. Should the low-level API use switches to protect against leaking a process spawn ?
3. I've exposed `Eio_luv.Low_level.Stream` is that alright (I don't know if it was just because there was no need for it or if there is some other reason it wasn't exposed) ?

### DIFF
--- a/lib_eio_luv/eio_luv.ml
+++ b/lib_eio_luv/eio_luv.ml
@@ -574,6 +574,51 @@ module Low_level = struct
       Handle.of_luv ~sw sock
   end
 
+  module Pipe = struct
+    type t = [`Stream of [`Pipe]] Handle.t
+
+    let init ?for_handle_passing ~sw () =
+      Luv.Pipe.init ~loop:(get_loop ()) ?for_handle_passing ()
+      |> or_raise
+      |> Handle.of_luv ~close_unix:true ~sw
+  end
+
+  module Process = struct
+    type t = {
+      proc : Luv.Process.t;
+      status : (int * int64) Promise.t;
+    }
+
+    let pid t = Luv.Process.pid t.proc
+
+    let to_parent_pipe ?readable_in_child ?writable_in_child ?overlapped ~fd ~parent_pipe () =
+      let parent_pipe = Handle.to_luv parent_pipe in
+      Luv.Process.to_parent_pipe ?readable_in_child ?writable_in_child ?overlapped ~fd ~parent_pipe ()
+
+    let await_exit t = Promise.await t.status
+
+    let has_exited t = Promise.is_resolved t.status
+
+    let send_signal t i = Luv.Process.kill t.proc i |> or_raise
+
+    let spawn ?cwd ?env ?uid ?gid ?(redirect=[]) ~sw cmd args =
+      let status, set_status = Promise.create () in
+      let hook = ref None in
+      let on_exit _ ~exit_status ~term_signal =
+        Option.iter Switch.remove_hook !hook;
+        Promise.resolve set_status (term_signal, exit_status)
+      in
+      let proc = Luv.Process.spawn ?environment:env ?uid ?gid ~loop:(get_loop ()) ?working_directory:cwd ~redirect ~on_exit cmd args |> or_raise in
+      if not (Promise.is_resolved status) then (
+        let h = Switch.on_release_cancellable sw (fun () ->
+            Luv.Process.kill proc Luv.Signal.sigkill |> or_raise;
+            ignore (Promise.await status)
+          ) in
+        hook := Some h
+      );
+    { proc; status }
+  end
+
   module Poll = Poll
 
   let sleep_ms delay =
--- a/lib_eio_luv/eio_luv.mli
+++ b/lib_eio_luv/eio_luv.mli
@@ -143,6 +143,58 @@ module Low_level : sig
     val write : [ `Stream of [< `Pipe | `TCP | `TTY ] ] Handle.t -> Luv.Buffer.t list -> unit
     (** [write handle bufs] writes the contents of [bufs] to [handle]. *)
   end
+
+  module Pipe : sig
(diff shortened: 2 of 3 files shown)


lib_eio_luv/eio_luv.ml
```diff
@@ -543,6 +543,44 @@ module Low_level = struct
...
+
+    let to_parent_pipe ?readable_in_child ?writable_in_child ?overlapped ~fd ~parent_pipe () =
+      let parent_pipe = Pipe.to_luv parent_pipe in
+      Luv.Process.to_parent_pipe ?readable_in_child ?writable_in_child ?overlapped ~fd ~parent_pipe ()
+
+    let inherit_fd = Luv.Process.inherit_fd
+
+    let await_exit (_, status) = Promise.await status
+
+    let stop (t, _) = Luv.Process.kill t Luv.Signal.sigkill |> or_raise
```
anmonteiro, line 573: this could probably be implemented on top of a more generic `val kill: t -> signal -> unit` that could allow sending a signal to the subprocess.

patricoferris: Good point thanks! I suppose as a low-level API it's enough to just give people the power to send whatever signals they want. I added this under a `send_signal` function which I prefer as a name over `kill` :))

talex5 (commented):
Thanks - looks good :-)

> Because this is the low-level API is it okay to just return (int * int64) from the callback function as the exit status (see https://aantron.github.io/luv/luv/Luv/Process/index.html#val-spawn for an explanation)?

Yes, I think that's fine.

> Should the low-level API use switches to protect against leaking a process spawn ?

Yes.

> I've exposed Eio_luv.Low_level.Stream is that alright (I don't know if it was just because there was no need for it or if there is some other reason it wasn't exposed) ?

Probably just no need before.

lib_eio_luv/eio_luv.mli
```diff
@@ -114,6 +114,69 @@ module Low_level : sig
...
+
+    val to_handle : ?close_unix:bool -> sw:Switch.t -> t -> [`Stream of [`Pipe]] Handle.t
+    (** Converts a pipe to a {!Handle.t}*)
+  end
+
+  module Process : sig
+    type t
+    (** A process *)
+
+    type redirection
```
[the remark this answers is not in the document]
patricoferris: Yep that's right, also why it has a separate type. The problem I seemed to have is that things won't work if the pipe initialisation doesn't contain the loop i.e.`Luv.Pipe.init ~loop:(get_loop ())`

lib_eio_luv/eio_luv.mli
```diff
@@ -114,6 +114,69 @@ module Low_level : sig
         @param sw The handle is closed when [sw] is released, if not closed manually first.
         @param close_unix if [true] (the default), calling [close] also closes [fd]. *)
   end
+
+  module Pipe : sig
+    type t
```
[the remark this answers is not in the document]
patricoferris: See comment about redirections below

lib_eio_luv/tests/process.md
```diff
@@ -0,0 +1,87 @@
...
+        read (acc + i)
+      with End_of_file -> acc
+    in read 0
+```
+
+A simple `echo hello` process redirects to stdout.
+
+```ocaml
+# Eio_luv.run @@ fun _env ->
+  Switch.run @@ fun sw ->
```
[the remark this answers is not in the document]
patricoferris: Yep, I hadn't pulled the use of switches in yet, fixed in https://github.com/ocaml-multicore/eio/pull/359/commits/6e681fec1f5894c29c28a042ab107675dd0fd518

lib_eio_luv/eio_luv.ml
```diff
@@ -543,6 +543,51 @@ module Low_level = struct
...
+      status : (int * int64) Promise.t;
+    }
+
+    let pid t = Luv.Process.pid t.proc
+
+    let to_parent_pipe ?readable_in_child ?writable_in_child ?overlapped ~fd ~parent_pipe () =
+      let parent_pipe = Handle.to_luv parent_pipe in
+      Luv.Process.to_parent_pipe ?readable_in_child ?writable_in_child ?overlapped ~fd ~parent_pipe ()
+
+    let await_exit t = Promise.await t.status
```
talex5, line 567: I moved the `remove_hook` to `spawn`. It needs to be called even if the application doesn't wait.

lib_eio_luv/eio_luv.ml
```diff
@@ -543,6 +543,51 @@ module Low_level = struct
...
+      let hook = ref None in
+      let on_exit _ ~exit_status ~term_signal =
+        Option.iter Switch.remove_hook !hook;
+        Promise.resolve set_status (term_signal, exit_status)
+      in
+      let proc = Luv.Process.spawn ?environment:env ?uid ?gid ~loop:(get_loop ()) ?working_directory:cwd ~redirect ~on_exit cmd args |> or_raise in
+      if not (Promise.is_resolved status) then (
+        let h = Switch.on_release_cancellable sw (fun () ->
+            Luv.Process.kill proc Luv.Signal.sigkill |> or_raise;
+            ignore (Promise.await status)
```
talex5, line 584: Added a wait here just so we can be sure the process has completely finished stopping by the time the switch finishes.

talex5 (approved):
I pushed a few more commits to tidy things up. If you're happy with my changes, this is ready to merge.

patricoferris:
Thanks! The changes in #371 plus these all look good to me :))

talex5:
Thanks!

Issue #341, opened 2022-10-07
Can't use unbound UDP sockets


haesbaert:
Currently we're forcing a bind in `Eio.Net.datagram_socket` by obliging the user to pass a `sockaddr`, this should be optional, an unbound UDP socket binds on the first `sendto` and the kernel chooses an ephemeral port.

haesbaert:
Fixed in #360

Issue #322, opened 2022-09-23
Support for socket options


anmonteiro:
Right now both `Eio.Net.{connect,accept}` return a socket, but I didn't find a way to e.g. set `TCP_NODELAY` on the stream socket. Did I miss an API?

talex5:
Looks like it's missing, indeed.

I was using it originally with the HTTP benchmark, but now that sends everything in a single packet it doesn't make any difference there.

As a work-around, you can of course get the FD directly:
```
Unix.setsockopt (Eio_unix.FD.peek_opt ssock |> Option.get) Unix.TCP_NODELAY true;
```

create2000:
Hello @patricoferris . My name is Anthony Onah. I will be contributing on this issue. I hope with your guidance and support, i should be able to resolve this.

talex5:
Note there is an old draft PR related to this at https://github.com/ocaml-multicore/eio/pull/575

create2000:
@talex5. In rebasing it to the current main, are you proposing I start with the previous issue? Also, is it a good first issue? 

I will love to contribute.

talex5:
Yes, I think that PR was going in the right direction. It looks like it just never got finished.

rbjorklin:
Leaving a breadcrumb for any future folks who find their way here.

The way to do this now seems to be by using `Eio_unix.Net.fd` and `Eio_unix.Fd.use`.

talex5:
Fixed in #859.

PR #360, opened 2022-10-31
Allow datagram_socket to be created without address

### LINKED ISSUE #345: Can't use connected UDP sockets
Usually you can "connect" a UDP socket and use normal writes and reads.
In former times a sendto on UDP was a connect+write, at the very least this should allow the kernel to cache a route lookup and etc.

bikallem:
This PR allows to create UDP `datagram_socket` without passing in the socket address. Being able to create `datagram_socket` without socket address is useful in udp client applications where it is convenient to send data to UDP severs without having to figure out the udp client address first. 

Fixes #345 #341

### DIFF
--- a/lib_eio/net.ml
+++ b/lib_eio/net.ml
@@ -195,7 +195,7 @@ let listen ?(reuse_addr=false) ?(reuse_port=false) ~backlog ~sw (t:#t) = t#liste
 let connect ~sw (t:#t) = t#connect ~sw
 
 let datagram_socket ?(reuse_addr=false) ?(reuse_port=false) ~sw (t:#t) addr =
-  let addr = (addr :> [ Sockaddr.datagram | `UdpV4 | `UdpV6]) in 
+  let addr = (addr :> [Sockaddr.datagram | `UdpV4 | `UdpV6]) in 
   t#datagram_socket ~reuse_addr ~reuse_port ~sw addr
 
 let getaddrinfo ?(service="") (t:#t) hostname = t#getaddrinfo ~service hostname
--- a/lib_eio/net.ml
+++ b/lib_eio/net.ml
@@ -181,7 +181,9 @@ class virtual t = object
   method virtual listen : reuse_addr:bool -> reuse_port:bool -> backlog:int -> sw:Switch.t -> Sockaddr.stream -> listening_socket
   method virtual connect : sw:Switch.t -> Sockaddr.stream -> <stream_socket; Flow.close>
   method virtual datagram_socket :
-    sw:Switch.t
+       reuse_addr:bool
+    -> reuse_port:bool
+    -> sw:Switch.t
     -> [Sockaddr.datagram | `UdpV4 | `UdpV6]
     -> <datagram_socket; Flow.close>
 
@@ -192,11 +194,9 @@ end
 let listen ?(reuse_addr=false) ?(reuse_port=false) ~backlog ~sw (t:#t) = t#listen ~reuse_addr ~reuse_port ~backlog ~sw
 let connect ~sw (t:#t) = t#connect ~sw
 
-type udp_socket = [ Sockaddr.datagram | `UdpV4 | `UdpV6]
-
-let datagram_socket ~sw (t:#t) addr =
-  let addr = (addr :> udp_socket) in 
-  t#datagram_socket ~sw addr
+let datagram_socket ?(reuse_addr=false) ?(reuse_port=false) ~sw (t:#t) addr =
+  let addr = (addr :> [ Sockaddr.datagram | `UdpV4 | `UdpV6]) in 
+  t#datagram_socket ~reuse_addr ~reuse_port ~sw addr
 
 let getaddrinfo ?(service="") (t:#t) hostname = t#getaddrinfo ~service hostname
 
--- a/lib_eio/net.ml
+++ b/lib_eio/net.ml
@@ -180,7 +180,11 @@ let recv (t:#datagram_socket) = t#recv
 class virtual t = object
   method virtual listen : reuse_addr:bool -> reuse_port:bool -> backlog:int -> sw:Switch.t -> Sockaddr.stream -> listening_socket
   method virtual connect : sw:Switch.t -> Sockaddr.stream -> <stream_socket; Flow.close>
-  method virtual datagram_socket : sw:Switch.t -> Sockaddr.datagram -> <datagram_socket; Flow.close>
+  method virtual datagram_socket :
+    sw:Switch.t
+    -> [Sockaddr.datagram | `UdpV4 | `UdpV6]
+    -> <datagram_socket; Flow.close>
+
   method virtual getaddrinfo : service:string -> string -> Sockaddr.t list
   method virtual getnameinfo : Sockaddr.t -> (string * string)
 end
@@ -188,7 +192,7 @@ end
 let listen ?(reuse_addr=false) ?(reuse_port=false) ~backlog ~sw (t:#t) = t#listen ~reuse_addr ~reuse_port ~backlog ~sw
 let connect ~sw (t:#t) = t#connect ~sw
 
-let datagram_socket ~sw (t:#t) = t#datagram_socket ~sw
+let datagram_socket ~sw (t:#t) addr = t#datagram_socket addr ~sw
 
 let getaddrinfo ?(service="") (t:#t) hostname = t#getaddrinfo ~service hostname
 
--- a/lib_eio/net.ml
+++ b/lib_eio/net.ml
@@ -192,7 +192,11 @@ end
 let listen ?(reuse_addr=false) ?(reuse_port=false) ~backlog ~sw (t:#t) = t#listen ~reuse_addr ~reuse_port ~backlog ~sw
 let connect ~sw (t:#t) = t#connect ~sw
 
-let datagram_socket ~sw (t:#t) addr = t#datagram_socket addr ~sw
+type udp_socket = [ Sockaddr.datagram | `UdpV4 | `UdpV6]
+
+let datagram_socket ~sw (t:#t) addr =
+  let addr = (addr :> udp_socket) in 
+  t#datagram_socket ~sw addr
 
 let getaddrinfo ?(service="") (t:#t) hostname = t#getaddrinfo ~service hostname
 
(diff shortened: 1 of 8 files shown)


tests/network.md
```diff
@@ -165,42 +165,50 @@ Exception: Failure "Simulated error".
...
 Handling one UDP packet using IPv4:
 
+<!-- $MDX non-deterministic=output -->
 ```ocaml
-# run (run_dgram Eio.Net.Ipaddr.V4.loopback);;
+# let addr = Eio.Net.Ipaddr.V4.loopback in
+  run @@ run_dgram addr (fun ~net sw server_addr ->
+    let e1 = `UdpV4 in
+    let e = Eio.Net.datagram_socket ~sw net e1 in
+    traceln "Sending data from %s to %a" "UdpV4" Eio.Net.Sockaddr.pp server_addr;
```
haesbaert, line 191: I'd just cut the "from %s" here, just `Sending data to %a`

tests/network.md
```diff
@@ -165,42 +165,50 @@ Exception: Failure "Simulated error".
...
+    Eio.Net.send e server_addr (Cstruct.of_string "UDP Message"));;
 +Waiting to receive data on udp:127.0.0.1:8082
-+Sending data from udp:127.0.0.1:8081 to udp:127.0.0.1:8082
-+Received message from udp:127.0.0.1:8081: UDP Message
++Sending data from UdpV4 to udp:127.0.0.1:8082
```
haesbaert, line 194: same here

tests/network.md
```diff
@@ -165,42 +165,50 @@ Exception: Failure "Simulated error".
...
+    let e1 = `UdpV6 in
+    let e = Eio.Net.datagram_socket ~sw net e1 in
+    traceln "Sending data from %s to %a" "UdpV6" Eio.Net.Sockaddr.pp server_addr;
+    Eio.Net.send e server_addr (Cstruct.of_string "UDP Message"));;
 +Waiting to receive data on udp:[::1]:8082
-+Sending data from udp:[::1]:8081 to udp:[::1]:8082
-+Received message from udp:[::1]:8081: UDP Message
++Sending data from UdpV6 to udp:[::1]:8082
++Received message from udp:[::1]:56756: UDP Message
 - : unit = ()
```
haesbaert, line 212: maybe don't print the random port and keep the deterministic output ?
I think it also makes sense to keep one test using the bound socket (Eio.Net.datagram_socket `Udp...)

bikallem: ack.

haesbaert:
Code LGTM, that's in the direction I was thinking as well, it feels a bit weird that `datagram_socket` is either a `sockaddr` or a `Udp4`, but I can't think of a better way to do it without changing everything.

tests/network.md
bikallem: Done.

tests/network.md
bikallem: Done.

tests/network.md
bikallem: Done.

haesbaert (approved):
LGTM maybe @talex5 wants to have a look as well

lib_eio/net.ml
```diff
@@ -180,15 +180,19 @@ let recv (t:#datagram_socket) = t#recv
...
+
   method virtual getaddrinfo : service:string -> string -> Sockaddr.t list
   method virtual getnameinfo : Sockaddr.t -> (string * string)
 end
 
 let listen ?(reuse_addr=false) ?(reuse_port=false) ~backlog ~sw (t:#t) = t#listen ~reuse_addr ~reuse_port ~backlog ~sw
 let connect ~sw (t:#t) = t#connect ~sw
 
-let datagram_socket ~sw (t:#t) = t#datagram_socket ~sw
+let datagram_socket ~sw (t:#t) addr = t#datagram_socket addr ~sw
```
talex5, line 195: (probably will need a cast here; see below)

lib_eio/net.mli
```diff
@@ -187,8 +191,12 @@ val accept_sub :
 
 (** {2 Datagram Sockets} *)
 
-val datagram_socket : sw:Switch.t -> #t -> Sockaddr.datagram -> <datagram_socket; Flow.close>
-(** [datagram_socket ~sw t addr] creates a new datagram socket that data can be sent to
+val datagram_socket :
+     sw:Switch.t
+  -> #t
+  -> [Sockaddr.datagram | `UdpV4 | `UdpV6]
```
talex5, line 197:
```suggestion
  -> [< Sockaddr.datagram | `UdpV4 | `UdpV6]
```

lib_eio/net.mli
```diff
@@ -187,8 +191,12 @@ val accept_sub :
...
+  -> <datagram_socket; Flow.close>
+(** [datagram_socket ?addr ~sw t] creates a new datagram socket that data can be sent to
```
talex5, line 199:
```suggestion
(** [datagram_socket ~sw addr t] creates a new datagram socket that data can be sent to
```
Needs to explain what `addr` is for (source address? destination?)

lib_eio_linux/eio_linux.ml
```diff
@@ -1166,6 +1167,10 @@ let net = object
       let sock = FD.of_unix ~sw ~seekable:false ~close_unix:true sock_unix in
       Unix.bind sock_unix addr;
       udp_socket sock
+    | (`UdpV4 | `UdpV6) as s ->
+      let socket_fd = Unix.socket ~cloexec:true (socket_domain_of s) Unix.SOCK_DGRAM 0 in
+      let sock = FD.of_unix ~sw ~seekable:false ~close_unix:true socket_fd in
```
talex5, line 1172: Does is matter that the socket options are different in this case?

(Aside: why is `SO_REUSEPORT` on by default? That seems like a dangerous option, allowing two programs to listen on the same port by mistake! )

tests/network.md
```diff
@@ -165,42 +165,61 @@ Exception: Failure "Simulated error".
...
-      let e = Eio.Net.datagram_socket ~sw net e1 in
-      traceln "Sending data from %a to %a" Eio.Net.Sockaddr.pp e1 Eio.Net.Sockaddr.pp e2;
-      Eio.Net.send e e2 (Cstruct.of_string "UDP Message"))
+    (fun () -> fn ~net sw server_addr)
 ```
 
 Handling one UDP packet using IPv4:
 
 ```ocaml
-# run (run_dgram Eio.Net.Ipaddr.V4.loopback);;
```
talex5, line 190: Why have these tests had to change?

lib_eio/net.mli
haesbaert: I was asking Bikal about this on slack, why is the `<` preferred in this case ?

lib_eio_linux/eio_linux.ml
haesbaert: Good point, I wonder if it actually has any effect if you don't bind(2) ?

lib_eio/net.mli
talex5: If you have a `Sockaddr.datagram` then you should be able to pass it to this function without first having to cast it to allow the extra tags.

Sometimes the function generating the value will have a `[> ...]`, which allows it to adapt there instead, but it's good to be flexible.

lib_eio/net.ml
bikallem: Done.

lib_eio/net.mli
bikallem: Done.

lib_eio/net.mli
bikallem: Done.

lib_eio_linux/eio_linux.ml
bikallem: Not sure about `SO_REUSEPORT` myself. I think we should remove these options altogether and leave it to the user of the lib to set them if they need it after https://github.com/ocaml-multicore/eio/issues/322 is implemented. Should both of reuseport and resuseaddress settings be removed? in both TCP and UDP?

tests/network.md
bikallem: reverted the change.

lib_eio_linux/eio_linux.ml
bikallem: Let's leave this to PR which will address #322

bikallem:
@talex5 I have addressed the review comments.

lib_eio_linux/eio_linux.ml
haesbaert: I think it's fine and even desirable to have `reuse_{addr,port}` as an optional argument, but it should be default off, I'd add the optional arguments to `datagram_socket` to mimick the behaviour of `listening_socket` (default off).
This would be a breaking change as old users of `datagram_socket Sockaddr` would now get default off, but I think that's better than the old behavior.

bikallem: I have opened https://github.com/ocaml-multicore/eio/issues/362 to discuss and track this issue.

bikallem:
@haesbaert I have now added `?(reuse_addr=false)` and `?(reuse_port=false)` to `Eio.Net.datagram_socket` function.

lib_eio_linux/eio_linux.ml
bikallem: > Does is matter that the socket options are different in this case?

Yes, `Udp4` creates socket domain of `Unix.PF_INET` and `Udp6` creates socket domain of `Unix.PF_INET6`

talex5 (approved):
I pushed some more changes (to be squashed if accepted):

- I removed some duplication in the test code (and removed a duplicate test).
- I changed the socket creation code to share the bits that were the same. That revealed that the old path wasn't setting close-on-exec.
- I removed the new functions from eio_unix. This is intended for users integrating with Unix, rather than as a place for sharing common code between backends, and these functions were quite confusing (they do nothing if passed `false`), and also not shared.

If that looks OK, I'm happy for it to be merged.

lib_eio_linux/eio_linux.ml
talex5: I meant I wasn't sure if the reuse options were useful even without a fixed source address. I guess it is probably best to ignore them, as you did.

bikallem:
LGTM. :+1:

PR #361, opened 2022-11-01
Buf_read: add `try_` and `<|>` parsers

bikallem:
This PR implements `try` and `<|>` parsers in `Buf_read`. This allows to implement ABNF alternative grammar (https://www.rfc-editor.org/rfc/rfc5234.html#section-3.2) without having to left factor them first.

### DIFF
--- a/lib_eio/buf_read.ml
+++ b/lib_eio/buf_read.ml
@@ -38,6 +38,8 @@ module Syntax = struct
   let ( *> ) a b t =
     ignore (a t);
     b t
+
+  let ( <|> ) a b t = try a t with _ -> b t
 end
 
 open Syntax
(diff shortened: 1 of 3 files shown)


talex5 (changes requested):
This won't work, as Buf_read is not a back-tracking parser. Once some bytes have been consumed, it is free to reuse that part of the buffer. You can't just reset the pointers and hope.

What exactly are you trying to do? I suspect most cases can be handled with something like:
```ocaml
let parse r =
  match take_while word_char r with
  | "foo" -> `Foo
  | "bar" -> `Bar
  | x -> `Unknown x
```

(also, changes to Buf_read need corresponding updates to `fuzz_buf_read.ml` to check they work)

bikallem:
Ah, thanks for the feedback. It seem this need a bit more that I thought here. I will close this PR for now.

PR #365, opened 2022-11-09
Add setsockopt() to eio sockets

### LINKED ISSUE #322: Support for socket options

bikallem:
This PR implements setsockopt() function to `Eio.Net.socket`. 

Fixes #322

### DIFF
--- a/lib_eio/mock/net.ml
+++ b/lib_eio/mock/net.ml
@@ -71,17 +71,51 @@ let on_getaddrinfo (t:t) actions = Handler.seq t#on_getaddrinfo actions
 
 let on_getnameinfo (t:t) actions = Handler.seq t#on_getnameinfo actions
 
+type stream_socket = <
+  Flow.t;
+  Eio.Net.stream_socket; 
+  on_setsockopt: unit Handler.t;
+>
+
 type listening_socket = <
   Eio.Net.listening_socket;
-  on_accept : (Flow.t * Eio.Net.Sockaddr.stream) Handler.t;
+  on_accept : (stream_socket * Eio.Net.Sockaddr.stream) Handler.t;
+  on_setsockopt: unit Handler.t;
 >
+
+let setsockopt (type a) label handler (opt: a Eio.Net.Sockopt.t) (x : a) : unit =
+  let traceln (pp : a Fmt.t) (x: a) s = 
+    traceln "%s: setsockopt %s %a" label s pp x 
+  in
+  (match opt with
+   | IPV6_ONLY   -> traceln Fmt.bool x "IPV6_ONLY"
+   | ACCEPTCONN  -> traceln Fmt.bool x "SO_ACCEPTCONN"
+   | BROADCAST   -> traceln Fmt.bool x "SO_BROADCAST"
+   | DEBUG       -> traceln Fmt.bool x "SO_DEBUG"
+   | DONTROUTE   -> traceln Fmt.bool x "SO_DONTROUTE"
+   | KEEPALIVE   -> traceln Fmt.bool x "SO_KEEPALIVE"
+   | LINGER      -> traceln Fmt.(option int) x "SO_LINGER"
+   | OOBINLINE   -> traceln Fmt.bool x "SO_OOBINLINE"
+   | RCVBUF      -> traceln Fmt.int x "SO_RCVBUF"
+   | RCVLOWAT    -> traceln Fmt.int x "SO_RCVLOWAT"
+   | REUSEADDR   -> traceln Fmt.bool x "SO_REUSEADDR"
+   | REUSEPORT   -> traceln Fmt.bool x "SO_REUSEPORT"
+   | SNDBUF      -> traceln Fmt.int x "SO_SNDBUF"
+   | SNDLOWAT    -> traceln Fmt.int x "SO_SNDLOWAT"
+   | RCVTIMEO    -> traceln Fmt.float x "SO_RCVTIMEO"
+   | SNDTIMEO    -> traceln Fmt.float x "SO_SNDTIMEO"
+   | TYPE        -> traceln Fmt.int x "SO_TYPE"
+   | TCP_NODELAY -> traceln Fmt.bool x "TCP_NODELAY"
+  );
+  Handler.run handler
 
 let listening_socket label =
   let on_accept = Handler.make (`Raise (Failure "Mock accept handler not configured")) in
-  object
+  object (self)
     inherit Eio.Net.listening_socket
 
     method on_accept = on_accept
+    method on_setsockopt = Handler.make (`Raise (Failure "Mock setsockopt handler not configured")) 
 
     method accept ~sw =
       let socket, addr = Handler.run on_accept in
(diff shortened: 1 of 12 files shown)


lib_eio/unix/eio_unix.ml
```diff
@@ -68,3 +68,24 @@ let getnameinfo (sockaddr : Eio.Net.Sockaddr.t) =
...
+  | SO_DEBUG      -> Unix.(setsockopt fd SO_DEBUG v)
+  | SO_BROADCAST  -> Unix.(setsockopt fd SO_BROADCAST v)
+  | SO_REUSEADDR  -> Unix.(setsockopt fd SO_REUSEADDR v)
+  | SO_KEEPALIVE  -> Unix.(setsockopt fd SO_KEEPALIVE v)
+  | SO_DONTROUTE  -> Unix.(setsockopt fd SO_DONTROUTE v)
+  | SO_OOBINLINE  -> Unix.(setsockopt fd SO_OOBINLINE v)
+  | SO_ACCEPTCONN -> Unix.(setsockopt fd SO_ACCEPTCONN v)
+  | TCP_NODELAY   -> Unix.(setsockopt fd TCP_NODELAY v)
+  | IPV6_ONLY     -> Unix.(setsockopt fd IPV6_ONLY v)
+  | SO_REUSEPORT  -> Unix.(setsockopt fd SO_REUSEADDR v)
```
haesbaert, line 83:
```suggestion
  | SO_REUSEPORT  -> Unix.(setsockopt fd SO_REUSEPORT v)
```

Also maybe sort all those matches alphabetically.

lib_eio_linux/eio_linux.ml
```diff
@@ -1053,15 +1059,17 @@ let udp_socket sock = object
         raise (Failure "Expected INET UDP socket address but got Unix domain socket address.")
 end
 
-let flow fd =
+class flow fd =
```
haesbaert, line 1062: Why are you changing these to a class ? Aren't they equivalent ?

lib_eio_linux/eio_linux.ml
```diff
@@ -1101,14 +1107,22 @@ let flow fd =
       | `Send -> Unix.SHUTDOWN_SEND
       | `All -> Unix.SHUTDOWN_ALL
 
-    method unix_fd op = FD.to_unix op fd
-  end
+    method unix_fd (op: [`Peek | `Take]) = FD.to_unix op fd
```
haesbaert, line 1110: leftover ?

haesbaert (changes requested):
I can't reason much about why changing the type declaration to classes, it seems to be a change for the sake of change ? Or what am I missing. If so, it would make more sense to put those on a different commit to make reviewer easier.

Since you're providing `setsockopt`, should we provide `getsockopt` as well ?

lib_eio_linux/eio_linux.ml
bikallem: This (explicit type annotation) was needed to silence compiler/dune warnings I believe.

lib_eio_linux/eio_linux.ml
bikallem: `let flow fd = ... ` defines an object which we cannot inherit. Therefore we need a `class` which defines a common ancestor class `flow`. `stream_socket` is flow + socket functionality and we inherit flow + socket capabilities here https://github.com/ocaml-multicore/eio/blob/2460bf54c0c6b4c74d8c2c887b99623f4fe84298/lib_eio_linux/eio_linux.ml#L1113-L1118

haesbaert: Ah I see, TIL.

lib_eio/unix/eio_unix.ml
bikallem: Good catch. Done.

bikallem:
> Since you're providing setsockopt, should we provide getsockopt as well ?

Indeed, I was thinking along the same lines but wanted to get this PR reviewed and merged first. I have opened issue https://github.com/ocaml-multicore/eio/issues/376 for this.

lib_eio/unix/eio_unix.ml
bikallem: Sorted alphabetically too.

lib_eio/mock/net.ml
```diff
@@ -71,17 +71,51 @@ let on_getaddrinfo (t:t) actions = Handler.seq t#on_getaddrinfo actions
...
+  | SO_RCVLOWAT   -> traceln Fmt.int x "SO_RCVLOWAT"
+  | SO_REUSEADDR  -> traceln Fmt.bool x "SO_REUSEADDR"
+  | SO_REUSEPORT  -> traceln Fmt.bool x "SO_REUSEPORT"
+  | SO_SNDBUF     -> traceln Fmt.int x "SO_SNDBUF"
+  | SO_SNDLOWAT   -> traceln Fmt.int x "SO_SNDLOWAT"
+  | SO_RCVTIMEO   -> traceln Fmt.float x "SO_RCVTIMEO"
+  | SO_SNDTIMEO   -> traceln Fmt.float x "SO_SNDTIMEO"
+  | SO_TYPE       -> traceln Fmt.int x "SO_TYPE"
+  | TCP_NODELAY   -> traceln Fmt.bool x "TCP_NODELAY"
+  ;
```
talex5, line 109: You can't use `;` to end a match block (try ocp-indent on the code and it will show the problem).

lib_eio/net.mli
```diff
@@ -78,10 +78,35 @@ module Sockaddr : sig
   val pp : Format.formatter -> [< t] -> unit
 end
 
+(** {2 Socket Options} *)
+
+type 'a sockopt =
```
talex5, line 83: Putting it in a module means you can remove the `SO_` prefixes:
```suggestion
module Sock_opt = struct
  type 'a t =
    | IPV6_ONLY  : bool sockopt  (** Forbid binding an IPv6 socket to an IPv4 address *)
    | ACCEPTCONN : bool sockopt  (** Report whether socket listening is enabled *)
    | BROADCAST  : bool sockopt  (** Permit sending of broadcast messages *)
    ...
```
Not sure what to call the module. It doesn't matter much, though, as OCaml will be able to infer it from the type.

talex5 (commented):
It's not actually necessary to create a separate `Mock.Net.stream_socket` type (which is an API change). We could just have `Flow.t` implement the socket methods too. It already works as both a source and a sink, for example.

We could just add the new method to the implementations in eio_linux, etc, rather than using inheritance, to shrink the PR. Though this seems fine to me either way.

lib_eio/mock/net.ml
bikallem: good catch. Fixed now.

bikallem:
> It's not actually necessary to create a separate Mock.Net.stream_socket type (which is an API change). We could just have Flow.t implement the socket methods too. It already works as both a source and a sink, for example.

> We could just add the new method to the implementations in eio_linux, etc, rather than using inheritance, to shrink the PR. Though this seems fine to me either way.

Indeed, this is an api change; however, I think this api addition should be mostly backwards compatible since I am not sure if there is too much (any?) code in the wild which inherits from `Eio.Net.socket` class. 

Additionally and indeed, I agonized a bit on where to put the new function - `Eio.Flow.two_wya, Eio.Flow.source or Eio.Flow.sink`. However, this line made the decision for me (https://github.com/ocaml-multicore/eio/blob/main/lib_eio/flow.mli#L1). IIUC, flow types are a general/common abstraction over files and network sockets. As such it didn't seem coherent/cogent to me to put it in one of the flow types, i.e. how would we implement `setsockopt` on file flow type?  `Eio.Net.socket` seemed to fit the requirement as it is a super class for both udp and tcp socket and is an abstraction for sockets.

bikallem:
Ah, just realized you meant only the `mock` flow type. I don't really have a strong preference for either in the mock case. I think I mostly followed the abstraction is the `Eio.Net.socket` as I didn't want to muddy the semantics even if it is for mock use case.

bikallem:
@talex5 @haesbaert please approve/merge this PR is there are no more objections.

bikallem:
closing this for now.

Issue #138, opened 2022-01-07
Integrate Eio's CTF tracing with OCaml's tracing


talex5:
Eio's `Ctf` module provides a ring-buffer for storing trace events. Currently, this only works for a single domain. OCaml has its own CTF-based `eventlog` system (https://ocaml.org/manual/instrumented-runtime.html) and we should probably just use that somehow. It might not currently support user-defined events. There's also a proposal to replace it with `eventring` (https://github.com/ocaml-multicore/ocaml-multicore/pull/793), which says:

> This implementation is designed to be open for extension to user tracing events in the future, as well as integration with statmemprof (e.g continuous allocation profiling).

/cc @sadiqj

talex5:
@sadiqj and I discussed a couple of ways of doing this today:
- eio could watch eventring to get GC events and then copy them to its own ring, or
- eventring could allow users to add custom events and we store everything there.

The second option seems better, because:
- Having a single ring means events get recorded in the right order.
- It saves us a lot of code managing CTF buffers.
- It allows other tools to add extra events of their own.

The default eventring size is quite small (64KB) but is configurable. Eio uses 1MB by default.

TheLortex:
Custom event rings PR: https://github.com/ocaml/ocaml/pull/11474
We should make sure that it fits Eio's requirements.

patricoferris:
Had a quick go at using custom runtime events: https://github.com/patricoferris/eio/commit/671a5b421e0eb2f861204322f98fe222cab0be67#diff-b48e492abda9e349dd9f1df758814b45a11c5a63fcd89d65d09dc0bb7194125b -- looks good @TheLortex, but I didn't test it fully.

(note: I had to do some [questionable changes to your OCaml PR to get it running on macOS](https://github.com/patricoferris/ocaml/commit/a139ea7522c355750df7e672e1d0b6bf98a59435))

talex5:
PR: https://github.com/ocaml-multicore/eio/pull/554

talex5:
This was done in #635.

PR #368, opened 2022-11-10
Make sure we mint exclusive ids

haesbaert:
This preserves the same id starting as 1, as fetch_and_add returns the value pre-increment. Therefore it now expresses next_id, not last_id.

### DIFF
--- a/lib_eio/core/ctf.ml
+++ b/lib_eio/core/ctf.ml
@@ -28,11 +28,10 @@ end
 
 type id = int
 
-let last_id = ref 0
+let next_id = Atomic.make 1
 
 let mint_id () =
-  incr last_id;
-  !last_id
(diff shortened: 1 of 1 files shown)


talex5:
Note: this will only make a difference with multiple domains, and tracing doesn't work there anyway at the moment. It might also cause a large slowdown as all domains fight over the single counter.

Perhaps this will be fixed by the larger changes in https://github.com/ocaml-multicore/eio/issues/138 (which @patricoferris is investigating).

haesbaert:
Didn't know it wasn't working for multiple domains, in that way disregard this until #138 is in shape, cause it will slow things for sure.

TheLortex:
Closing as we came up with an implementation with less contention: https://github.com/ocaml-multicore/eio/pull/480

Issue #321, opened 2022-09-20
Signals abstraction


haesbaert:
As discussed in #301 we might need to abstract Signals.

I've got a working prototype that works on uring and libuv but we have to define some things before I go forward.

### 1. Do we want to export more signals, less or the same as Sys ?
Sys is a bit conservative and I think we should export more, for example it misses SIGWINCH and SIGINFO (linux doesn't have this but it's popular in Unix).

### 2. Do we want to restrict what we export ?
Like type `signum = Sigint | Sigwhatever`, or we just keep accepting an `int` and if the user wants to use whatever other signal he has, he can.
I think we should keep taking an int, but do some discovery ourselves (for the value of SIGWINCH for example) and export that as a 
`val sigwinch : int`
Worth noting that the default ocaml signal interface accepts arbitrary integers so that would be no issue, also not an issue with Luv.

### 3. What do we do with signals that are not supported ? 
Do we want to fail hard, fail silently or give it a "button" to control the behavior, this is likely very relevant for windows as they have only a small set of signals.

haesbaert:
### Progress
I have something not in a reviewable state, but I think now I understand all the nits for us to decide what to do:
https://github.com/haesbaert/eio/tree/signal (disregard the curse words, I have to make this fun somehow)

The runtime processes signals in a more native way than `libuv`, I'll try to describe how both work so we can decide what we want, the differences are particularly important for multiple domains.

### Ocaml runtime signals + Domains
The runtime doesn't do a lot, basically signals are recorded in a bitmap and processed at some point outside the trampoline.
The signal bitmap is shared between multiple domains, and code is careful enough when accessing it concurrently, so all good.
This also means we fall into standard `pthread` signals semantics, it's undefined which `pthread` (therefore which `Domain`) gets the signal. Normally each `pthread` can mask them (but `pthread_setmask` is not exported), but the delivery is 'to any `pthread/Domain` not blocking it at this time'.

This makes signals pretty hard to use with multiple domains, if we wanted to cancel some `Fiber` for example we would have to maybe cross Domains depending where the signal was delivered.

### Libuv Is a Good Boi (tm)
Libuv handles signals in a different way, they have their own signal dispatch mechanism (therefore also run signals outside the trampoline). Each `signal handle` is associated with a `loop` and each `loop` is associated with a `Domain` (in our case).
[...]

Issue #301, opened 2022-09-01
Signals don't get processed while Fibers are scheduled?


anmonteiro:
I couldn't get eio to handle my signals while another fiber is running (even if sleeping). Take this pathological case:
```ocaml
open Eio.Std

let () =
  Sys.(
    set_signal
      sigint
      (Signal_handle
         (fun i ->
           Format.eprintf "handle signal: %d@." i;
           exit 0)));
  Eio_main.run (fun _env ->
      Switch.run (fun sw ->
          Fiber.fork ~sw (fun () -> Eio.Time.sleep (Eio.Stdenv.clock _env) 5.)))
```

You can send `ctrl+C` to the program while it's running and it won't process it until after 5 seconds (when the eio fiber ends).

Tested on macOS (with `eio_luv`)

anmonteiro:
This seems to work well with eio_linux.

anmonteiro:
I'm not sure whether this can work with eio_luv at all (it looks like it takes over signal handling while its event loop is running). A workaround I've found is to install my own signal handling with Luv and run a separate loop:

```ocaml
open Eio.Std

let () =
  let loop = Luv.Loop.init () |> Result.get_ok in
  let () =
    match Luv.Signal.init ~loop () with
    | Ok handle ->
      let handler () =
        Format.eprintf "handle signal@.";
        Luv.Handle.close handle ignore
      in
      Result.get_ok (Luv.Signal.start handle Luv.Signal.sigint handler)
    | Error _ -> ()
  in

  Eio_main.run (fun _env ->
      Switch.run (fun sw ->
          Fiber.fork ~sw (fun () ->
              Eio_unix.run_in_systhread (fun () ->
                  let (_ : bool) = Luv.Loop.run ~loop () in
                  ()));
          Fiber.fork ~sw (fun () -> Eio.Time.sleep (Eio.Stdenv.clock _env) 5.)))
```

This works for now but it imposes a dependency on luv on linux.

talex5:
We probably need an abstraction for signals (an `env#signals` perhaps, or something in `Eio_unix`).

haesbaert:
I'm having a quick jab at this, one of the possible issues is that luv exports a considerably smaller set of signals, blame is on windows support:
```
(** {1:signals Signal numbers}

    For the moment, the signals exposed are those that are both present on Unix
    and present or emulated by libuv on Windows. See
[...]

PR #374, opened 2022-11-29
Signal

haesbaert:
This implements Signals in a publish/subscriber pattern, the main goal is to
give a bit better semantics than pure POSIX signals while still being able to
work in cenarios where a native signal is needed (think a debugger and whatnot).

### Rationale

POSIX semantics for multithreading are very weak, signal dipositions continue to
be process-wide while signal masking is pthread-wide, this means when a signal
arrives to the process PID, any thread can end up handling the signal, this
together with the normal semantics of signals of being processed in whatever
altstack and interrupting code make it difficult to make proper use of it in
places like EIO.

I propose a publish/subscriber way of dealing with signals, a bit inspired by
how Libuv already handles it, users can subscribe to a signal number in a signal
box, this box can be waited on signals and integrates well with EIO Fibers,
blocking and so on.

A signal box (subscription) is always linked to a Switch.t in order to guarantee
the subscription is removed when the handler goes out of scope.

Multithreading is allowed, any Domain can create subscriptions, this makes it
interesting for things like every domain installing a handler for SIGTERM and
then doing a proper finishing when needed, or maybe handling a SIGHUP and
parking everyone until the main domain reconfigures and whatnot.

### Signal Numbers

The signals from Sys are in a negative form and the runtime translates them to
the system signal number, this would be ok if Sys exported all signals we want,
but it doesnt, so we have to discover our signal numbers. Sys actually works
with its own idea of a signal number (the negative) as well as the positive (the
system), with some minor inconveniences.

Since not every system has every signal, signals that are not guaranteed to
exist are exposed as an optional type, this forces the user to acknowledge "ok
this is not portable" as well as preventing him from using an invalid number.
The actual number is hidden from the user as a private int, but can be acquired
via Signal.signum_to_int.

### Libuv/Luv Implementation

Libuv provides pretty much the publish/subscriber mechanism already, so we just
make use of that, in luv signals are not really processed in the signal handler,
they just schedule something to be processed in the main loop.

### Uring Implementation

In uring we implement the publish/subscriber mechanism ourselves, since a signal
can have multiple listeners, we have to be careful to install a new handler when
it's the first and restore the original one when it's the last one being
removed. We do not make sigprocmask dances as they become a nightmare due to
inheritance.

The signal handler in uring must wake the signalbox (an Eio.Stream.t), which
means it must go into the scheduler, which means if we just handled it from the
signal handler, we would end up recursing, which means we wouldn't be able to
grab locks, which means we wouldn't be able to do Eio.Stream.add.

We solve this by using the old pipe trick, the signal handler just writes a byte
to a pipe via Unix.write (must be outside of EIO). We then have an extra Fiber
(like the eventfd listener) running by default that reads from the other end of
the pipe, gets the signal and publishes to all listeners via Eio.Stream.add.

### Why Not signalfd

I've actually written one version with signalfd, but it's a pretty terrible API
if you need multithreading, to make usage of it you basically have to mask all
signals on every Domain (which means you have to decide which ones you will mask
beforehand, possibly killing uses of catching SIGSEGV/SIGSTOP). Doing the pipe
trick gives us what we wanted from signalfd and is also portable if we want it
in other backends. I still have the signalfd code in a branch if anyone wants
it.
  There's a half-decent rant here:
  https://ldpreload.com/blog/signalfd-is-useless

### Why Stream.addcanfail

I've added addcanfail to Stream since signals can be lost, we don't want to
block the sender, we also don't want to be forking senders and/or buffering
signals indefinately. Actual POSIX signals can only be pending, there are never
"two SIGINT" to be processed, two SIGINT would be coalesced into one. We do a
bit more, when we introduce the pipe trick we end up buffering signals, which
means two SIGINTS would be two SIGINTS, so we're bound by the pipe buffer size.
Since draining the pipe is not dependant on the user calling anything, it's
virtually impossible to saturate it.

We still need addcanfail since the pipe reader cannot block on publish, imagine
you have multiple signals but the user never reads from one of the boxes, if the
pipe reader would block trying to send, it would not publish anything else.

Worth noting that libuv as well as libevent do the exact same, as it's an old
solution.

Pokes at #321

### DIFF
--- a/lib_eio/dune
+++ b/lib_eio/dune
@@ -3,3 +3,7 @@
   (public_name eio)
   (flags (:standard -open Eio__core -open Eio__core.Private))
   (libraries eio__core cstruct lwt-dllist fmt bigstringaf optint mtime))
+
+(rule
+ (targets config.ml)
+ (action (run ./include/discover.exe)))
--- a/lib_eio/eio.ml
+++ b/lib_eio/eio.ml
@@ -28,6 +28,7 @@ module Time = Time
 module File = File
 module Fs = Fs
 module Path = Path
+module Signal = Signal
 
 module Stdenv = struct
   type t = <
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -120,6 +120,9 @@ module Fs = Fs
 (** Accessing paths on a file-system. *)
 module Path = Path
 
+(** Setting up signal handlers. *)
+module Signal = Signal
+
 (** Control over debugging. *)
 module Debug : sig
   (** Example:
--- a/lib_eio/include/discover.ml
+++ b/lib_eio/include/discover.ml
@@ -1,0 +1,102 @@
+module C = Configurator.V1
+
+let mangle (s:string) =
+  let is_alpha = function 'a' .. 'z' | 'A' .. 'Z' -> true | _ -> false in
+  let rec first_alpha i =
+    if is_alpha (String.get s i) then
+      i
+    else
+      first_alpha (succ i)
+  in
+  let skip = first_alpha 0 in
+  String.sub s skip ((String.length s) - skip) |>
+  String.lowercase_ascii
+
+let () =
+  let open C.C_define in
+  C.main ~name:"discover" @@
+  fun c ->
+  let sig_build =
+    C.C_define.import c
+      ~c_flags:["-I"; Filename.concat (Sys.getcwd ()) "include"]
+      ~includes:["signal.h"]
+  in
+  let process =
+    List.map
+      (function
+        | name, Value.Int v ->
+          Printf.sprintf "  let %s = %d" (mangle name) v
+        | name, Value.Switch v ->
+          if not v then
+            Printf.sprintf "  let %s_opt = None" (mangle name)
+          else
+            (match sig_build Type.[ name, Int; ] with
+             | (name, Value.Int v) :: [] ->
+               Printf.sprintf "  let %s_opt = Some %d" (mangle name) v
+             | _ -> assert false)
+        | _ -> assert false)
+  in
+  let signums =
+    sig_build
+      Type.[
+        "SIGHUP", Int;
+        "SIGINT", Int;
+        "SIGQUIT", Int;
+        "SIGILL", Int;
+        "SIGTRAP", Int;
+        "SIGABRT", Int;
+        "SIGEMT", Switch;
+        "SIGFPE", Int;
+        "SIGKILL", Int;
+        "SIGBUS", Int;
+        "SIGSEGV", Int;
+        "SIGSYS", Int;
+        "SIGPIPE", Int;
+        "SIGALRM", Int;
+        "SIGTERM", Int;
+        "SIGURG", Int;
+        "SIGSTOP", Int;
+        "SIGTSTP", Int;
+        "SIGCONT", Int;
+        "SIGCHLD", Int;
+        "SIGTTIN", Int;
+        "SIGTTOU", Int;
+        "SIGIO", Int;
+        "SIGXCPU", Int;
+        "SIGXFSZ", Int;
+        "SIGVTALRM", Int;
+        "SIGPROF", Int;
+        "SIGWINCH", Int;
+        "SIGINFO", Switch;
+        "SIGUSR1", Int;
+        "SIGUSR2", Int;
+        "SIGTHR", Switch;
+      ] |> process
+  in
+  let nsig =
+    let maxfound =
+      sig_build
+        Type.[
+          "NSIG", Switch;
+          "_NSIG", Switch;
+          "_SIG_MAXSIG", Switch;
+        ] |>
+      List.map
+        (function
+          | name, Value.Switch v ->
+            if not v then
+              0
+            else
+              (match sig_build Type.[ name, Int; ] with
+               | (_, Value.Int v) :: [] -> v
+               | _ -> assert false)
+          | _ -> assert false)
+    in
+    let maxfound = List.fold_left max 64 maxfound in
+    [ Printf.sprintf "  let nsig = %d" maxfound ]
+  in
+  C.Flags.write_lines "config.ml" @@
+  [ "module Signum = struct" ] @
+  nsig @
+  signums @
+  [ "end" ]
--- a/lib_eio/include/dune
+++ b/lib_eio/include/dune
@@ -1,0 +1,4 @@
+(executable
+ (name discover)
+ (modules discover)
+ (libraries dune-configurator))
--- a/lib_eio/signal.ml
+++ b/lib_eio/signal.ml
@@ -1,0 +1,70 @@
+(*
+ * Copyright (C) 2022 Christiano Haesbaert
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *)
+
+include Config.Signum
+
+type sigbox = {
+  box        : unit Stream.t;
+  signum     : int;
+  hook       : Switch.hook ref;
+  subscribed : bool ref;
+}
+
+type signum = int
+
+let signum_to_int signum = (signum :> int)
+
+module Private = struct
+  type _ Effect.t +=
+    | Subscribe : (int * unit Stream.t) -> unit Effect.t
+    | Unsubscribe : (int * unit Stream.t) -> unit Effect.t
+    | Publish : int -> unit Effect.t
+
+  let subscribe sigbox = Effect.perform @@ Subscribe (sigbox.signum, sigbox.box)
+  let unsubscribe sigbox = Effect.perform @@ Unsubscribe (sigbox.signum, sigbox.box)
+  let publish signum = Effect.perform @@ Publish signum
+end
+
+let check_subscription sigbox =
+  if not !(sigbox.subscribed) then
+    invalid_arg "sigbox is not subscribed"
+
+let unsubscribe sigbox =
+  check_subscription sigbox;
+  Private.unsubscribe sigbox;
+  sigbox.subscribed := false;
+  Switch.remove_hook !(sigbox.hook)
+
+let subscribe ~sw signum =
+  assert (signum > 0 && signum < nsig);
+  let sigbox =
+    { box = Stream.create 1; signum;
+      hook = ref Switch.null_hook; subscribed = ref false }
+  in
+  sigbox.hook := Switch.on_release_cancellable sw (fun () -> unsubscribe sigbox);
+  Private.subscribe sigbox;
+  sigbox.subscribed := true;
+  sigbox
+
+let publish signum = Private.publish signum
+
+let wait sigbox =
+  check_subscription sigbox;
+  Stream.take sigbox.box
+
+let wait_one signum = Switch.run @@ fun sw -> subscribe ~sw signum |> wait
+
+let is_pending sigbox = not (Stream.is_empty sigbox.box)
--- a/lib_eio/signal.mli
+++ b/lib_eio/signal.mli
@@ -1,0 +1,107 @@
+(** Signals
+
+    Signals in EIO work on a subscription basis. A {!sigbox} is
+    subscribed to a specific {!signum} via [subscribe], which can in
+    turn be waited on with [wait].
+
+    When a signal arrives, all {!sigbox}es subscribed to the
+    corresponding signal are populated and [wait] returns.
+
+    Signals are buffered but not indefinately, the size of the buffer
+    and the number of signals queued depend on the backend implementation.
+
+    There can be multiple subscriptions for the same {!signum}, even
+    across different domains, but each individual {!sigbox} is local
+    to its domain and tied to a {!Switch.t}
+*)
+
+(** Example:
+    {[
+      Switch.run @@ fun sw ->
+      let sighupbox = Signal.(subscribe ~sw sighup) in
+      while !run do
+        Signal.wait sighupbox;
+        traceln "Got sighup";
+        daemon_reconfigure ();
+      done;
+      Signal.unsubscribe sighupbox
+    ]}
+
+    {[
+      Fiber.first
+        (fun () -> Signal.(wait_one sigint))
+        (fun () -> Time.sleep clock 1.0)
+    ]}
+*)
+
+type sigbox
+(** Stores signals that can be [wait]ed on, each {!sigbox} listens to
+    one {!signum}, you can have multiple {!sigbox} for the same
+    {!signum}, and in multiple domains.
+
+*)
+
+module Private : sig
+  type _ Effect.t +=
+    | Subscribe : (int * unit Stream.t) -> unit Effect.t
+    | Unsubscribe : (int * unit Stream.t) -> unit Effect.t
+    | Publish : int -> unit Effect.t
+end
+
+type signum = private int
+
+val subscribe : sw:Switch.t -> signum -> sigbox
+(** Create a subscription for waiting on signal {!signum}. *)
+
+val unsubscribe : sigbox -> unit
+(** Remove a subscription, possibly restoring the orignal {!signum} behaviour. *)
+
+val publish : signum -> unit
+(** Raise a signal, like kill. **)
+
+val wait : sigbox -> unit
+(** Wait for signals, it is an error to [wait] on an unsubscribed {!sigbox}. *)
+
+val wait_one : signum -> unit
+(** Like [wait] but only for a single occurrence of {!signum}. *)
+
+val is_pending : sigbox -> bool
+(** false if [wait] would block. *)
+
+val signum_to_int : signum -> int
+(** The system's idea of {!signum}, always a positive number, unlike signals from {!Sys}. *)
+
+(** Signals, some are optional since not every system supports those. *)
+val sighup : signum
+val sigint : signum
+val sigquit : signum
+val sigill : signum
+val sigtrap : signum
+val sigabrt : signum
+val sigemt_opt : signum option
+val sigfpe : signum
+val sigkill : signum
+val sigbus : signum
+val sigsegv : signum
+val sigsys : signum
+val sigpipe : signum
+val sigalrm : signum
+val sigterm : signum
+val sigurg : signum
+val sigstop : signum
+val sigtstp : signum
+val sigcont : signum
+val sigchld : signum
+val sigttin : signum
+val sigttou : signum
+val sigio : signum
+val sigxcpu : signum
+val sigxfsz : signum
+val sigvtalrm : signum
+val sigprof : signum
+val sigwinch : signum
+val siginfo_opt : signum option
+val sigusr1 : signum
+val sigusr2 : signum
+val sigthr_opt : signum option
+val nsig : int
--- a/lib_eio/stream.ml
+++ b/lib_eio/stream.ml
@@ -64,6 +64,22 @@ let add t item =
         )
     )
 
+let add_nonblocking t item =
+  Mutex.lock t.mutex;
+  match Waiters.wake_one t.readers item with
+  | `Ok -> Mutex.unlock t.mutex; true
+  | `Queue_empty ->
+    (* No-one is waiting for an item. Queue it. *)
+    if Queue.length t.items < t.capacity then (
+      Queue.add item t.items;
+      Mutex.unlock t.mutex;
+      true
+    ) else (
+      Mutex.unlock t.mutex;
+      false)
+
+let add_canfail t item = ignore @@ add_nonblocking t item
+
 let take t =
   Mutex.lock t.mutex;
   match Queue.take_opt t.items with
--- a/lib_eio/stream.mli
+++ b/lib_eio/stream.mli
@@ -28,6 +28,16 @@ val add : 'a t -> 'a -> unit
 
     If this would take [t] over capacity, it blocks until there is space. *)
 
+val add_nonblocking : 'a t -> 'a -> bool
+(** [add_nonblocking t item] adds [item] to [t].
+
+    Returns true if added, false if it would block. *)
+
+val add_canfail : 'a t -> 'a -> unit
+(** [add_canfail t item] tries to add [item] to [t].
+
+    Adding can fail if [t] is full, only use this when you want unreliable streams. *)
+
 val take : 'a t -> 'a
 (** [take t] takes the next item from the head of [t].
 
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -33,6 +33,8 @@ type amount = Exactly of int | Upto of int
 
 let system_thread = Ctf.mint_id ()
 
+let signal_pipe_key = Domain.DLS.new_key @@ fun () -> None
+
 let wrap_errors path fn =
   try fn () with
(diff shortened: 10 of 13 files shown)


haesbaert:
Also cc #301

talex5:
> Why Not signalfd

Well, that's annoying! The `signalfd` man-page even admits this:

> Be aware, however, that this won't be  possible  in  the  case of a helper program spawned behind the scenes by any library function that the program may call.  In such cases, one must fall back to using a traditional signal handler that writes  to a file descriptor monitored by select(2), poll(2), or epoll(7).

This PR seems a bit ad-hoc, though. Perhaps we can start with something simpler. The basic building block we need is the ability to schedule something to run in the main event loop. e.g. for `Eio_linux` we should provide a way to get this `enqueue` function (e.g. via `env` or by performing an effect):

```ocaml
let enqueue fn =
  Lf_queue.push st.run_q (Fn fn);
  if Atomic.get st.need_wakeup then wakeup st
```
where `Fn` is handled in `schedule` with:
```ocaml
  | Some Fn fn -> fn (); schedule st
```

Then you could handle signals like this:
```ocaml
open Eio.Std

let () =
  Eio_main.run @@ fun env ->
  let enqueue = (* get it somehow *) in
  let handler (_ : int) =
    (* (running in signal handler) *)
    enqueue @@ fun () ->
    (* (called from event loop) *)
    traceln "Interrupt!"
  in
  Sys.(set_signal sigint) (Sys.Signal_handle handler);
  traceln "Sleeping for 5 seconds...";
  Eio.Time.Mono.sleep env#mono_clock 5.
```
That would give people the flexibility to try different schemes. e.g. they could set a flag and signal a condition if it was previously clear.

Possibly we should store `enqueue` in domain-local storage, so it's always available. That would also be useful for GC finalizers.

(note: `wakeup` is not quite signal safe currently, as it takes a mutex and might deadlock if sent from the same domain, but the pipes thing has the same problem where a write to the pipe races with closing the pipe)

haesbaert:
> > Why Not signalfd
> 
> Well, that's annoying! The `signalfd` man-page even admits this:
> 
> > Be aware, however, that this won't be  possible  in  the  case of a helper program spawned behind the scenes by any library function that the program may call.  In such cases, one must fall back to using a traditional signal handler that writes  to a file descriptor monitored by select(2), poll(2), or epoll(7).
> 
> This PR seems a bit ad-hoc, though. Perhaps we can start with something simpler. The basic building block we need is the ability to schedule something to run in the main event loop. e.g. for `Eio_linux` we should provide a way to get this `enqueue` function (e.g. via `env` or by performing an effect):
> 

What do you mean by ad-hoc specifically ?

> ```ocaml
> let enqueue fn =
>   Lf_queue.push st.run_q (Fn fn);
>   if Atomic.get st.need_wakeup then wakeup st
> ```
> 
> where `Fn` is handled in `schedule` with:
> 
> ```ocaml
>   | Some Fn fn -> fn (); schedule st
> ```

This could also be done with a Stream and it was one of the ideas, but it's a bit tricky, you either have to fork a fiber indefinately so to allow the handler to block freely, or you would constrain the handler to not being able to block, which is kinda bad. I like the idea of a generic "please domain X run Y for me" but the semantics need to be defined and it's not something users should have to learn.

Most of my design is to be able to free the user from any constraint and also to make sure we're dealing with being bound and not forking things indefinately. I don't like the idea of the user having to learn different contexts, most of EIO is waiting for things, so why not channel signals through the same way he is already used to ?

> 
> Then you could handle signals like this:
> 
> ```ocaml
> open Eio.Std
> 
> let () =
>   Eio_main.run @@ fun env ->
>   let enqueue = (* get it somehow *) in
>   let handler (_ : int) =
>     (* (running in signal handler) *)
>     enqueue @@ fun () ->
>     (* (called from event loop) *)
>     traceln "Interrupt!"
>   in
>   Sys.(set_signal sigint) (Sys.Signal_handle handler);
>   traceln "Sleeping for 5 seconds...";
>   Eio.Time.Mono.sleep env#mono_clock 5.
> ```
> 
> That would give people the flexibility to try different schemes. e.g. they could set a flag and signal a condition if it was previously clear.

Not sure I follow this, but I assume you don't mean this:

```
let rec handler =
  Signal.wait sigint_box;
  set_flag foo 
```

> 
> Possibly we should store `enqueue` in domain-local storage, so it's always available. That would also be useful for GC finalizers.
> 
> (note: `wakeup` is not quite signal safe currently, as it takes a mutex and might deadlock if sent from the same domain, but the pipes thing has the same problem where a write to the pipe races with closing the pipe)

That's not entirely true as the pipe is only closed after all fibers have finished, basically when domain exits, also, the closing of the pipe would just trigger an error on the write, which would not lead to any damage.

The recursive mutex is a real issue (wakeup recursing into mutex) and I don't see how we could start with that.

I fail to see what's wrong with the pipe, but the proper solution if that is not wanted is an Atomic.t that is modified by the signal handler and pokes the mainloop via eventfd, and we would check that Atomic.t at every loop iteration, I prefer the pipe though.

haesbaert:
Not sure if I was clear but you can do things like
```
(* wait for 5 seconds or a sigint *)
Fiber.first (fun () -> Eio.Time.Mono.sleep env#mono_clock 5)
            (fun () -> Signal.wait_one sigint))
```

talex5:
> the closing of the pipe would just trigger an error on the write, which would not lead to any damage.

It can, though, because another domain could open a different file reusing the FD. Then writing to the pipe will corrupt that file. This is why `wakeup` takes a mutex.

> The recursive mutex is a real issue (wakeup recursing into mutex) and I don't see how we could start with that.

There are other ways of solving this problem. I think the easiest is to keep a free list of eventfds instead of closing them. Then you might just get a spurious wakeup, which doesn't matter.

> What do you mean by ad-hoc specifically ?

I mean it's doing a load complicated stuff that seems only useful for some specific use. Ideally, I think it would look like this:
```ocaml
open Eio.Std

let () =
  Eio_main.run @@ fun _ ->
  let interrupted = Eio.Condition.create () in
  Sys.(set_signal sigint) (Sys.Signal_handle (fun _ -> Eio.Condition.broadcast interrupted));
  Fiber.first
    (fun () -> Eio.Condition.await_no_mutex interrupted; traceln "Interrupted!")
    (fun () ->
       traceln "Running job (Ctrl-C to cancel)";
       Fiber.await_cancel ()
    );
  traceln "Done"
```

Note that the above doesn't require any new APIs at all. It *almost* works already; the only problem is that `Eio.Condition.broadcast` isn't signal-safe because it takes a mutex (so in theory it could deadlock). But we want to make that lock-free anyway.

haesbaert:
> > the closing of the pipe would just trigger an error on the write, which would not lead to any damage.
> 
> It can, though, because another domain could open a different file reusing the FD. Then writing to the pipe will corrupt that file. This is why `wakeup` takes a mutex.

Ack, true I was mixing up with channels where there is a flag for closed, but my point stands: see below.

> 
> > The recursive mutex is a real issue (wakeup recursing into mutex) and I don't see how we could start with that.
> 
> There are other ways of solving this problem. I think the easiest is to keep a free list of eventfds instead of closing them. Then you might just get a spurious wakeup, which doesn't matter.

It's actually simpler than that, the pipe is only open/read/written/closed from within the same domain, so just setting a flag that the fd is closed is enough, then if you happen to get a signal during domain exit, it just become a nop.

> 
> > What do you mean by ad-hoc specifically ?
> 
> I mean it's doing a load complicated stuff that seems only useful for some specific use. Ideally, I think it would look like this:

I don't see it as the specific use tbh, I see is as the very general way of handling signals, it escapes exposing anything that is not safe to the user while profiting from the same patterns he would use to wait for IO.

> 
> ```ocaml
> open Eio.Std
> 
> let () =
>   Eio_main.run @@ fun _ ->
>   let interrupted = Eio.Condition.create () in
>   Sys.(set_signal sigint) (Sys.Signal_handle (fun _ -> Eio.Condition.broadcast interrupted));
>   Fiber.first
>     (fun () -> Eio.Condition.await_no_mutex interrupted; traceln "Interrupted!")
>     (fun () ->
>        traceln "Running job (Ctrl-C to cancel)";
>        Fiber.await_cancel ()
>     );
>   traceln "Done"
> ```
> 
> Note that the above doesn't require any new APIs at all. It _almost_ works already; the only problem is that `Eio.Condition.broadcast` isn't signal-safe because it takes a mutex (so in theory it could deadlock). But we want to make that lock-free anyway.

Not only that, Condition is a bit more complex because you'd either have to have one condition per listener, or keep flipping the state condition state all the time, would each listener flip its own condition to wait for the next signal ? would we accumulate ? (I need to read Condition better but so far that's what I see). IIRC our Condition implementation is more suitable for "I'm gonna wait for this to happen and I'm done".

But do you want to expose all that to the user, or that would be hidden ? There is virtually nothing he can do Eio wise that is safe from an actual signal handler, I see it as a step backwards to expose something he will shoot himself in the foot.

haesbaert:
> > Condition is a bit more complex because you'd either have to have one condition per listener, or keep flipping the state condition state all the time, would each listener flip its own condition to wait for the next signal ? would we accumulate ?
> 
> I'm not sure what you mean. Any number of fibers can wait on a condition. e.g. with two fibers:
> 
> ```
>   Fiber.both
>     (fun () ->
>        Fiber.first
>          (fun () -> Eio.Condition.await_no_mutex interrupted; traceln "Interrupted 1!")
>          (fun () ->
>             traceln "Running job 1 (Ctrl-C to cancel)";
>             Fiber.await_cancel ()
>          )
>     )
>     (fun () ->
>        Fiber.first
>          (fun () -> Eio.Condition.await_no_mutex interrupted; traceln "Interrupted 2!")
>          (fun () ->
>             traceln "Running job 2 (Ctrl-C to cancel)";
>             Fiber.await_cancel ()
>          )
>     );
> ```
> 
> I get:
> 
> ```
> +Running job 1 (Ctrl-C to cancel)
> +Running job 2 (Ctrl-C to cancel)
> ^C+Interrupted 1!
> +Interrupted 2!
> +Done
> ```
> 
> Also, it doesn't matter if the domain where the signal was registered exits or is busy when the signal is delivered.
> 
> (the condition we're waiting for is "more interrupts have been received than when we started waiting", but we don't need to keep a count or check it, since the condition will always be true when the await returns)

I was referring to keeping the "interrupt signaled state somewhere", since you don't want to lose signals while there are no listeners. 

In your example (IIUC), if the broadcast happens before the waiter is on `await_no_mutex`, the signal is lost. The second issue is that it doesn't buffer subsequent signals, usually you can be processing SIGFOO while there is a subsequent SIGFOO pending. In other words, signals should be buffered (at least one).

talex5:
> I was referring to keeping the "interrupt signaled state somewhere", since you don't want to lose signals while there are no listeners.

Well, I do want that. Ctrl-C should cancel the jobs that are running when I press it, not any future jobs I start later. If I press Ctrl-C three times trying to kill one job, it shouldn't then kill the next two as well.

Of course, if I wanted that then I could use a stream instead.

> The second issue is that it doesn't buffer subsequent signals, usually you can be processing SIGFOO while there is a subsequent SIGFOO pending. In other words, signals should be buffered (at least one).

There might be cases where you want that, but it's not wanted in this job example. Pressing Ctrl-C a second time while one job is cancelling will cancel any new jobs started since then, but not jobs started after the second Ctrl-C.

The point is, the user needs to be able to choose the handling schema that makes sense for their use.

talex5:
BTW, here's the config-reload example using Eio.Condition (^C to reload, for easier testing):
```ocaml
open Eio.Std

let () =
  Eio_main.run @@ fun _ ->
  let interrupted = Eio.Condition.create () in
  Sys.(set_signal sigint) (Sys.Signal_handle (fun _ -> Eio.Condition.broadcast interrupted));
  while true do
    Fiber.both
      (fun () -> Eio.Condition.await_no_mutex interrupted)
      (fun () ->
         traceln "Reading configuration...";
         Eio_unix.sleep 2.0;
         traceln "Finished reading configuration";
      )
  done
```
(we reload the configuration when we've both finished the previous load and received a signal since that started)
```
+Reading configuration...
+Finished reading configuration

^C+Reading configuration...
+Finished reading configuration

^C+Reading configuration...
^C+Finished reading configuration
+Reading configuration...
+Finished reading configuration

^C+Reading configuration...
^C^C^C^C^C^C^C^C^C+Finished reading configuration
+Reading configuration...
+Finished reading configuration
```
(line-breaks added for clarity)

talex5:
> There's virtually nothing the user can do in a signal handler except somehow deferring whatever he wants to do to a sane fiber context, that's the whole point of my design, to cut the middle ground.

Sometimes though there are other things you want to do, such as updating an atomic counter, or dumping debug info (which might not be completely safe, but still good enough for debugging why your signal handler isn't waking the main loop). But whether we provide a high-level API or not, we can still implement it this way without needing special support in the backends.

> The user also now needs to be carefull to always keep a fiber awaiting it, and even so there is a window where he misses the next signal

I don't think there is. `Fiber.both` runs the first fiber (that subscribes to the signal) first, and only runs the second branch once that's safely suspended. If a signal arrives before the `Eio.Condition.await_no_mutex` then loading the configuration is certain to happen in the future anyway.

You can certainly argue that's a bit subtle and should be wrapped in something. However, if so, we should be providing an API that wraps Eio.Condition (not just signals). There are other places where we need to recalculate something when a condition changes. For example, OCurrent needs to recalculate when it receives a web-hook, including if it gets another web-hook event while recalculating.

> It's pretty hard for the user to get it wrong, since there are no hidden/obsucre semantics he must understand beforehand.

I disagree with that. My version uses only OCaml signals and Eio.Condition, which are both things you might know anyway. Sigbox is a new abstraction, used only for signals. And I think it is quite surprising:

1. Any domain can register for signals, but processing them requires action from whichever domain registered first.
2. If that domain is slow, handling the signal will be slow for all domains.
3. If that domain exits, further signals will be lost (I think?).

haesbaert:
> > There's virtually nothing the user can do in a signal handler except somehow deferring whatever he wants to do to a sane fiber context, that's the whole point of my design, to cut the middle ground.
> 
> Sometimes though there are other things you want to do, such as updating an atomic counter, or dumping debug info (which might not be completely safe, but still good enough for debugging why your signal handler isn't waking the main loop). But whether we provide a high-level API or not, we can still implement it this way without needing special support in the backends.
> 
> > The user also now needs to be carefull to always keep a fiber awaiting it, and even so there is a window where he misses the next signal
> 
> I don't think there is. `Fiber.both` runs the first fiber (that subscribes to the signal) first, and only runs the second branch once that's safely suspended. If a signal arrives before the `Eio.Condition.await_no_mutex` then loading the configuration is certain to happen in the future anyway.
> 
> You can certainly argue that's a bit subtle and should be wrapped in something. However, if so, we should be providing an API that wraps Eio.Condition (not just signals). There are other places where we need to recalculate something when a condition changes. For example, OCurrent needs to recalculate when it receives a web-hook, including if it gets another web-hook event while recalculating.
> 

I think there is, remember, the broadcast happens from any domain, and since there is no state being kept, lost wakeups can occur at anytime the signal is broadcasted before or after someone `wait_no_mutex` for them. This happens in two places in your example, right after the handler is installed and between each loop interaction (I marked as three just for visualization).

```ocaml
let () =
  Eio_main.run @@ fun _ ->
  let interrupted = Eio.Condition.create () in
  Sys.(set_signal sigint) (Sys.Signal_handle (fun _ -> Eio.Condition.broadcast interrupted));
-----> signal broadcasted here is lost since there is no one waiting
  while true do
-----> signal broadcasted here is lost since there is no one waiting
    Fiber.both
      (fun () -> Eio.Condition.await_no_mutex interrupted)
      (fun () ->
         traceln "Reading configuration...";
         Eio_unix.sleep 2.0;
         traceln "Finished reading configuration";
      )
-----> signal broadcasted here is lost since there is no waiting
  done
  ```

The wait/broadcast dance needs to keep state, and not just get a signal _when_ it's waiting.

> > It's pretty hard for the user to get it wrong, since there are no hidden/obsucre semantics he must understand beforehand.
> 
> I disagree with that. My version uses only OCaml signals and Eio.Condition, which are both things you might know anyway. Sigbox is a new abstraction, used only for signals. And I think it is quite surprising:
> 
> 1. Any domain can register for signals, but processing them requires action from whichever domain registered first.

This is incorrect. Each Fiber can register for signals with a sigbox, this Fiber can be in whatever Domain. The first sigbox that is created installs the hard handler, which doesn't require any action from the the user code.

> 2. If that domain is slow, handling the signal will be slow for all domains.

This is incorrect, each sigbox is independantly buffered, the hard handler never blocks on write, or waits for anyone, if the user creates a sigbox and never `waits` for it, nothing happens with further signals, of any kind, on any other box.
Writing to the pipe _and_ draining the pipe is never dependant on user calling `wait` (that's why the initial pipe draining fiber is there, without user interaction)

> 3. If that domain exits, further signals will be lost (I think?).

This is incorrect. If the switch containing the sigbox ends, that sigbox is destroyed, it has no impact on other sigbox or further signals of that type. If this was the last sigbox, the hard handler is unregistered.

**Comment by talex5** [comment]
> This happens in two places in your example, right after the handler is installed and between each loop interaction (I marked as three just for visualization).

It doesn't matter if the signal is lost in any of the places you marked. If we're at any of those points then we haven't yet started reading the configuration, so we will read the configuration after the signal was sent, as required.

```ocaml
let () =
  Eio_main.run @@ fun _ ->
  let interrupted = Eio.Condition.create () in
  Sys.(set_signal sigint) (Sys.Signal_handle (fun _ -> Eio.Condition.broadcast interrupted));
-----> signal broadcasted here is lost since there is no one waiting
-----> doesn't matter; we're about to enter the loop and read the configuration anyway
  while true do
-----> signal broadcasted here is lost since there is no one waiting
-----> doesn't matter; we're about to start reading anyway
    Fiber.both
      (fun () -> Eio.Condition.await_no_mutex interrupted)
      (fun () ->
         traceln "Reading configuration...";
         Eio_unix.sleep 2.0;
         traceln "Finished reading configuration";
      )
-----> signal broadcasted here is lost since there is no one waiting
-----> doesn't matter; we're about to loop and read the config again anyway
  done
```

>> Any domain can register for signals, but processing them requires action from whichever domain registered first.
>
> This is incorrect. Each Fiber can register for signals with a sigbox, this Fiber can be in whatever Domain. The first sigbox that is created installs the hard handler, which doesn't require any action from the the user code.

I'm looking at this code in this PR:
```ocaml
 let subscribe signum box =
    match Domain.DLS.get signal_pipe_key with
    | None -> ()
    | Some pipe ->
      with_signal (fun () ->
          let ol = Atomic.get sigmap.(signum) in
          let first = ol = [] in
          Atomic.set sigmap.(signum) (box :: ol);
          if first then
            Sys.set_signal signum
              (Sys.Signal_handle
                 (fun _sysnum ->
                    let b = Bytes.create 1 in
                    Bytes.set_uint8  b 0 signum;
                    let n = Unix.single_write (snd pipe) b 0 1 in
                    assert (n = 1))))
```
Let's say:
1. Domain 1 registers for signal 1.
2. This gets the pipe to domain 1's event loop and installs a handler to write to it.
3. Domain 2 registers for signal 1. This isn't `first`, so it just adds it to the list.
4. Domain 1 is now busy doing some CPU stuff. Signal 1 occurs. The handler runs and sends a wake-up to domain 1's event loop. This won't be processed until domain 1 is ready, even though domain 2 is ready now.

haesbaert:
> 



> > This happens in two places in your example, right after the handler is installed and between each loop interaction (I marked as three just for visualization).
> 
> It doesn't matter if the signal is lost in any of the places you marked. If we're at any of those points then we haven't yet started reading the configuration, so we will read the configuration after the signal was sent, as required.
> 

I have to disagree here, it does matter, the configuration file changed _again_ and the signal happens while you were in the blank spots, you miss the event, that's the whole point of signals being leveled and not edge triggered. Your code makes them all edge triggered.

> ```ocaml
> let () =
>   Eio_main.run @@ fun _ ->
>   let interrupted = Eio.Condition.create () in
>   Sys.(set_signal sigint) (Sys.Signal_handle (fun _ -> Eio.Condition.broadcast interrupted));
> -----> signal broadcasted here is lost since there is no one waiting
> -----> doesn't matter; we're about to enter the loop and read the configuration anyway
>   while true do
> -----> signal broadcasted here is lost since there is no one waiting
> -----> doesn't matter; we're about to start reading anyway
>     Fiber.both
>       (fun () -> Eio.Condition.await_no_mutex interrupted)
>       (fun () ->
>          traceln "Reading configuration...";
>          Eio_unix.sleep 2.0;
>          traceln "Finished reading configuration";
>       )
> -----> signal broadcasted here is lost since there is no one waiting
> -----> doesn't matter; we're about to loop and read the config again anyway
>   done
> ```
> 
> > > Any domain can register for signals, but processing them requires action from whichever domain registered first.
> > 
> > 
> > This is incorrect. Each Fiber can register for signals with a sigbox, this Fiber can be in whatever Domain. The first sigbox that is created installs the hard handler, which doesn't require any action from the the user code.
> 
> I'm looking at this code in this PR:
> 
> ```ocaml
>  let subscribe signum box =
>     match Domain.DLS.get signal_pipe_key with
>     | None -> ()
>     | Some pipe ->
>       with_signal (fun () ->
>           let ol = Atomic.get sigmap.(signum) in
>           let first = ol = [] in
>           Atomic.set sigmap.(signum) (box :: ol);
>           if first then
>             Sys.set_signal signum
>               (Sys.Signal_handle
>                  (fun _sysnum ->
>                     let b = Bytes.create 1 in
>                     Bytes.set_uint8  b 0 signum;
>                     let n = Unix.single_write (snd pipe) b 0 1 in
>                     assert (n = 1))))
> ```
> 
> Let's say:
> 
> 1. Domain 1 registers for signal 1.
> 2. This gets the pipe to domain 1's event loop and installs a handler to write to it.
> 3. Domain 2 registers for signal 1. This isn't `first`, so it just adds it to the list.
> 4. Domain 1 is now busy doing some CPU stuff. Signal 1 occurs. The handler runs and sends a wake-up to domain 1's event loop. This won't be processed until domain 1 is ready, even though domain 2 is ready now.

Yes, if you mean that signal can be delayed until the receiving Domain blocks on a Fiber, yes it is delayed, if this is a problem, then the whole assumption of a concurrent scheduler is broken and we should delete it all and make it all preemptive :) throught SIGALARM tricks.

talex5:
> I have to disagree here, it does matter, the configuration file changed again and the signal happens while you were in the blank spots, you miss the event, that's the whole point of signals being leveled and not edge triggered.

Perhaps you could give an example sequence of actions where it doesn't work? If the user changes the config file and then we reach one of the spots you indicated, we will certainly read the updated configuration (in fact, we'll do that even if the user forgets to send a signal at all, in those specific cases).

The only difficult case is where a signal occurs after we start reading the configuration and before we start watching for the signal. But that can't happen because we have already started watching for the signal before starting to read the config.

>> Domain 1 is now busy doing some CPU stuff. Signal 1 occurs. The handler runs and sends a wake-up to domain 1's event loop. This won't be processed until domain 1 is ready, even though domain 2 is ready now.
>
> Yes, if you mean that signal can be delayed until the receiving Domain blocks on a Fiber, yes it is delayed, if this is a problem, then the whole assumption of a concurrent scheduler is broken and we should delete it all and make it all preemptive through SIGALARM tricks :).

I would expect this to work (and it will, with Eio.Condition and lock-free waiters). In the example I gave:

1. Domain 2 registered for a signal.
2. The signal happened.
3. Domain 2 is ready to process it, but it doesn't get delivered because an unrelated domain happens to be busy.

haesbaert:
> > I have to disagree here, it does matter, the configuration file changed again and the signal happens while you were in the blank spots, you miss the event, that's the whole point of signals being leveled and not edge triggered.
> 
> Perhaps you could give an example sequence of actions where it doesn't work? If the user changes the config file and then we reach one of the spots you indicated, we will certainly read the updated configuration (in fact, we'll do that even if the user forgets to send a signal at all, in those specific cases).
> 

I can't, but you end up processing the file twice for one signal, no ?

1 - user changed the configuration
2 - user sent a SIGHUP
3 - you caught the SIGHUP, so you _will_ read the new file
(here you're not watching for the new signal, Fiber1 is done and you're in Fiber2)
4 - you start reading the new file
5 - a signal is broadcasted and lost
6 - Fiber2 finishes updating configuration
7 - your loop restart
6 - Fiber1 starts watching for signals, there is none, so it just yields
7 - Fiber2 reads the configuration again <---


> The only difficult case is where a signal occurs after we start reading the configuration and before we start watching for the signal. But that can't happen because we have already started watching for the signal before starting to read the config.
> 
> > > Domain 1 is now busy doing some CPU stuff. Signal 1 occurs. The handler runs and sends a wake-up to domain 1's event loop. This won't be processed until domain 1 is ready, even though domain 2 is ready now.
> > 
> > 
> > Yes, if you mean that signal can be delayed until the receiving Domain blocks on a Fiber, yes it is delayed, if this is a problem, then the whole assumption of a concurrent scheduler is broken and we should delete it all and make it all preemptive through SIGALARM tricks :).
> 
> I would expect this to work (and it will, with Eio.Condition and lock-free waiters). In the example I gave:
> 
> 1. Domain 2 registered for a signal.
> 2. The signal happened.
> 3. Domain 2 is ready to process it, but it doesn't get delivered because an unrelated domain happens to be busy.

haesbaert:
What I might be missing, is that in your example, after start, one Fiber has always finished, but the one waiting the signal did not, so Fiber.both is always half complete.

PR #381, opened 2022-12-05
Make Eio_linux.wakeup signal-safe

talex5:
This will eventually allow Waiters to be used from signal handlers and GC finalizers (once that's lock-free too).

Previously, `wakeup` couldn't be used from a signal handler because it took a lock (to ensure that the eventfd wasn't closed while it wrote to it). Now, we avoid ever closing eventfds and simply keep free ones in a pool.

/cc @haesbaert


haesbaert:
This is nice, although the forever leak of fds is kidna evil, especially when we have a single process running tests with multiple `runs`.

I can fix the leak with barriers in the future, it's a pretty simple solution, I just have to write it.

haesbaert:
> That would be useful, but I don't see how it could be implemented. It's not just `all_domains`: signals and GC finalizers can also run in other systhreads and non-Eio domains.

We would have to make concessions like "you're only supposed to have Eio Domains", or X is only safe within an Eio context and unsafe if accessed outside.

lib_eio_linux/eio_linux.ml
```diff
@@ -964,7 +963,42 @@ module Low_level = struct
...
+  (* Take the next free eventfd from the pool, or create a new one if the pool's empty.
+     You might get a few spurious events from it as other threads are shutting down,
+     so you must be able to cope with that. *)
+
+  val put : Unix.file_descr -> unit
+  (* [put fd] adds [fd] to the free pool. *)
+end = struct
+  external eio_eventfd : int -> Unix.file_descr = "caml_eio_eventfd"
+
+  let free = Lf_queue.create ()
```
haesbaert, line 992: Can we change this to something else, `free` for me is the opposite of `alloc`.

PR #391, opened 2022-12-16
Update README for OCaml 5 release

talex5:
Currently waiting for the Docker image to finish pushing to hub so I can test that.


talex5:
Docker Hub is going to take hours, so I've left it updating to 5.0 in the Dockerfile itself for now.

PR #392, opened 2022-12-16
lib_linux: expose log src

bikallem:
Expose log source so that we can individually control log level. 

Related to https://github.com/mirleft/ocaml-tls/pull/461

### DIFF
--- a/lib_eio_linux/eio_linux.mli
+++ b/lib_eio_linux/eio_linux.mli
@@ -249,3 +249,5 @@ module Low_level : sig
       an ipaddress. *)
 
 end
+
+val src : Logs.src


bikallem:
@talex5 if you don't have any objections, could we merge this as it would make debugging a bit more productive. Thanks.

talex5:
I wonder if we should just get rid of all the debug logging in Eio_linux. Do you find it useful for anything?

bikallem:
I have found them to be somewhat useful, sometimes. Perhaps, leave them until 1.0? In any case, until they are completely removed, I would like to be able to control log levels to control log noise so that such debugging sessions are a bit more manageable and productive.

avsm:
I'd prefer to reduce the dependency cone and get rid of the Logs.  When I need debug logging in the backend, I'm just putting in whatever I need.  It's rarely useful to turn on the backend specific logging as part of a bigger application (which is the point of the Logs library).

bikallem:
closing this PR since #403 supersedes it.

PR #394, opened 2022-12-17
Add a warning to the tutorial about Fiber.first

talex5:
/cc @haesbaert

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,11 @@ If you want to make an operation non-cancellable, wrap it with `Cancel.protect`
 - : unit = ()
 ```
 
+Note: using `Fiber.first` to ensure that *exactly one* of two actions is performed is not reliable.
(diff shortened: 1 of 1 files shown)


haesbaert:
Makes sense, I just wonder if it's worth mentioning uring, or if people would decide on using the luv backend to abuse the semantics of cancellation, or if people would write code on their mac, test, think it's correct and then break once they run on uring.

Also, I think we should provide one example of correct input IO multiplexing in the manual, and clarify what are the semantics we can guarantee, I'm guessing this would would be a `Fiber.first` with three fibers: two draining a flow and feeding a `Stream` that the third fiber drains

talex5:
Yes, that could be made clearer. I used uring to show why this isn't fixable in general, but indeed we don't want to suggest it's OK on luv either (it isn't).

We might want to have a separate HOWTO document listing various problems and solutions and link that.

PR #395, opened 2022-12-19
time: clarify Eio.Time.now is UTC

bikallem:


talex5:
https://v2.ocaml.org/api/Unix.html#VALtime says:
> Return the current time since 00:00:00 GMT, Jan. 1, 1970, in seconds.

I'm not sure what "in UTC" means for a float. The important thing is where it is measured from. It can be displayed in any timezone.

talex5:
Thanks!

PR #398, opened 2023-01-03
Make Eio.Semaphore lock-free

talex5:
This also changes the semaphore benchmark. Before, two domains just passed control back and forth, which was really just timing how long it takes to wake a domain. Now, we keep the domains busy while waiting for the semaphore. We also now have 4 domains trying to use 2-4 resources, which is a more realistic use of a semaphore.

This is a little faster in the single domain case, and quite a bit faster with multiple domains:
semaphore

### DIFF
--- a/lib_eio/sem_state.ml
+++ b/lib_eio/sem_state.ml
@@ -1,0 +1,176 @@
+(* A lock-free semaphore, using Cells.
+
+   We have a number of resources (1 in the case of a mutex). Each time a user
+   wants a resource, the user decrements the counter. When finished with the
+   resource, the user increments the counter again.
+
+   If there are more users than resources then the counter will be negative. If
+   a user decrements the counter to a non-negative value then it gets ownership
+   of one of the free resources. If it decrements the counter to a negative
+   value then it must wait (by allocating a cell). When a user with a resource
+   increments the counter *from* a negative value, that user is responsible for
+   resuming one waiting cell (transferring ownership of the resource). This
+   ensures that every waiter will get woken exactly once.
+
+   Cancellation
+
+   We could consider cancelling a request to be simply replacing the callback
+   with a dummy one that immediately releases the resource. However, if callers
+   keep cancelling then the list of cancelled requests would keep growing.
+
+   Instead, we'd like cancellation simply to undo the effects of suspending, by
+   incrementing the counter and marking the cell as Finished (so that the
+   resumer will ignore it and move on to the next waiter, and the Finished
+   cell can be freed).
+
+   If the cancelling user increments from a negative value then it is responsible
+   for waking one user, which is fine as it is waking itself. However, it may find
+   itself incrementing from a non-negative one if it is racing with a resumer
+   (if the count is non-negative then once all current operations finish there
+   would be no suspended users, so the process of waking this user must have
+   already begun).
+
+   To handle this, a cancelling user first transitions the cell to In_transition,
+   then increments the counter, then transitions to the final Finished state,
+   in the usual case where it incremented from a negative value.
+
+   If a resumer runs at the same time then it may also increment the counter
+   from a non-negative value and try to wake this cell. It transitions the cell
+   from In_transition to Finished. The cancelling user will notice this when it
+   fails to CAS to Finished and can handle it.
+
+   If the cancelling user sees the Finished state after In_transition then it
+   knows that the resuming user has transferred to it the responsibility of
+   waking one user. If the cancelling user is also responsible for waking one
+   user then it performs an extra resume on behalf of the resuming user.
+
+   Finally, if the cancelling user is not responsible for waking anyone (even
+   itself) then it leaves the cell in In_transition (the CQS paper uses a
+   separate Refused state, but we don't actually need that). This can only
+   happen when a resume is happening at the same time. The resumer will
+   transition to Finished, creating an obligation to resume, but we've just
+   done that anyway. We know this In_transition state can't last long because
+   at the moment when the canceller incremented the counter all current
+   waiters, including itself, were in the process of being resumed. *)
+
+module Cell = struct
+  type _ t =
+    | In_transition                     (* The suspender will try to CAS this soon. *)
+    | Request of (unit -> unit)         (* Waiting for a resource. *)
+    | Finished                          (* Ownership of the resource has been transferred,
+                                           or the suspender cancelled. *)
+
+  let init = In_transition
+  (* We only resume when we know another thread is suspended or in the process of suspending. *)
+
+  let segment_order = 2
+
+  let dump f = function
+    | Request _ -> Fmt.string f "Request"
+    | Finished -> Fmt.string f "Finished"
+    | In_transition -> Fmt.string f "In_transition"
+end
+
+module Cells = Cells.Make(Cell)
+
+type cell = unit Cell.t
+
+type t = {
+  state : int Atomic.t;         (* Free resources. Negative if there are waiters waiting. *)
+  cells : unit Cells.t;
+}
+
+type request = t * unit Cells.segment * unit Cell.t Atomic.t
+
+(* Wake one waiter (and give it the resource being released). *)
+let rec resume t =
+  let cell = Cells.next_resume t.cells in
+  match (Atomic.exchange cell Finished : cell) with
+  | Request r ->
+    (* The common case: there was a waiter for the value.
+       We pass ownership of the resource to it. *)
+    r ()
+  | Finished ->
+    (* The waiter has finished cancelling. Ignore it and resume the next one. *)
+    resume t
+  | In_transition ->
+    (* The consumer is in the middle of doing something and will soon try to
+       CAS to a new state. It will see that we got there first and handle the
+       resume when it's done. *)
+    ()
+
+(* [true] on success, or [false] if we need to suspend.
+   You MUST call [suspend] iff this returns [false].
+   The reason for splitting this is because e.g. [Semaphore] needs to get
+   the continuation for the fiber between [acquire] and [suspend]. *)
+let acquire t =
+  let s = Atomic.fetch_and_add t.state (-1) in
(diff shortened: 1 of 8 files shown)


lib_eio/sem_state.ml
```diff
@@ -0,0 +1,189 @@
...
+    | Cancelling ->
+      (* The waker has started cancelling. Let it know we want to resume it
+         and then let it handle it (it may also have finished cancelling but
+         decided there's nothing left to do anyway). *)
+      if not (Atomic.compare_and_set cell Cancelling Resumed) then aux ()
+    | Resumed ->
+      (* This state is unreachable because we (the provider) haven't set this yet *)
+      assert false
+  in
+  aux ()
```
polytypic, line 113: I do not yet fully understand the `Cells` abstraction.  However, looking at the code here it seems that `Resumed` is a terminal state.  If that is the case, then it may be safe to always exchange the state to `Resumed`.  Like I said, I do not yet fully understand `Cells` so this might be wrong.

However, assuming that exchanging to `Resumed` is a safe, then the above could be changed to eliminate the loop:

```ocaml
  let rec resume t =
    let cell = Cells.next_resume t.cells in
    match (Atomic.exchange cell Resumed : cell) with
    | Request r ->
      (* The common case: there was a waiter for the value *)
      r ()
    | Cancelled ->
      (* The waker had finished cancelling. Ignore it and resume the next one. *)
      resume t
    | (Empty
       (* The consumer had reserved this cell but not yet stored the request.
          We placed Resumed there and the consumer will handle it soon. *)
      | Cancelling
        (* The waker had started cancelling. We let it know we want to resume it
           and then let it handle it. *)
      | Resumed
        (* We lost the race, just ignore *)) ->
      ()
```

Assuming it is safe to always set to `Resumed`, the above version avoids the loop overhead and only does a single atomic read-write.  Assuming there are not many racing to do the same this may be slightly faster. This also likely results in a smaller amount of machine code.

talex5: > This also likely results in a smaller amount of machine code.

We could combine the `Empty` and `Cancelled` handlers to save some code though (they're identical). But maybe it's clearer as it is.

polytypic: Let me just play a devil's advocate a bit here. :)

Perhaps one possibility might be to merge `Resumed` and `Cancelled` into a single, let's say, `Terminal` state.  There is a race to set the cell to `Terminal` state.  The action depends on the state from which the `Terminal` state was reached.

> One side-effect is that calling cancel twice currently reliably raises Invalid_argument

About `cancel`.  Hmm... Under which circumstances one might call `cancel` twice?  (Asking because I'm not sure I understand the concern.)

Assuming there is a race to call `cancel`, I would tend to think that it is important that `cancel` returns `true` at most once.  This way any follow up actions to successful cancellation would only be performed once.  If this makes sense, then the API might be more user friendly without raising Invalid_argument.  John Ousterhout talks about this in his talk on software design [here](https://youtu.be/bmSAYlu0NcY?t=1312).

talex5:
> Perhaps one possibility might be to merge Resumed and Cancelled into a single, let's say, Terminal state.

Makes sense. I updated the PR (I called it Finished though, mainly because I started just before seeing your comment!).

> Under which circumstances one might call cancel twice?

It would always be a bug. I was just noting it as a change to this module's external behaviour (but it's only used by Semaphore anyway, and that will only call it once).

Assuming the new version is correct, I think we should stop and admire that fact that our version has only 4 states, while the original paper had 9! (https://arxiv.org/pdf/2111.12682.pdf section 3.1 plus the REFUSE state added in 3.2)

Thanks!

lib_eio/sem_state.ml
```diff
@@ -0,0 +1,181 @@
...
+      );
+      true
+    ) else false          (* We got resumed first *)
+  | Finished -> false     (* We got resumed first *)
+  | Cancelling -> invalid_arg "Already cancelling!"
+  | Empty ->
+    (* To call [cancel] the user needs a [request] value,
+       which they only get once we've reached the [Request] state.
+       [Empty] is unreachable from [Request]. *)
+    assert false
```
polytypic, line 169: Hmm... Speaking of reducing states, I wonder whether `Empty` and `Cancelling` states could actually be merged.

Is this the only place where they are handled differently?  And both indicate a programming error?

lib_eio/semaphore.ml
```diff
@@ -1,55 +1,42 @@
...
-    Ctf.note_read t.id;
-    t.state <- Free (pred n);
-    Mutex.unlock t.mutex
+let acquire t =
+  if not (Sem_state.acquire t.state) then (
+    (* No free resources.
+       We must wait until one of the existing users increments the counter and resumes us.
+       It's OK if they resume before we suspend; we'll just pick up the token they left. *)
+    Suspend.enter_unchecked (fun ctx enqueue ->
+        match Sem_state.suspend t.state (fun () -> enqueue (Ok ())) with
```
polytypic, line 24: Is there a potential here to avoid the `Suspend.enter_unchecked` call (which, I assume, captures the continuation) by exposing the `Sem_state` protocol a bit more.  Basically, require the acquirer to separately create the cell.  After creating the cell, the acquirer would then read the cell state (as is done in `Sem_state.suspend` at the moment) before capturing the continuation.  If the state has already been set to `Finished`, then there is no need to capture a continuation.  Otherwise capture continuation and proceed as before.  Basically, this results in a kind of double-checked pattern.  This could improve performance in highly contested cases where it is possible that a resumer actually manages to see the `InTransition` (or `Empty`) state.

talex5: I don't think so. This case wouldn't work:

1. The suspender decrements the count to say they're planning to suspend.
2. The resumer resumes the cell.
3. The suspender creates the cell.

We have to be able to initialise cells before the suspender does anything beyond modifying the counter. We don't want it creating cells before changing the counter because that includes the fast path (where we don't need a cell at all).

lib_eio/sem_state.ml
talex5: Good point - done :-)

lib_eio/semaphore.ml
polytypic: Just to be clear, this is what I had in mind:

```diff
diff --git a/lib_eio/sem_state.ml b/lib_eio/sem_state.ml
index 24a301c..7ffebf6 100644
--- a/lib_eio/sem_state.ml
+++ b/lib_eio/sem_state.ml
@@ -109,6 +109,26 @@ let acquire t =
      which happens if we decremented *from* a positive one. *)
   s > 0
 
+let prepare_suspend t =
+  Cells.next_suspend t.cells
+
+let is_in_transition (_, (cell: cell Atomic.t)) =
+  match Atomic.get cell with
+  | In_transition -> true
+  | Finished | Request _ -> false
+
+let perform_suspend t (segment, (cell: cell Atomic.t)) k : request option =
+  if Atomic.compare_and_set cell In_transition (Request k) then Some (t, segment, cell)
+  else match Atomic.get cell with
+    | Finished ->
+      (* We got resumed before we could add the waiter. *)
+      k ();
+      None
+    | Request _ | In_transition ->
+      (* These are unreachable from the previously-observed non-In_transition state
+         without us taking some action first *)
+      assert false
+
 let suspend t k : request option =
   let (segment, cell) = Cells.next_suspend t.cells in
   if Atomic.compare_and_set cell In_transition (Request k) then Some (t, segment, cell)
diff --git a/lib_eio/semaphore.ml b/lib_eio/semaphore.ml
index 2733be5..a90fa64 100644
--- a/lib_eio/semaphore.ml
+++ b/lib_eio/semaphore.ml
@@ -20,21 +20,24 @@ let acquire t =
     (* No free resources.
        We must wait until one of the existing users increments the counter and resumes us.
        It's OK if they resume before we suspend; we'll just pick up the token they left. *)
-    Suspend.enter_unchecked (fun ctx enqueue ->
-        match Sem_state.suspend t.state (fun () -> enqueue (Ok ())) with
-        | None -> ()   (* Already resumed *)
-        | Some request ->
-          Ctf.note_try_read t.id;
-          match Fiber_context.get_error ctx with
-          | Some ex ->
-            if Sem_state.cancel request then enqueue (Error ex);
-            (* else already resumed *)
-          | None ->
-            Fiber_context.set_cancel_fn ctx (fun ex ->
-                if Sem_state.cancel request then enqueue (Error ex)
-                (* else already resumed *)
-              )
-      )
+    let segment_cell = Sem_state.prepare_suspend t.state in
+    if Sem_state.is_in_transition segment_cell then (
+      Suspend.enter_unchecked (fun ctx enqueue ->
+          match Sem_state.perform_suspend t.state segment_cell (fun () -> enqueue (Ok ())) with
+          | None -> ()   (* Already resumed *)
+          | Some request ->
+            Ctf.note_try_read t.id;
+            match Fiber_context.get_error ctx with
+            | Some ex ->
+              if Sem_state.cancel request then enqueue (Error ex);
+              (* else already resumed *)
+            | None ->
+              Fiber_context.set_cancel_fn ctx (fun ex ->
+                  if Sem_state.cancel request then enqueue (Error ex)
+                  (* else already resumed *)
+                )
+        )
+    )
   );
   Ctf.note_read t.id
```

talex5: Ah, I see. Pushing some of the work to outside of the suspend in case we can skip the suspend by the time it's done, at the cost of an extra `Atomic.get`.

I tried measuring it by incrementing `slow` and `fast` counters in `is_in_transition` and running the benchmark gives:
```
slow=850475, fast=75, frac=0.01%
```
So probably not worth it, I think.

polytypic (approved):

lib_eio/semaphore.ml
talex5: OK, I guess I should have disabled the single-process runs before measuring that... it's `slow=51221, fast=98, frac=0.19%` now (0.19% of the slow path cases). But I don't think that changes anything.

PR #399, opened 2023-01-04
Add Buf_read.BE/LE parsers

### LINKED ISSUE #390: Add `Buf_read.LE` `Buf_read.BE`
`Buf_write` exposes `LE` and `BE` modules for writing various sized integers and floats.

Would it make sense to add the corresponding modules to `Buf_read`?

Cjen1:
This PR adds the relevant parsers for integer primitives, equivalent to `Buf_write.BE/LE` to `Buf_read`.

It resolves #390.

Completed portions:
- [x] Big endian parsers
- [x] Big endian tests/fuzzers
- [ ] Little endian parsers
- [ ] Little endian tests/fuzzers
- [ ] uint48 parsers


Cjen1:
This PR doesn't yet have the little endian equivalents, nor the uint48 since I wanted to check that the fuzzers etc were acceptable first.

fuzz/fuzz_buf_read.ml
```diff
@@ -61,6 +61,48 @@ module Model = struct
       t := String.with_index_range s ~first:1;
       String.get_head s
 
+  module BE = struct
+    let uint16 t =
+      match !t with
+      | "" -> raise End_of_file
```
talex5, line 67: This case seems unnecessary. The next one will catch it anyway.

tests/buf_reader.md
```diff
@@ -393,6 +393,80 @@ Exception: Failure "skip_while1".
 - : string = "bbbccc"
 ```
 
+## Big Endian
+```ocaml
+# let i = R.of_flow mock_flow ~max_size:100;;
```
talex5, line 398: Note: the mock flow is used to check that it's pulling the right number of bytes at a time. For fixed-sized reads the output isn't very interesting. You might want to use `R.parse_exn` instead (as in the "Convenience wrapper" section).

talex5 (commented):
Looks good so far!

tests/buf_reader.md
talex5: I'd just do one test for each function, like this:
```
# R.(parse_string_exn BE.uint16) "\128\001" |> Printf.sprintf "0x%x";;
- : string = "0x8001"
```
If it gets that right (using an example with the top-bit set, using all bytes, and not the same LE and BE), it's very unlikely to fail on anything else (and the fuzz tests will check everything anyway). We only really need to check that we read the right number of bytes and didn't confuse signed/unsigned or endianness.

Cjen1: Excellent I'll do that, I had forgotten the backslash notation!

Cjen1:
@talex5 This should be a completed PR. 

I think that the uint48 implementation may not be fully correct but I'm testing to make sure now.
(I think `Int64.of_int32` does not preserve the msb)

Cjen1:
> I think that the uint48 implementation may not be fully correct but I'm testing to make sure now. (I think `Int64.of_int32` does not preserve the msb)

It does not preserve the sign (or rather it preserves the value rather than the bits).
This is now fixed by doing everything in the int64 representation.

lib_eio/buf_read.mli
```diff
@@ -107,6 +107,48 @@ val string : string -> unit parser
...
+(** Big endian parsers *)
+module BE : sig
+  val uint16 : int parser
+  (** [uint16] returns the next 2 bytes as the lower 16 bits of an [int] in big-endian byte order *)
+
+  val uint32 : int32 parser
+  (** [uint32] returns the next 4 bytes as an [int32] in big-endian byte order *)
+
+  val uint48 : int64 parser
+  (** [uint48] returns the next 6 bytes as an [int48] in big-endian byte order *)
```
talex5, line 119:
```suggestion
  (** [uint48] returns the next 6 bytes as an [int64] in big-endian byte order *)
```
Assuming this is supposed to be an OCaml type. Though it seems unnecessary to mention it, as it's already in the function signature.
```suggestion
  (** [uint48] parses 6 bytes as a 48-bit unsigned big-endian integer. *)
```

lib_eio/buf_read.ml
```diff
@@ -143,6 +143,98 @@ let as_flow t =
...
+  let uint32 t = 
+    ensure t 4;
+    let data = Bigstringaf.get_int32_le t.buf t.pos in
+    consume t 4;
+    data
+
+  let uint48 t = 
+    ensure t 6;
+    let lower_16 = Bigstringaf.get_int16_le t.buf t.pos |> Int64.of_int in
+    let middle_16 = Bigstringaf.get_int16_le t.buf (t.pos + 2) |> Int64.of_int in
```
talex5, line 208: This might be better (haven't benchmarked it though):
```suggestion
    let lower_32 = Bigstringaf.get_int32_le t.buf t.pos |> Int64.of_int |> Int64.log_and 0xffffffffL in
```
Then the total is just `Int64.(logor lower_32 (shift_left upper_16 32))`.

fuzz/fuzz_buf_read.ml
```diff
@@ -61,6 +61,126 @@ module Model = struct
       t := String.with_index_range s ~first:1;
       String.get_head s
 
+  module BE = struct
+    let uint16 t =
+      match !t with
+      | s when String.length s < 2 -> raise End_of_file
+      | s ->
+          t := String.with_index_range s ~first:2;
+          s |> Bytes.unsafe_of_string |> fun b -> Bytes.get_uint16_be b 0
```
talex5, line 70: If you move `BE` after `take` then you can use that to simplify it a lot:
```suggestion
  let take_bytes n t = Bytes.unsafe_of_string (take n t)

  module BE = struct
    let uint16 t = Bytes.get_uint16_be (take_bytes 2 t) 0
```

talex5 (commented):
Looks right to me, but some simplifications/speedups are possible I think.

fuzz/fuzz_buf_read.ml
talex5: Actually, we don't even need to convert to bytes:
```ocaml
let uint16 t = Stdlib.String.get_uint16_be (take 2 t) 0
```

lib_eio/buf_read.ml
Cjen1: I'm not sure this is correct for `-1L` since the of_int moves the sign bit such that the value is the same (but not the bits).
This was the correctness bug I was concerned about previously.

`logor` may be faster though than just adding.

talex5: It doesn't move the sign bit, it just replicates it into the extra bits (which is correct for signed numbers but not for unsigned ones). The `logand` gets rid of the extra copies:
```
# Int64.of_int32 0xe7654321l |> Printf.sprintf "0x%Lx";;
- : string = "0xffffffffe7654321"
# Int64.of_int32 0xe7654321l |> Int64.logand 0xffffffffL |> Printf.sprintf "0x%Lx";;
- : string = "0xe7654321"
```

Cjen1: Good to know! I've left the fuzz model using the 3 * 16 bit one for validation.

talex5:
Thanks! I squashed and rebased it on top of #402 to avoid needing the `Stdlib.` prefix.

PR #403, opened 2023-01-06
Remove debug-level logging

talex5:
This doesn't seem useful and we also have tracing anyway. It clutters the logs when people turn on debug-level logging for their application. We might want to remove all logging eventually, but this is a start.

Alternative to #392.

It seems to make the noop benchmark very slightly faster:

noop

/cc @bikallem


bikallem:
Thanks for this PR. Yes, this would solve #392.

PR #408, opened 2023-01-16
net: add run_server to run eio servers

bikallem:
This PR adds `run_server` - to be used to spawn network socket server. The current version supports spawning both single domain and multi domain servers. The default is a single domain concurrent server.

Based on discussion in https://github.com/mirage/ocaml-cohttp/issues/961#issuecomment-1378661543.

/cc @talex5 @patricoferris

### DIFF
--- a/lib_eio/net.ml
+++ b/lib_eio/net.ml
@@ -269,3 +269,31 @@ let with_tcp_connect ?(timeout=Time.Timeout.none) ~host ~service t f =
   | exception (Exn.Io _ as ex) ->
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "connecting to %S:%s" host service
+
+(* Run a server loop in a single domain. *)
+let run_server_loop ~connections ~on_error ~stop listening_socket connection_handler =
+  Switch.run @@ fun sw ->
+  let rec accept () =
+    Semaphore.acquire connections;
+    accept_fork ~sw ~on_error listening_socket (fun conn addr ->
+        Fun.protect (fun () -> connection_handler conn addr)
+            ~finally:(fun () -> Semaphore.release connections)
+      );
+    accept ()
+  in
+  match stop with
+  | None -> accept ()
+  | Some stop -> Fiber.first accept (fun () -> Promise.await stop)
+
+let run_server ?(max_connections=Int.max_int) ?(additional_domains) ?stop ~on_error listening_socket connection_handler : 'a =
+  if max_connections <= 0 then invalid_arg "max_connections";
+  Switch.run @@ fun sw ->
+  let connections = Semaphore.make max_connections in
+  let run_server_loop () = run_server_loop ~connections ~on_error ~stop listening_socket connection_handler in
+  additional_domains |> Option.iter (fun (domain_mgr, domains) ->
+      if domains < 0 then invalid_arg "additional_domains";
+      for _ = 1 to domains do
+        Fiber.fork ~sw (fun () -> Domain_manager.run domain_mgr (fun () -> ignore (run_server_loop () : 'a)))
+      done;
+    );
+  run_server_loop ()
--- a/lib_eio/net.ml
+++ b/lib_eio/net.ml
@@ -170,6 +170,8 @@ class virtual listening_socket = object
   method virtual close : unit
 end
 
+type connection_handler = stream_socket -> Sockaddr.stream -> unit
+
 let accept ~sw (t : #listening_socket) = t#accept ~sw
 
 let accept_fork ~sw (t : #listening_socket) ~on_error handle =
--- a/lib_eio/net.mli
+++ b/lib_eio/net.mli
@@ -191,11 +191,17 @@ val accept_fork :
   on_error:(exn -> unit) ->
   connection_handler ->
   unit
-(** [accept_fork socket fn] accepts a connection and handles it in a new fiber.
+(** [accept_fork ~sw ~on_error socket fn] accepts a connection and handles it in a new fiber.
 
     After accepting a connection to [socket], it runs [fn flow client_addr] in a new fiber.
 
-    [flow] will be closed when [fn] returns. *)
+    [flow] will be closed when [fn] returns. The new fiber is attached to [sw].
+
+    @param on_error Called if [connection_handler] raises an exception.
+                    This is typically a good place to log the error and continue.
+                    If the exception is an {!Eio.Io} error then the caller's address is added to it.
+                    If you don't want to handle connection errors,
+                    use [~on_error:raise] to cancel the caller's context. *)
 
 val accept_sub :
   sw:Switch.t ->
@@ -204,6 +210,38 @@ val accept_sub :
   (sw:Switch.t -> stream_socket -> Sockaddr.stream -> unit) ->
   unit
 [@@deprecated "Use accept_fork instead"]
+
+(** {2 Running Servers} *)
+
+val run_server :
+  ?max_connections:int ->
+  ?additional_domains:(#Domain_manager.t * int) ->
+  ?stop:'a Promise.t ->
+  on_error:(exn -> unit) ->
+  #listening_socket ->
+  connection_handler ->
+  'a
+(** [run_server ~on_error sock connection_handler] establishes a concurrent socket server [s].
+
+    It accepts incoming client connections on socket [sock] and handles them with {!accept_fork}
+    (see that for the description of [on_error] and [connection_handler]).
+
+    {b Running a Parallel Server}
+
+    By default [s] runs on a {e single} OCaml {!module:Domain}. However, if [additional_domains:(domain_mgr, domains)]
+    parameter is given, then [s] will spawn [domains] additional domains and run accept loops in those too.
+    In such cases you must ensure that [connection_handler] only accesses thread-safe values.
+    Note that having more than {!Domain.recommended_domain_count} domains in total is likely to result in bad performance.
+
+    @param max_connections The maximum number of concurrent connections accepted by [s] at any time.
+                           The default is [Int.max_int].
+    @param stop Resolving this promise causes [s] to stop accepting new connections.
+                [run_server] will wait for all existing connections to finish and then return.
+                This is useful to upgrade a server without clients noticing.
+                To stop immediately, cancelling all connections, just cancel [s]'s fiber instead.
+    @param on_error Connection error handler (see {!accept_fork}).
+    @raise Invalid_argument if [max_connections <= 0].
+                            if [additional_domains = (domain_mgr, domains)] is used and [domains < 0]. *)
 
 (** {2 Datagram Sockets} *)
 
--- a/lib_eio/net.mli
+++ b/lib_eio/net.mli
@@ -182,11 +182,14 @@ val accept :
     The new socket will be closed automatically when [sw] finishes, if not closed earlier.
     If you want to handle multiple connections, consider using {!accept_fork} instead. *)
 
+type connection_handler = stream_socket -> Sockaddr.stream -> unit
+(** [connection_handler] handles incoming connections from a listening socket. *)
+
 val accept_fork :
   sw:Switch.t ->
   #listening_socket ->
   on_error:(exn -> unit) ->
-  (stream_socket -> Sockaddr.stream -> unit) ->
+  connection_handler ->
   unit
 (** [accept_fork socket fn] accepts a connection and handles it in a new fiber.
 
--- a/tests/network.md
+++ b/tests/network.md
@@ -259,7 +259,7 @@ Extracting file descriptors from Eio objects:
   traceln "Listening socket has Unix FD: %b" (Eio_unix.FD.peek_opt server <> None);
   let have_client, have_server =
     Fiber.pair
-      (fun () -> 
+      (fun () ->
          let flow = Eio.Net.connect ~sw net addr in
          (Eio_unix.FD.peek_opt flow <> None)
       )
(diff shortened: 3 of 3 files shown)


bikallem:
The `~additional_domains` bit doesn't work with luv backend. Should this be reflected in the API doc?

lib_eio/net.ml
```diff
@@ -267,3 +269,52 @@ let with_tcp_connect ?(timeout=Time.Timeout.none) ~host ~service t f =
...
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "connecting to %S:%s" host service
+
+let accept_fork_domain ~on_error (domain_mgr, domains) (t : #listening_socket) connection_handler () =
+  Switch.run @@ fun sw ->
+  let flow, addr = accept ~sw t in
+  Fun.protect
+    ~finally:(fun () ->
+      Flow.close flow;
+      Semaphore.release domains
```
polytypic, line 279: Hmm...  Assuming this `release` matches the below `acquire` call

```ocaml
    (fun () ->
      Domain_manager.run domain_mgr (fun () ->
        Semaphore.acquire domains ;
```

then this seems out of place to me unless there is a strong guarantee that `Domain_manager.run` always calls the passed function and e.g. never raises exceptions.

lib_eio/net.ml
```diff
@@ -267,3 +269,52 @@ let with_tcp_connect ?(timeout=Time.Timeout.none) ~host ~service t f =
   | exception (Exn.Io _ as ex) ->
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "connecting to %S:%s" host service
+
+let accept_fork_domain ~on_error (domain_mgr, domains) (t : #listening_socket) connection_handler () =
+  Switch.run @@ fun sw ->
+  let flow, addr = accept ~sw t in
+  Fun.protect
+    ~finally:(fun () ->
+      Flow.close flow;
```
polytypic, line 278: Hmm... Is this `close` necessary?  The documentation [seems to suggests](https://github.com/ocaml-multicore/eio#networking) that a switch automatically closes attached resources.

lib_eio/net.ml
```diff
@@ -267,3 +269,52 @@ let with_tcp_connect ?(timeout=Time.Timeout.none) ~host ~service t f =
...
+    | None -> (fun () -> accept_fork ~sw listening_socket ~on_error connection_handler)
+  in
+  let rec loop () =
+    Fiber.first
+      (fun () ->
+        Semaphore.acquire connections;
+        acceptor ();
+        loop ()
+      )
+      (fun () -> Promise.await shutdown )
```
polytypic, line 318: Hmm... Took me moment to understand this pattern.

If I understand correctly, then for each connection two fibers are created.  The second fiber merely waits for the shutdown signal and if that arrives, then the other fiber will be cancelled.

This makes me wonder, however, whether it would be possible to use something like a single switch for the cancellation of all the connections running under that switch.

Perhaps one could even remove the `shutdown` argument and instead have the whole server run under a switch that one can then cancel (e.g. by using `first` outside of the server).

bikallem: The idea with the current implementation is to gracefully stop the accepting loop, rather than (abruptly?) cancelling the currently running fibers. However, of course users can implement the cancel functionality as you suggested outside of eio.

lib_eio/net.ml
bikallem: Good catch. Indeed, it doesn't need to be.

bikallem:
@polytypic Thank you for the review. I have pushed some changes that should address the raised issues.

lib_eio/net.ml
talex5: This system is not tail-recursive. Every new connection starts another fiber waiting for shutdown.

Instead, you can replace `?shutdown` with a `?sw` argument to let the connections outlive the accept loop if desired, and then use Eio's built-in cancellation.

lib_eio/net.ml
```diff
@@ -267,3 +269,49 @@ let with_tcp_connect ?(timeout=Time.Timeout.none) ~host ~service t f =
...
+    | Some (domain_mgr, domains) ->
+      (if domains < 1 then invalid_arg "additional_domains");
+      let domains = Semaphore.make domains in
+      accept_fork_domain (domain_mgr, domains) listening_socket ~on_error connection_handler
+    | None -> (fun () -> accept_fork ~sw listening_socket ~on_error connection_handler)
+  in
+  let rec loop () =
+    Fiber.first
+      (fun () ->
+        Semaphore.acquire connections;
```
talex5, line 311: Make sure this is released on error too!

lib_eio/net.ml
```diff
@@ -267,3 +269,49 @@ let with_tcp_connect ?(timeout=Time.Timeout.none) ~host ~service t f =
...
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "connecting to %S:%s" host service
+
+  let accept_fork_domain ~on_error (domain_mgr, domains) (t : #listening_socket) connection_handler () =
+  Switch.run @@ fun sw ->
+  let flow, addr = accept ~sw t in
+  Fun.protect
+    (fun () ->
+      Semaphore.acquire domains ;
+      Domain_manager.run domain_mgr (fun () ->
```
talex5, line 279: I don't understand what this is trying to do. You seem to be spawning a new domain just to handle one connection, which seems highly inefficient.

I would expect `run_server` to start one accept loop for each domain. That's how the existing cohttp-eio version does it, right?

lib_eio/net.ml
bikallem: Done.

lib_eio/net.ml
bikallem: I have clarified the cancellation in documentation explicitly.

lib_eio/net.mli
```diff
@@ -202,6 +205,42 @@ val accept_sub :
   unit
 [@@deprecated "Use accept_fork instead"]
 
+(** {2 Running Servers} *)
+
+val run_server :
+  ?max_connections:int ->
+  ?shutdown:unit Promise.t ->
```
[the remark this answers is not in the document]
talex5: From the user's point of view, I'd expect something like:
```ocaml
Switch.run @@ fun sw ->
Fiber.first
  (fun () -> run_server ~sw connection_handler)
  (fun () -> ... decide to shut down ...);
(* At this point, the accept loop(s) are finished
   but existing connections continue to be served in sw *)
```
Note that shutting down the accept loops in other domains (if any) while letting the domains continue handling their existing connections will require a bit of care. Internally, a promise shared between domains may well be a good way to do that (but you only need to await it once per domain, not once per connection).

lib_eio/net.ml
bikallem: Yes, the cohttp-eio has one acceptor per domain.

>You seem to be spawning a new domain just to handle one connection, which seems highly inefficient.

Good point. I completely overlooked the efficiency of spawning domain per connection.

lib_eio/net.mli
```diff
@@ -202,6 +205,42 @@ val accept_sub :
...
+  ?additional_domains:(#Domain_manager.t * int) ->
+  ?on_error:(exn -> unit) ->
```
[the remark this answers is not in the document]
bikallem: Changed ~on_error to required.

lib_eio/net.mli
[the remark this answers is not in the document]
bikallem: I have removed the shtudown logic.

lib_eio/net.ml
bikallem: the accept loop is similar to cohttp one now.

tests/network.md
```diff
@@ -403,6 +403,85 @@ If the fork itself fails, we still close the connection:
...
+      Eio.Flow.shutdown flow `Send;
+      let msg = read_all flow in
+      traceln "client received: %S" msg;
+      if id < clients then () else Promise.resolve set_shutdown ()
+    in
+    let connection_handler clock flow _addr =
+      Fun.protect (fun () ->
+        let msg = read_all flow in
+        traceln "Server received: %S" msg;
+        Eio.Time.sleep clock 0.01
```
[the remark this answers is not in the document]
bikallem: Done.

lib_eio/net.ml
bikallem: Done.

tests/network.md
```diff
@@ -403,6 +403,85 @@ If the fork itself fails, we still close the connection:
...
+val run_eio_server :
+  max_conn:int ->
+  clients:int ->
+  < clock : #Eio.Time.clock; net : #Eio.Net.t; .. > -> Switch.t -> unit =
+  <fun>
+```
+max_connections = 1
+
+```ocaml
+# Eio_main.run @@ fun env ->
```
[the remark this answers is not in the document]
bikallem: I am getting some issues while attempting to use mock to test `run_server`. I have reverted to using only one client using `eio_main` for now.

bikallem:
@talex5 I believe I have addressed all of your points. Could we include this PR in the next release please.

polytypic (approved):
LGTM

talex5 (commented):
I see you scaled back the tests since last time. I'm putting them back (using mocks) and fixing up some bugs that uncovered...

tests/network.md
```diff
@@ -403,6 +403,53 @@ If the fork itself fails, we still close the connection:
...
+val run_client : < net : #Eio.Net.t; .. > -> Switch.t -> unit = <fun>
+```
+
+```ocaml
+# Eio_main.run @@ fun env ->
+  try
+    Switch.run @@ fun sw ->
+    Fiber.all 
+      [(fun () -> test_run_server env sw)
+      ;(fun () -> run_client env sw; Switch.fail sw Graceful_shutdown)
```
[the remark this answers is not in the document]
bikallem: It seems doing below doesn't invoke the error handler as it does in `accept_fork`
```
  let on_error = function
    | Graceful_shutdown -> print_string "graceful"; ()
    | ex -> traceln "Error handling connection: %s" (Printexc.to_string ex)
  in
```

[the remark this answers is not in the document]
bikallem: I have changed to raising the exn rather than failing the switch along the lines of

```
let test_address addr ~net sw =
  let server = Eio.Net.listen net ~sw ~reuse_addr:true ~backlog:5 addr in
  Fiber.both
    (fun () -> run_server ~sw server)
    (fun () ->
      run_client ~sw ~net ~addr;
      traceln "Client finished - cancelling server";
      raise Graceful_shutdown
    )
```

lib_eio/net.ml
```diff
@@ -267,3 +269,25 @@ let with_tcp_connect ?(timeout=Time.Timeout.none) ~host ~service t f =
...
+      (fun () ->
+        Semaphore.acquire connections;
+        accept_fork ~sw ~on_error listening_socket connection_handler;
+      )
+      ~finally:(fun () -> Semaphore.release connections);
+    accept sw
+  in
+  (match additional_domains with
+   | Some (domain_mgr, domains) ->
+     if domains < 1 then invalid_arg "additional_domains";
```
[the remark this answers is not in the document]
bikallem: Done.

lib_eio/net.ml
```diff
@@ -267,3 +269,25 @@ let with_tcp_connect ?(timeout=Time.Timeout.none) ~host ~service t f =
   | exception (Exn.Io _ as ex) ->
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "connecting to %S:%s" host service
+
+let run_server ?(max_connections=Int.max_int) ?(additional_domains) ~sw ~on_error listening_socket connection_handler =
+  if max_connections < 0 then invalid_arg "max_connections";
```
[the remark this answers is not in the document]
bikallem: Done.

bikallem:
> I see you scaled back the tests since last time. 

Yes, I was getting some error while trying to use the mocks. So I scaled it to testing just one client connection for now.

> I'm putting them back (using mocks) and fixing up some bugs that uncovered..

That would be nice indeed.

talex5:
I pushed my suggested changes here: https://github.com/talex5/eio/commits/run_server_multicore

- Added mock tests for single and multi-domain cases, maximum connections, cancellation and graceful shutdown.
- I brought back your `stop` parameter - I agree that it's a nicer way to do it.
- The documentation said `additional_domains` gives the number of *additional* domains, but the code used it as the total. I changed the code to match.
- I moved `Semaphore.acquire` outside of the `protect`. Otherwise failing to acquire it will do a release.
- Graceful shutdown previously only stopped the main domain. Now it stops all of them.
- I left the test for `max_connections` allowing 0. It's not exactly wrong, so I guess we should allow it.

bikallem:
> I pushed my suggested changes here: https://github.com/talex5/eio/commits/run_server_multicore

The changes look good to me, especially the tests are much more illustrative of the functionality of `run_server`.

One nit, I see that the documentation for @raise has been removed but we still raise `Invalid_argument`. Is this (the documentation) not required?

bikallem:
@talex5 I have cherry-picked your commit to this PR and re-added the @raise documentation. 

> I left the test for max_connections allowing 0. It's not exactly wrong, so I guess we should allow it.

`max_connections` value at <=0 raises `Invalid_argument` as you suggested here (https://github.com/ocaml-multicore/eio/pull/408#discussion_r1086454269).

bikallem:
As an aside, Interestingly, `Eio.Semaphore.make` allows `n=0`. However, in such cases, how does `Eio.Semaphore.acuire s` work? It should never return right?

I just tried below in utop:
```
─( 13:55:02 )─< command 0 >─────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # let m = Eio.Semaphore.make 0;;
val m : Eio.Semaphore.t = <abstr>
─( 13:55:02 )─< command 1 >─────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Eio.Semaphore.acquire;;
- : Eio.Semaphore.t -> unit = <fun>
─( 13:55:14 )─< command 2 >─────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Eio.Semaphore.acquire m;;
Exception: Stdlib.Effect.Unhandled(Eio__core__Suspend.Suspend(_))
```

bikallem:
Yes, it does hang. See below the correct utop session below:

```
utop # Eio_main.run @@ fun env ->
let m = Eio.Semaphore.make 0 in
Eio.Semaphore.acquire m;;
....
```
The `....` bit denotes the `Eio.Semaphore.acquire m` never returns. Should `Eio.Semaphore.make n` require `n > 0`?

bikallem:
In order to not derail this PR and track the issue separately, I have opened an issue for Semaphore https://github.com/ocaml-multicore/eio/issues/415.

lib_eio/net.ml
```diff
@@ -270,24 +270,24 @@ let with_tcp_connect ?(timeout=Time.Timeout.none) ~host ~service t f =
...
-  Switch.run (fun sw -> loop sw)
+  (match additional_domains with
+   | Some (domain_mgr, domains) ->
+     if domains < 0 then invalid_arg "additional_domains";
+     for _ = 2 to domains do
+       Fiber.fork ~sw @@ fun () ->
+       Domain_manager.run domain_mgr (fun () -> Switch.run @@ fun sw -> accept sw);
+     done;
+   | None -> ());
+  Switch.run @@ fun sw -> accept sw
```
polytypic: Oh!  I missed this during my review.

So, IIUC, this now creates a new switch that is different from the switch passed in.

I wonder whether there might be ways to make these kinds of programming errors with switches more difficult to make?

polytypic: Would it make sense to somehow explicitly pass some (parent) context for a switch?  @talex5

talex5: This was due to my (bad) suggestion in https://github.com/ocaml-multicore/eio/pull/408#discussion_r1083843017. In [my branch](https://github.com/talex5/eio/commits/run_server_multicore) I instead replaced this with a single switch and a promise (as @bikallem had it originally), which avoids the confusion.

polytypic:
> * I brought back your `stop` parameter - I agree that it's a nicer way to do it.

@talex5 Hmm... I kind of find this unfortunate as being able to generally rely on cancellation sounds nice in principle.  At any rate, it might be nice to have some "standard" way for servers to allow them to be stopped.  One might have applications running multiple servers for different purposes/types of connections.  So, some sort of standard shutdown mechanism could be nice.

lib_eio/net.ml
polytypic: Well, I was basically suggesting the same (use cancellation) in one of my comments.

bikallem:
>I often don't document Invalid_argument exceptions, as we don't (necessarily) want to promise to raise them, and we don't expect people to catch them. Usually I prefer to say that you have to ensure domains >= 0 when calling it, for example (but it seemed obvious in this case). But there's also no harm in documenting this here I think.

@talex5 Yes, same here. IIUC `Invalid_argument` usually (mostly?) means programmer error rather than a runtime error that user should handle. However, stdlib seems to document it quite a bit (https://github.com/ocaml/ocaml/blob/trunk/stdlib/string.mli#L88) so I thought it was a good practice to follow. For future reference, do you prefer this not being documented? I don't mind either but want to be consistent.

talex5:
>> I brought back your stop parameter - I agree that it's a nicer way to do it.
>
> @talex5 Hmm... I kind of find this unfortunate as being able to generally rely on cancellation sounds nice in principle.

You can rely on cancellation to stop the server and all connections. That is useful if something failed and you want to stop everything and report the error. But graceful shutdown seems a bit different, as it's not an error. So you have to take the shutdown request, create an exception for it, use that to cancel, then catch the exception again and continue.

Everything needs to support cancellation by default because a crash might happen anywhere in a program and you don't want to be stuck waiting for an uncancellable operation before you can report the error. But graceful shutdown is something you need to support explicitly anyway.

> At any rate, it might be nice to have some "standard" way for servers to allow them to be stopped.

I expect this `?stop` argument will be the standard way. It's already used in some other projects, e.g. [Dream.run](https://github.com/aantron/dream/blob/2386083170766255021171870b8568d1c8c933b5/src/dream.mli#L2196) has `?stop:unit promise`. I used `'a Promise.t` though to make it clear that the server *only* finishes by stopping, and if you don't pass `?stop` then it never returns.

bikallem:
Thanks.

Issue #400, opened 2023-01-05
luv fails to handle signals promptly


talex5:
OCaml signal handlers can't run at any time (unlike C ones). So when an OCaml signal occurs, OCaml just records the fact and runs them at the next safe point. Typically the system call that was running will return with EINTR, the C stub will call `caml_leave_blocking_section`, and the signal handler will then run.

This works fine with ocaml-uring. However, libuv automatically retries on EINTR, without giving OCaml a chance to do anything.

This is really a problem with luv, but we might want to make some kind of work-around. libuv does provide its own signal handling system (using handlers written in C to wake the libuv loop).

Reacquiring the runtime lock briefly in e.g. libuv's epoll.c fixes it. However, that would prevent using the system version of libuv.

```diff
diff --git a/src/unix/epoll.c b/src/unix/epoll.c
index 97348e25..cc272d64 100644
--- a/src/unix/epoll.c
+++ b/src/unix/epoll.c
@@ -24,6 +24,9 @@
 #include <errno.h>
 #include <sys/epoll.h>
 
+void caml_enter_blocking_section (void);
+void caml_leave_blocking_section (void);
+
 int uv__epoll_init(uv_loop_t* loop) {
   int fd;
   fd = epoll_create1(O_CLOEXEC);
@@ -288,6 +291,9 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
         reset_timeout = 0;
       }
 
+      caml_leave_blocking_section();
+      caml_enter_blocking_section();
+
       if (timeout == -1)
         continue;
 
```

haesbaert:
It's a shame that it seems not even `Luv.run ~mode:ONCE` could do it, it seems the only thing that unwinds the epoll loop is getting a timeout or an event.

haesbaert:
At any rate, if you choose to go this path, it would make more sense to make the loop unwind so that we end up processing the signal up there, and not run ocaml code while luv is in the middle of an epoll loop.

talex5:
After discussion with @haesbaert, we could mask events in luv threads and run a separate domain to handle signals under luv. Here's a proof-of-concept:
```ocaml
open Eio.Std

let i = ref 3

let interrupted = Eio.Condition.create ()

let handle (_:int) =
  Eio.Condition.broadcast interrupted

let _ =
  Domain.spawn (fun () -> while true do Unix.sleep 10 done)

let () =
  Eio_main.run @@ fun _env ->
[...]

PR #412, opened 2023-01-23
Fix luv signals (issue #400)

haesbaert:
This makes sure we can process signals in luv the same way we do in uring. As stated in #400 the main issue is that luv's mainloop will restart on EINTR and we will never unwind back to ocaml land, so even though the process got the signal, the runtime will not see it until something else goes on.

The trick here is to abuse POSIX thread semantics and shove all signals into one specific systhread by blocking them in the other threads.

Additionally, I've fixed a bug in Ocaml 5.0 where the systhreads end up starting with all signals blocked: https://github.com/ocaml/ocaml/pull/11880. This PR works even with/without the bug.

### Danger Zone
This is tricky ! If you're thinking we can do better than pipes, think again ! Unix.sigsuspend doesn't work on multithreaded, we don't have a real Unix.pause (it's implemented on top of sigsuspend !).
The semantics for kill(2) to "self" are different than "from outside" when multithreaded, and the runtime doesn't expose pthread_kill(2). That's why on the test we have to fork and signal back to the parent. I know, it's horrible.

### DIFF
--- a/lib_eio_luv/eio_luv.ml
+++ b/lib_eio_luv/eio_luv.ml
@@ -1186,7 +1186,7 @@ let rec wakeup ~async ~io_queued run_q =
       Luv.Async.send async |> or_raise
   | None -> ()
 
-let rec run : type a. (_ -> a) -> a = fun main ->
+let rec run2 : type a. (_ -> a) -> a = fun main ->
   let loop = Luv.Loop.init () |> or_raise in
   let run_q = Lf_queue.create () in
   let io_queued = ref false in
@@ -1198,7 +1198,7 @@ let rec run : type a. (_ -> a) -> a = fun main ->
         Luv.Loop.stop loop
     ) |> or_raise in
   let st = { loop; async; run_q; fd_map = Fd_map.empty } in
-  let stdenv = stdenv ~run_event_loop:run in
+  let stdenv = stdenv ~run_event_loop:run2 in
   let rec fork ~new_fiber:fiber fn =
     Ctf.note_switch (Fiber_context.tid fiber);
     let open Effect.Deep in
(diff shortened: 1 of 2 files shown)


talex5 (commented):
Do we really want a thread and not a domain? As I understand it, a thread is attached to a domain and can't run while another thread in that domain is running. So if domain 0 is busy doing CPU stuff then no other domains will be able to get signals I think.

haesbaert:
> Do we really want a thread and not a domain? As I understand it, a thread is attached to a domain and can't run while another thread in that domain is running. So if domain 0 is busy doing CPU stuff then no other domains will be able to get signals I think.

I wouldn't really worry about it being busy, signals are already delayed and not supposed to be realtime.

Having it as a Domain would break the consecutive id's if people want to store things per-domain somehow, as the first Domain spawned by EIO would be id 2 and not 1.

Another issue is that having one an extra idle Domain, even if completely idle, can slow things considerably (see https://github.com/ocaml-multicore/domainslib/issues/77#issuecomment-1291676255)  since everyone has to take part in GC, having more Domains are a nono for now.

A third and minor point would be keeping things similar to other backends, the systhread is a hack/deviation of the luv backend, but the things we do in a Domain remain the same in luv/uring/whatever, as in they do heavy bulk.

talex5:
Thinking about this more, I don't think the busy CPU thing will prevent the signal being handled, because OCaml already has safe points to handle this in the single-domain case.

talex5:
After fixing the compile error, `make test_luv` failed:
```
rm -rf _build
EIO_BACKEND=luv dune runtest
File "tests/dune", line 1, characters 0-54:
1 | (mdx
2 |   (package eio_main)
3 |   (deps (package eio_main)))
Command got signal INT.
```
Are we sure that waitpid will only return *after* the signal is delivered? Also, I don't think you can use a ref cell here. It needs to be `Atomic.t`.

I pushed a commit that fixes it for me. Does that look right to you?

haesbaert:
> After fixing the compile error, `make test_luv` failed:
> 
> ```
> rm -rf _build
> EIO_BACKEND=luv dune runtest
> File "tests/dune", line 1, characters 0-54:
> 1 | (mdx
> 2 |   (package eio_main)
> 3 |   (deps (package eio_main)))
> Command got signal INT.
> ```
> 
> Are we sure that waitpid will only return _after_ the signal is delivered? Also, I don't think you can use a ref cell here. It needs to be `Atomic.t`.
> 

Good point ! Not at all, the signal can arrive, after or before, I did not consider that.
I must have been (un)lucky, I ran make test_luv in a loop for an hour and never triggered it.

Yeah, since the waitpid doesn't hold, a ref cell only would be racy.

> I pushed a commit that fixes it for me. Does that look right to you?

Yep :+1:

PR #413, opened 2023-01-24
Add cancellable lock-free synchronous channel

talex5:
This is used when you ask for an Eio.Stream with capacity 0. It's slightly slower in the single-domain case, but much faster with multiple domains (running `bench_stream.ml`):

sync

I also tried adding a bit of work to the benchmark (`for _ = 1 to 1000 do () done;` in `run_recv`). Then multiple domains are actually faster than one domain, but the benefits are still present:

sync-work

I originally planned to write a stream replacement that worked for all capacities, but making a single data structure work for both zero and non-zero capacities turned out to be difficult, especially with non-blocking take.

For a non-blocking take on a non-zero capacity stream you have the advantage that you can reserve one of the items already in the stream, or fail if there isn't one. But with a zero capacity stream a non-blocking take always involves both parties. But zero capacity streams also offer simplifications.

The algorithm is described in detail in the comment at the top of `sync.ml`. `lib_eio/tests/sync.md` contains a walk-through of some (non-racing) cases, showing the internal state of the stream at each step. `lib_eio/tests/dscheck/test_sync.ml` tests all possible interleavings of atomic operations.

It should be fairly easy to extend this to provide a take operation that takes from exactly one of several streams. This is hard to do with non-0-capacity streams because while one consumer is considering whether to accept an item, all other consumers have to wait, so that the items are processed in order.

I also simplified the old locking version in `stream.ml` as it no longer needs special cases for `capacity=0`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ test_luv:
 	EIO_BACKEND=luv dune runtest
 
 dscheck:
+	dune exec -- ./lib_eio/tests/dscheck/test_sync.exe
 	dune exec -- ./lib_eio/tests/dscheck/test_semaphore.exe
 	dune exec -- ./lib_eio/tests/dscheck/test_cells.exe
 
--- a/lib_eio/sync.ml
+++ b/lib_eio/sync.ml
@@ -1,0 +1,448 @@
+(* A lock-free synchronous channel with cancellation, using Cells.
+
+   Producers and consumers are paired off and then the producer transfers its
+   value to the consumer. This is effectively a bounded queue with a capacity
+   of zero.
+
+   Both producers and consumers can cancel while waiting.
+
+   There is an atomic int ([balance]), plus two queues ([consumers] and
+   [producers]) made using Cells. When [balance] is positive, it is the number
+   of producers waiting with values that no one is yet responsible for
+   resuming. When negative, it is the (negative) number of waiting consumers
+   that no one is responsible for resuming.
+
+   To put an item:
+
+   1. The producer increments [balance].
+   2. If it was negative, the producer resumes one waiting consumer on the [consumers] queue.
+      Otherwise, it suspends itself on the [producers] queue.
+
+   To take an item:
+
+   1. The consumer decrements [balance].
+   2. If it was positive, the consumer resumes one waiting producer on the [producers] queue.
+      Otherwise, it suspends itself on the [consumers] queue.
+
+   Therefore, we never try to resume on a queue unless another party has
+   started the process of suspending on it.
+
+   The system will not become idle while a client is responsible for resuming
+   something. Therefore, when idle:
+
+   - If [balance <= 0] then there are no waiting producers.
+   - If [balance >= 0] then there are no waiting consumers.
+   - So, we never have waiting consumers and producers at the same time.
+
+   As usual with Cells, either party may get to the new cell first. Whichever party
+   arrives first writes a callback, which the other party will then call when they arrive.
+
+   Note on terminology:
+
+   - The "suspender" of a cell is the party that incremented the queue's suspend index,
+     and the "resumer" of a cell is the party that incremented the resume index.
+
+   - Whether "suspending" or "resuming" a cell, you may still have to suspend
+     your fiber and resume it later.
+
+   States
+
+   There are four cell states:
+
+   - [In_transition] indicates that the cell is still being initialised, or might be
+     getting cancelled. Either way, the suspending party is actively working to
+     change the cell's state.
+
+   - [Item] indicates that the producer is ready to provide an item.
+
+   - [Slot] indicates that the consumer is ready to receive an item.
+
+   - [Finished] indicates that the cell is no longer being used (the value has
+     been consumed or the cell has finished being cancelled).
+
+   The possible sequences of states on the [producers] queue are:
+
+   In_transition -C> Slot -P> Finished    (consumer arrives first)
+                 `P> Item -C> Finished    (producer arrives first)
+                          `P> In_transition -P> Finished   (producer cancels)
+                                            `C> Slot -P> Finished   (cancellation interrupted)
+
+   Only the producer can cancel here. For the [consumers] queue it's the
+   opposite - the consumer can cancel its [Slot].
+
+   Cancellation
+
+   Note that there are two kinds of cancellation here:
+
+   1. A cancelled cell is not considered part of its queue. Anyone seeing one
+      (due to a race) will skip over it and use the next cell.
+
+   2. After a consumer and producer have been paired off (and the cell removed
+      from its queue), the consumer callback may reject the value. If this
+      happens, the producer must start all over again to find another consumer.
+
+   Whenever a consumer sets its callback to reject values, it should then start
+   the process of cancelling its cell (if acting as a suspender) so that the
+   cell can be GC'd.
+
+   A consumer can only cancel its cell when it's on the [consumers] queue.
+   If it's on [producers], it knows a wake up will be coming shortly anyway.
+   A consumer cancels its cell as follows:
+
+   1. The consumer sets its cell in [consumers] to [In_transition].
+   2. It increments [balance] (from a negative value). It is now committed to cancelling.
+   3. It sets its cell to [Finished].
+
+   (1) will fail if the cell got resumed first. In that case the consumer just
+   rejects the cancellation attempt.
+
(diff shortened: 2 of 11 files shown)


polytypic:
BTW, without looking in detail, this reminds me of [Parallel Concurrent ML](http://manticore.cs.uchicago.edu/papers/icfp09-parallel-cml.pdf).

Makefile
```diff
@@ -19,6 +19,7 @@ test_luv:
 	EIO_BACKEND=luv dune runtest
 
 dscheck:
+	dune exec -- ./lib_eio/tests/dscheck/test_sync.exe
```
polytypic, line 22: If I run `make dschek` on my mac, it only seems to run the first of these tests:

```sh
➜  eio git:(sync) ✗ make dscheck      
dune exec -- ./lib_eio/tests/dscheck/test_sync.exe
Done: 40% (86/215, 129 left) (jobs: 0)run: 1000
run: 2000
... snip lots of similar lines ...
run: 28880000
rumake: *** [dscheck] Killed: 9
➜  eio git:(sync) ✗ 
```

polytypic: On the same note, running `make bench` fails:

```sh
dune exec -- ./lib_eio_linux/tests/bench_noop.exe
Error: Program "./lib_eio_linux/tests/bench_noop.exe" not found!
make: *** [bench] Error 1         
```

polytypic: Well, the same also goes for `dune build`:

```sh
➜  eio git:(sync) ✗ dune build
File "lib_eio_linux/tests/dune", line 30, characters 21-30:
30 |  (libraries alcotest eio_linux))
                          ^^^^^^^^^
Error: Library "eio_linux" in _build/default/lib_eio_linux is hidden
(unsatisfied 'enabled_if').
-> required by _build/default/lib_eio_linux/tests/test.exe
-> required by alias lib_eio_linux/tests/all
-> required by alias default
Error: The package eio_linux does not have any user defined stanzas attached
to it. If this is intentional, add (allow_empty) to the package definition in
the dune-project file
-> required by _build/default/eio_linux.install
-> required by alias all
-> required by alias default
```

It would be nice that the build wouldn't appear to fail on mac.  I don't know how to fix this immediately, but I can take a look after reviewing this PR.

talex5: `make dscheck` currently requires a patched version of dscheck, as the official version is too slow to be useful: https://github.com/ocaml-multicore/dscheck/pull/3. However, that version has a bug which means it may also miss some cases.

Ideally, we'd replace the explicit commands in `make bench` and `make dscheck` with e.g. `dune build @dscheck`. Then dune could be smarter about filtering things out on mac. But then you have to stop it running things in parallel, disable buffered output, and force it to let you run the benchmarks multiple times without changes to the code, which is a bit of a pain.

Fixing the main build on mac is important, though! I don't know why dune is trying to build it though. We already have:
```
(test
 (name test)
 (package eio_linux)
 (enabled_if (= %{system} "linux"))
 (modules test)
 (libraries alcotest eio_linux))
```

lib_eio/sync.ml
```diff
@@ -0,0 +1,448 @@
...
+  match Atomic.get cell, value with
+  | Finished, _ -> add_to_cell queue value (Q.next_resume queue)     (* Cancelled - skip *)
+  | (Slot kc   as old), Item item
+  | (Item item as old), Slot kc ->
+    if Atomic.compare_and_set cell old Finished then exchange item kc
+    else add_to_cell queue value cell
+  | In_transition, _ ->
+    if Atomic.compare_and_set cell In_transition value then ()
+    else add_to_cell queue value cell
+  | (Slot _ | Item _), _ -> assert false
```
polytypic, line 208: Just a thought: You could probably rule this case out by using either GADTs or polymorphic variants and make the two queues to be of different type.  I'm not sure it is worth the effort, but it should be possible.

lib_eio/sync.ml
```diff
@@ -0,0 +1,448 @@
...
+
+(* Cancelling *)
+
+let rec decr_balance_if_positive t =
+  let cur = Atomic.get t.balance in
+  if cur > 0 then (
+    if Atomic.compare_and_set t.balance cur (cur - 1) then true
+    else decr_balance_if_positive t
+  ) else false
```
polytypic, line 217: You could also use a combination of `&&` and `||` here and [avoid having your coolness factor drop](https://twitter.com/andrejbauer/status/1296728568706326535). 😄

Seriously speaking this is probably clearer with an `if then else`.

lib_eio/sync.ml
```diff
@@ -0,0 +1,448 @@
...
+          | _, Short _ ->
+            (* We can't cancel while in the process of resuming a cell on the [consumers] queue.
+               We could set [cancel] to [`Resuming] here, but there's no need as trying to use the
+               old cancel function will find the old cell is cancelled and set [request.cancel]
+               to [`Cancelled]), as required. *)
+            ()
+        in aux ()
+      )
+
+(* We tried to [put] and no value was immediately available.
```
polytypic, line 337: I guess this should say "no ~value~slot was immediately available".

talex5: Yes, that would be clearer. Note that I am using "value" to mean "item or slot" in some places (e.g. in `add_to_cell`).

lib_eio/sync.ml
talex5: That doesn't work - you can get either type on either queue (it just depends who writes the cell first). e.g. on the `consumers` queue usually the consumer gets there first and writes a `Slot`, sometimes the producer gets there first and writes an `Item` instead.

polytypic:
I haven't found a smoking gun yet and the basic approach seems like it should work.  I still feel I want to stare at the code a little bit more to understand exactly how things work, but I need to take a break now.  I'll try to finish my review by tomorrow morning.

polytypic:
Alright.  I believe I'm now reasonably satisfied that this should work. So, I'm accepting this.

However, I'm getting a strong feeling now that there should be a more modular way to handle cancellation.  In the semaphore implementation cancellation didn't require that much logic.  In this channel implementation there is quite a bit of logic related to cancellation, which worries me.  Also, I don't believe the cancellation logic is strong enough to support selective operations such as trying to acquire at most one of two different semaphores or more generally, trying to do at most one of two different cancellable operations.  Being able to do such things is what languages like Concurrent ML, Go, and some others allow you to do via selective communication.  It is something I've had in mind to explore (in my `par-ml` project: "TODO: Composable synchronization primitives").  The Parallel Concurrent ML paper I linked to presents an approach to that and I've implemented it previously in F#.  However, I feel that it is not entirely satisfactory as it is not completely lock-free and I believe it is possible to better &mdash; perhaps the Cells abstraction is a key to that.

polytypic (approved):
🚀 🌔

talex5:
Yes, I want to support that kind of thing in future (thanks for the PCML paper link!). There's also the [Reagents](https://github.com/ocaml-multicore/reagents) work that e.g. @bartoszmodelski and @kayceesrk are working on.

I think the only change required here is to provide a version of `Sync.take` that lets the user provide `kc` (the consumer callback, which decides whether to accept a value or reject it).

On its own, that should allow us to provide a `Sync.take_choose` operation, to remove an item from exactly one of several streams atomically, without any overhead with the other operations. The `kc` function would CAS an atomic (shared between all the streams being selected over) before accepting the value, to ensure it accepted only one.

For the full CML API (allowing e.g. *pushing* to one of multiple streams), I think it could be built on top of that, by defining e.g.

```ocaml
type 'a t = 'a state Atomic.t Sync.t
```

So the provider provides its own atomic value and the consumer marks that and its own atomic before accepting. But that will come with a little overhead (not sure how much) so we might not want to do it in all cases. Not having used CML, it's not clear to me how useful the full system is vs just having `take_choose`.

Whether we provide CML-like features everywhere or as an additional layer, I think that sync.ml itself won't need to change much.

Makefile
polytypic: A related issue in dune: [The documentation/value of %{system} is not consistent](https://github.com/ocaml/dune/issues/4895).

I also tried to work around this by using the `(* -*- tuareg -*- *)` magic, but that didn't work either as that hits a limitation in dune: [(dirs ...) not recognised in dune2](https://github.com/ocaml/dune/issues/2938).

talex5: I've reported it at https://github.com/ocaml/dune/issues/6938

PR #414, opened 2023-01-24
eio-luv: improve `process.md` test

smondet:
The function `read_all` in the example seems to make a wrong assumption about
the behavior of `Eio_luv.Low_level.Stream.read_into`.

The test worked because the output of the process tested is smaller than the
buffer-size so the `read` would never loop.

With more data we see that the test fails:

```patch
diff --git a/lib_eio_luv/tests/process.md b/lib_eio_luv/tests/process.md
index 46aae4ac..3085a054 100644
--- a/lib_eio_luv/tests/process.md
+++ b/lib_eio_luv/tests/process.md
@@ -43,7 +43,9 @@ Using a pipe to redirect output to a buffer.
   let redirect = Eio_luv.Low_level.Process.[
     to_parent_pipe ~fd:Luv.Process.stdout ~parent_pipe ()
   ] in
-  let t = Process.spawn ~sw ~redirect "echo" [ "echo"; "Hello,"; "World!" ] in
+  let t = Process.spawn ~sw ~redirect "echo" [ "echo"; 
+  "0123456789012345678901234567890123456789";
+  "Hello,"; "World!" ] in
   let read = read_all parent_pipe buf in
   let _ = Process.await_exit t in
   Luv.Buffer.to_string (Luv.Buffer.sub buf ~offset:0 ~length:read);;
```

with an exception

```default
 $ dune build @lib_eio_luv/tests/runtest
File "lib_eio_luv/tests/process.md", line 1, characters 0-0:
diff --git a/_build/default/lib_eio_luv/tests/process.md b/_build/default/lib_eio_luv/tests/.mdx/process.md.corrected
index 3085a054..01a92389 100644
--- a/_build/default/lib_eio_luv/tests/process.md
+++ b/_build/default/lib_eio_luv/tests/.mdx/process.md.corrected
@@ -49,7 +49,7 @@ Using a pipe to redirect output to a buffer.
   let read = read_all parent_pipe buf in
   let _ = Process.await_exit t in
   Luv.Buffer.to_string (Luv.Buffer.sub buf ~offset:0 ~length:read);;
-- : string = "Hello, World!\n"
+Exception: Invalid_argument "Bigarray.sub: bad sub-array".
```
 
This patch just fixes the `read_all` function for the lazy users (like me) who
just copy code from the tests :)

### DIFF
--- a/lib_eio_luv/tests/process.md
+++ b/lib_eio_luv/tests/process.md
@@ -11,12 +11,17 @@ module Process = Eio_luv.Low_level.Process
 A helper function for reading all of the bytes from a handle.
 
 ```ocaml
-let read_all handle buf =
-  let rec read acc =
+let read_all handle =
+  let buf = Luv.Buffer.create 32 in 
+  let acc_buffer = Buffer.create 42 in
+  let rec read () =
     match Eio_luv.Low_level.Stream.read_into handle buf with
-    | i -> read (acc + i)
-    | exception End_of_file -> acc
-  in read 0
+    | i ->
+      Buffer.add_string acc_buffer
+        (Luv.Buffer.to_string (Luv.Buffer.sub buf ~offset:0 ~length:i));
+      read ()
+    | exception End_of_file -> Buffer.contents acc_buffer
+  in read ()
 ```
 
 A simple `echo hello` process redirects to stdout.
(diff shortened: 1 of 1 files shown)


patricoferris (approved):
Thanks!

(I think that's what @talex5 meant here and I only switched the order ^^" https://github.com/ocaml-multicore/eio/pull/359#discussion_r1015501954)

talex5 (approved):
Thanks - that function was misleading indeed!

I hope we can get process support in the main Eio API soon, so people won't need to copy low-level backend-specific code.

> (I think that's what @talex5 meant here and I only switched the order ^^" https://github.com/ocaml-multicore/eio/pull/359#discussion_r1015501954)

No, that was a different problem - I hadn't spotted that `read_all` was broken :-(

patricoferris:
> I hope we can get process support in the main Eio API soon, so people won't need to copy low-level backend-specific code.

I have some working linux stubs somewhere that I worked on after splitting #330, I'll see if I can get that up to date and open a PR soon(ish).

PR #417, opened 2023-01-26
Update network examples to use run_server

talex5:
/cc @bikallem

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ opam install eio_main utop
 Try out the examples interactively by running `utop` in the shell.
 
 First `require` the `eio_main` library. It's also convenient to open the [Eio.Std][]
-module, as follows. (The leftmost `#` shown below is the Utop prompt, so enter the text after the 
+module, as follows. (The leftmost `#` shown below is the Utop prompt, so enter the text after the
 prompt and return after each line.)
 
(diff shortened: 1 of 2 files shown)


bikallem:
LGTM.

talex5:
It didn't quite work reliably. The client ends the test when it finishes its part, but since it only sent to the server and didn't wait for a response there was a race. I've rewritten the example so that the server sends and the client reads instead.

bikallem:
> It didn't quite work reliably. The client ends the test when it finishes its part, but since it only sent to the server and didn't wait for a response there was a race. I've rewritten the example so that the server sends and the client reads instead.

Is this only when both client and server are running in the same process (exe)? I created a simple test where client and server were separate processes and they seemed to work okay - even when the client was initiating the sending of the data.

talex5:
The bad sequence is:

1. We start the server.
2. The client sends a message to it.
3. The client is finished, so it stops the test, shutting down the server before the server got the message.

Sometimes the server will get the message before it gets shut down and sometimes it won't. It wasn't a problem before because before the server only handled one message and then stopped itself, and so we could just wait for it to finish. But now it uses `run_server` which accepts multiple connections, so doing it there is messy.

PR #418, opened 2023-01-27
Buf_write: Fix BE.uint48 and LE.uint48.

adatario:
Arguments to logical shift are provided in bits, not bytes.

No tests added yet. It would be nice to test the `Buf_write` <-> `Buf_read` roundtrip. I'd be happy to add them, but am not sure where the best place would be.

### DIFF
--- a/lib_eio/buf_read.ml
+++ b/lib_eio/buf_read.ml
@@ -241,6 +241,8 @@ let any_char t =
   consume t 1;
   c
 
+let uint8 t = Char.code (any_char t)
+
 let peek_char t =
   match ensure t 1 with
   | () -> Some (get t 0)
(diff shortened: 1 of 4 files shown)


talex5:
`tests/buf_write.md` would be the best place for some tests. There are already tests for the reading side:

https://github.com/ocaml-multicore/eio/blob/80f5352526b260e9c32d7110844c7e7408e905f4/tests/buf_reader.md?plain=1#L418-L419

Note: this code is from [Faraday](https://github.com/inhabitedtype/faraday/blob/c66fc020fdc2327556c36fc1e912524b427d7da8/lib/faraday.ml#L290-L297), so we should probably report it there too.

Thanks!

talex5:
I added some round-trip tests, and also fixed a second bug in `BE.uint48` :scream:

I also added `Eio.Buf_read.uint8`, which was missing.

adatario:
> I added some round-trip tests

Cool, thanks!

> and also fixed a second bug in BE.uint48 :scream:

:) Interesting that this was never encountered in Faraday.

Issue #125, opened 2022-01-03
Test on Windows


talex5:
Eio needs to support Windows. It should mostly work already, using the eio_luv backend, but there are probably some things that need fixing (e.g. #124).

- [ ] Test it on Windows and find out what's broken.
- [ ] Set up CI to run tests on Windows.

nojb:
Hiya, just FYI I am beginning to try this on Windows. I managed to build the library with the `mingw64` compiler and the `libuv` backend (not completely trivial due to the many dependencies :)). A trivial test program
```
let () =
  Eio_luv.run @@ fun env ->
  let clock = Eio.Stdenv.clock env in
  Eio.traceln "The time is now %f" (Eio.Time.now clock)
```
fails with 
```
Fatal error: exception Invalid_argument("Sys.signal: unavailable signal")
```
Fix is probably easy (just ignore signals?) but anyway, I don't think you should run to fix this one issue; I am planning to experiment a bit more and will report back (adding to this issue).

talex5:
Ah, great! Yes, it's probably just needs this disabled:

https://github.com/ocaml-multicore/eio/blob/8beda18c9950aa8f0230b9f699d47b8d7bfe214a/lib_eio_luv/eio_luv.ml#L28

(`OCAMLRUNPARAM=b` should give the location if not)

nojb:
BTW what do you use for testing? The files in `test/`?

talex5:
Yes. `dune runtest` should run all the tests.

`make bench` runs the benchmarks.

nojb:
> Yes. `dune runtest` should run all the tests.

Unfortunately they rely on `mdx` and its status on Windows is not very clear (see realworldocaml/mdx#295). I will try to run them manually.

talex5:
Ah, is this because of different line-endings? It should be possible to configure Git to check the files out with Unix line endings, for testing. It looks like we're not using any shell blocks, so that should be OK.

polytypic:
I'm looking into getting Eio working on Windows.  I was just able to reproduce the signal issue:

```
C:\...\eio>make bench
dune exec -- ./bench/bench_buf_read.exe
+Read 100000000 bytes in 0.875s
dune exec -- ./bench/bench_mutex.exe
Fatal error: exception Invalid_argument("Sys.signal: unavailable signal")
make: *** [Makefile:8: bench] Error 2
```

polytypic:
[...]

PR #420, opened 2023-01-27
Windows: fix "unavailable signal" error

talex5:
Only try to ignore SIGPIPE signals on Unix-type systems. Otherwise, it fails with:

    Fatal error: exception Invalid_argument("Sys.signal: unavailable signal")

Also, delay doing this until `Eio_luv.run` is called, so that merely linking with `Eio_luv` doesn't change the behaviour of your program.

Reported-by: Nicolás Ojeda Bär in https://github.com/ocaml-multicore/eio/issues/125#issuecomment-1230171255.

@polytypic can you confirm this works on Windows? I assume it's still needed.

### DIFF
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -26,9 +26,6 @@ module Suspended = Eio_utils.Suspended
 module Zzz = Eio_utils.Zzz
 module Lf_queue = Eio_utils.Lf_queue
 
-(* SIGPIPE makes no sense in a modern application. *)
-let () = Sys.(set_signal sigpipe Signal_ignore)
-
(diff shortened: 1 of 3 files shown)


polytypic (approved):
Yes, this should work and is needed on Windows.  (I didn't try to run the code, though.)

talex5:
I modified eio_linux to delay setting the handler until `run` is called too, and added a test.

Note: The behaviour when the default handler *is* present is a bit odd: the write fails with EPIPE but when the process eventually decides to quit it fails with SIGPIPE. I suspect this is due to #412. @haesbaert mentioned:

> The semantics for kill(2) to "self" are different than "from outside" when multithreaded

It doesn't seem like a problem, anyway.

PR #421, opened 2023-01-27
eio_luv: fix some resource leaks

### LINKED ISSUE #419: FD leak in Eio_luv.run
This program fails with `EMFILE` after a bit:
```ocaml
let () =
  while true do
    Eio_luv.run ignore
  done
```
I suspect it's not closing async and the loop in the right order.

talex5:
Previously, we asked to close the `async` notifier after the loop had stopped. Presumably because the loop wasn't running, this close operation never ran, which leaked some FDs. Also, the `Loop.close` it was supposed to call after closing itself never got called either.

Once that was fixed, `Loop.close` reported further resource leaks in the Poll and Process modules, which this commit also fixes.

/cc @patricoferris

Fixes #419.

### DIFF
--- a/lib_eio_luv/eio_luv.ml
+++ b/lib_eio_luv/eio_luv.ml
@@ -192,7 +192,8 @@ end = struct
     let m = if Lwt_dllist.is_empty events.read then m else `READABLE :: m in
     if m = [] then (
       Luv.Poll.stop events.handle |> or_raise;
-      t.fd_map <- Fd_map.remove events.fd t.fd_map
+      t.fd_map <- Fd_map.remove events.fd t.fd_map;
+      Luv.Handle.close events.handle (fun () -> ())
     ) else (
       Luv.Poll.start events.handle m (poll_callback t events)
     )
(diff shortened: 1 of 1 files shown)


talex5:
> The changes look good to me, I ran it on macOS which uncovered a few more leaks (it's probably worth adding some macOS GitHub action CI).

Thanks - I applied your fix and fixed domain_mgr too. Agree it would be good to add a macOS CI runner.

patricoferris (approved):
All working for me :))

lib_eio_luv/eio_luv.ml
```diff
@@ -192,7 +192,8 @@ end = struct
     let m = if Lwt_dllist.is_empty events.read then m else `READABLE :: m in
     if m = [] then (
       Luv.Poll.stop events.handle |> or_raise;
-      t.fd_map <- Fd_map.remove events.fd t.fd_map
+      t.fd_map <- Fd_map.remove events.fd t.fd_map;
+      Luv.Handle.close events.handle (fun () -> ())
```
polytypic, line 196: BTW, I just tested with [godbolt](https://ocaml.godbolt.org/) that `Fun.id` instead of `fun () -> ()` or `ignore` generates one instruction less.  `Fun.id` is a function that just returns.  `fun () -> ()` and `ignore` are functions that return the unit value (and that takes an additional instruction).  I could imagine a type directed optimization that would eliminate this difference in case `ignore` is called with the unit value and it should also be possible to optimize `fun () -> ()`, but the compiler doesn't seem to do that yet.

I don't suggest to change the code in this PR, but it is something I might want to consider in performance critical cases.

talex5: Ah, interesting! I didn't want to use `ignore` here in case `close` changed in future to pass a result type or something, but `Fun.id` would have worked.

PR #425, opened 2023-01-30
net: add tcp_connect - an alternate version of with_tcp_connect accepting Switch.t argument. 

bikallem:
Part of the fix for https://github.com/mirage/ocaml-cohttp/issues/965. 

After this is merged Client.call should take an (optional? not decided yet) sw argument too.

### DIFF
--- a/lib_eio/net.ml
+++ b/lib_eio/net.ml
@@ -246,8 +246,7 @@ let getnameinfo (t:#t) sockaddr = t#getnameinfo sockaddr
 
 let close = Flow.close
 
-let with_tcp_connect ?(timeout=Time.Timeout.none) ~host ~service t f =
-  Switch.run @@ fun sw ->
+let tcp_connect ?(timeout=Time.Timeout.none) ~sw ~host ~service t =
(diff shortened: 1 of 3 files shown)


talex5 (changes requested):
The point of the `with_*` functions is that you don't need a switch. It would be better to have a separate `tcp_connect` that takes `~sw` and doesn't take `f` if you need that.

bikallem:
@talex5 I have added new function `tcp_connect` as you suggested. This takes switch as an argument so the connection outlives the function call. 

ci seems to be failing due to an error below.
```
2023-01-31 09:30.50: Waiting for resource in pool OCluster

2023-01-31 09:30.50: Waiting for worker…

2023-01-31 09:30.50: Got resource from pool OCluster

Building on cree.ocamllabs.io

Uncaught exception: Unix.Unix_error(Unix.ENOMEM, "fork", "")

2023-01-31 09:30.51: Job failed: Failed: Internal error
```

bikallem:
closing this for now.

PR #426, opened 2023-01-30
Describe secure_random as an infinite source

### LINKED ISSUE #422: Document that `secure_random` isn't quite a `Flow.source`
I accidentally tried to use `secure_random` as a `Flow.source` with `Flow.copy` which of course in retrospect is silly because there's no `End_of_file` to raise and the program just hangs. Maybe others might accidentally do the same and get confused, the best thing I think is maybe to add some documentation to stdenv ?

https://github.com/ocaml-multicore/eio/blob/80f5352526b260e9c32d7110844c7e7408e905f4/lib_eio/eio.mli#L246-L247

patricoferris:
Fixes #422

### DIFF
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -244,7 +244,7 @@ module Stdenv : sig
   (** {1 Randomness} *)
 
   val secure_random : <secure_random : #Flow.source as 'a; ..> -> 'a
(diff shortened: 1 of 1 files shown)


talex5 (approved):
Thanks!

PR #428, opened 2023-01-31
eio_linux: call submit as needed

### LINKED ISSUE #409: eio_linux backend hangs
Consider the following program that uses `Eio.Net.accept_fork` to establish a server on port 8081 and attempts to establish client connections to this server. The number of clients is user configurable via cli option `-clients`. If the number <= 62, then we get one output as follows
```
+Server accepted connection from client
+Server received: "Hello from client"
+client received: "Bye"
```
However, if the number is >= 63, then the program doesn't even print the above and just (hangs?).

```
$ cat test_run_server.ml
open Eio

let addr = `Tcp (Eio.Net.Ipaddr.V4.loopback, 8081)

let read_all flow =
  let b = Buffer.create 100 in
  Eio.Flow.copy flow (Eio.Flow.buffer_sink b);
  Buffer.contents b

let eio_run_server ~clients env sw =
  let run_client id () =
    traceln "client: Connecting to server ...%d" id;
    let flow = Eio.Net.connect ~sw env#net addr in
    Eio.Flow.copy_string "Hello from client" flow;
    Eio.Flow.shutdown flow `Send;
    let msg = read_all flow in
    traceln "client received: %S" msg;
  in
  let connection_handler clock flow _addr =
    traceln "Server accepted connection from client";
    Fun.protect (fun () ->
      let msg = read_all flow in
      traceln "Server received: %S" msg;
      Eio.Time.sleep clock 0.01
    ) ~finally:(fun () -> Eio.Flow.copy_string "Bye" flow)
  in
  let server_sock = Eio.Net.listen ~reuse_addr:true ~backlog:128 ~sw env#net addr in
  let connection_handler = connection_handler env#clock in
  let clients = List.init clients (fun id -> run_client (id+1)) in
  let server () =
    traceln "starting server ..."; 
    Eio.Net.accept_fork ~sw server_sock ~on_error:raise connection_handler 
  in
  Fiber.all (server :: clients)

let () =
  let clients = ref 60 in
  Arg.parse
    [ ("-clients", Arg.Set_int clients, " total clients to spawn")]
    ignore "test Eio.Net.fork_accept()";

  Printexc.record_backtrace true ;
  Eio_main.run @@ fun env ->
  Switch.run @@ fun sw ->
  eio_run_server ~clients:!clients env sw
```

```
$ cat dune
(executable (name test_run_server) (libraries eio eio_main))
```

talex5:
We may fail to submit a job because the SQE queue is full. Previously we would wait until some existing request completed, but that might never happen. Instead, we just flush the SQE queue and retry.

Fixes #409.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -39,7 +39,7 @@
   (logs (>= 0.7.0))
   (fmt (>= 0.8.9))
   (cmdliner (and (>= 1.1.0) :with-test))
-  (uring (>= 0.4))))
+  (uring (>= 0.5))))
 (package
  (name eio_luv)
  (synopsis "Eio implementation using luv (libuv)")
--- a/eio_linux.opam
+++ b/eio_linux.opam
@@ -16,7 +16,7 @@ depends: [
   "logs" {>= "0.7.0"}
   "fmt" {>= "0.8.9"}
   "cmdliner" {>= "1.1.0" & with-test}
-  "uring" {>= "0.4"}
+  "uring" {>= "0.5"}
   "odoc" {with-doc}
 ]
 build: [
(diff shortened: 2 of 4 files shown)


talex5:
The noop benchmark goes a bit faster with this, when you have more operations than queue slots:

noop

PR #435, opened 2023-02-02
Add low-level process support to eio_linux

patricoferris:
The next instalment for #330. This adds initial low-level process support to `Eio_linux` using the [spawn](https://github.com/janestreet/spawn) library which has [no dependencies](https://github.com/janestreet/spawn/blob/a1af0bf74d9b8436094ddb46243c698b6824d656/spawn.opam#L28-L32). I started implementing it and realised they already had, and it supports windows and has a lot of signal and error handling built-in that I otherwise would have ended up copying. It is also released under MIT so presumably we could reuse their code to avoid the dependency.

I put it under `Eio_unix` because I envisaged reusing spawn with other backends like IOCP/kqueue I think that means we have to add `optional` or something to the `eio_unix` library? 

It doesn't do program resolution so I tried to do that in OCaml, I also default to the paths `[ "/usr/bin"; "/usr/local/bin" ]` if we can't find `PATH` -- I don't know if that's reasonable. Also I'm not sure on what error we should raise if we cannot find the program.

### DIFF
--- a/lib_eio/unix/eio_unix.ml
+++ b/lib_eio/unix/eio_unix.ml
@@ -77,3 +77,18 @@ let getnameinfo (sockaddr : Eio.Net.Sockaddr.t) =
   run_in_systhread (fun () ->
     let Unix.{ni_hostname; ni_service} = Unix.getnameinfo sockaddr options in
     (ni_hostname, ni_service))
+
+module Spawn = struct
+  let resolve_program ~paths prog =
+    if not (Filename.is_relative prog) then Some prog
+    else 
+    let rec loop = function
+      | path :: rest ->
+        let p = Filename.concat path prog in
+        if Sys.file_exists p then Some p else loop rest
+      | [] -> None
+    in
+    loop paths
+  
+  let spawn = Spawn.spawn
+end
--- a/lib_eio/unix/eio_unix.ml
+++ b/lib_eio/unix/eio_unix.ml
@@ -77,18 +77,3 @@ let getnameinfo (sockaddr : Eio.Net.Sockaddr.t) =
   run_in_systhread (fun () ->
     let Unix.{ni_hostname; ni_service} = Unix.getnameinfo sockaddr options in
     (ni_hostname, ni_service))
-
-module Spawn = struct
-  let resolve_program ~paths prog =
-    if not (Filename.is_relative prog) then Some prog
-    else 
-    let rec loop = function
-      | path :: rest ->
-        let p = Filename.concat path prog in
-        if Sys.file_exists p then Some p else loop rest
-      | [] -> None
-    in
-    loop paths
-  
-  let spawn = Spawn.spawn
-end
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -903,6 +903,46 @@ module Low_level = struct
 
       external pidfd_wait : Unix.file_descr -> Unix.process_status = "caml_eio_pidfd_wait"
 
+      (* Some parts of the following code are modified from the spawn library. 
+         See the the eio_spawn_stubs.c file for the full license. *)
+      module Env = struct
+        type t = string list
+      
+        let no_null s =
+          if String.contains s '\000' then
+            Printf.ksprintf invalid_arg
+              "Spawn.Env.of_list: NUL bytes are not allowed in the environment but \
+               found one in %S"
+              s
+      
+        let of_list l =
+          List.iter no_null l;
+          l
+      end
+
+      module Cwd = struct
+        type internal =
+          | Path of string
+          | Fd of Unix.file_descr
+
+        type t = 
+          | Path of Eio.Fs.dir Eio.Path.t 
+          | Fd of FD.t
+      end
+
+      external spawn :
+        env:Env.t option
+        -> cwd:Cwd.internal
+        -> prog:string
+        -> argv:string list
+        -> stdin:Unix.file_descr
+        -> stdout:Unix.file_descr
+        -> stderr:Unix.file_descr
+        -> use_vfork:bool
+        -> setpgid:int option
+        -> sigprocmask:(Unix.sigprocmask_command * int list) option
+        -> int = "spawn_unix_byte" "spawn_unix"
+
       type t = {
         process : FD.t;
         mutable hook : Switch.hook;
@@ -929,16 +969,20 @@ module Low_level = struct
         in
         List.find_map exists paths
       
-      let spawn ?env ?cwd ?stdin ?stdout ?stderr ~sw prog argv =
+      let spawn ?env ~cwd ~stdin ~stdout ~stderr ~sw prog argv =
         let paths = Option.map (fun v -> String.split_on_char ':' v) (Sys.getenv_opt "PATH") |> Option.value ~default:[ "/usr/bin"; "/usr/local/bin" ] in
         let prog = match resolve_program ~paths prog with
           | Some prog -> prog
           | None -> raise (Eio.Fs.err (Eio.Fs.Not_found (Eio_unix.Unix_error (Unix.ENOENT, "", ""))))
         in
-        let stdin = Option.map (FD.to_unix `Peek) stdin in
-        let stdout = Option.map (FD.to_unix `Peek) stdout in
-        let stderr = Option.map (FD.to_unix `Peek) stderr in
-        let pid = Spawn.spawn ?env ?cwd ?stdin ?stdout ?stderr ~prog ~argv () in
+        let stdin = FD.to_unix `Peek stdin in
+        let stdout = FD.to_unix `Peek stdout in
+        let stderr = FD.to_unix `Peek stderr in
+        let cwd : Cwd.internal = match cwd with
+          | Cwd.Path p -> Cwd.Path (snd p)
+          | Cwd.Fd fd -> Cwd.Fd (FD.to_unix `Peek fd)
+        in
+        let pid = spawn ~env ~cwd ~stdin ~stdout ~stderr ~prog ~argv ~use_vfork:true ~setpgid:None ~sigprocmask:None in
         let fd = pidfd_open pid in
         let process = FD.of_unix ~sw ~seekable:false ~close_unix:true fd in
         let t = { process; hook = Switch.null_hook; status = None } in
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -949,7 +949,7 @@ module Low_level = struct
         t.hook <- hook;
         t
 
-        let send_signal t i = pidfd_send_signal (FD.to_unix `Peek t.process) i
+        let signal t i = pidfd_send_signal (FD.to_unix `Peek t.process) i
     end
 end
 
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -916,14 +916,23 @@ module Low_level = struct
             Option.iter Switch.remove_hook t.hook;
             t.status <- Some status;
             status
+
+      let resolve_program ~paths prog =
+        if not (Filename.is_implicit prog) then Some prog
+        else
+        let exists path =
+          let p = Filename.concat path prog in
+          if Sys.file_exists p then Some p else None
+        in
+        List.find_map exists paths
       
       let spawn ?env ?cwd ?stdin ?stdout ?stderr ~sw prog argv =
         let paths = Option.map (fun v -> String.split_on_char ':' v) (Sys.getenv_opt "PATH") |> Option.value ~default:[ "/usr/bin"; "/usr/local/bin" ] in
-        let prog = match Eio_unix.Spawn.resolve_program ~paths prog with
+        let prog = match resolve_program ~paths prog with
           | Some prog -> prog
           | None -> raise (Eio.Fs.err (Eio.Fs.Not_found (Eio_unix.Unix_error (Unix.ENOENT, "", ""))))
         in
-        let pid = Eio_unix.Spawn.spawn ?env ?cwd ?stdin ?stdout ?stderr ~prog ~argv () in
+        let pid = Spawn.spawn ?env ?cwd ?stdin ?stdout ?stderr ~prog ~argv () in
         let fd = pidfd_open pid in
         let process = FD.of_unix ~sw ~seekable:false ~close_unix:true fd in
         let t = { process; pid; hook = None; status = None } in
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -896,6 +896,45 @@ module Low_level = struct
     Eio_unix.run_in_systhread @@ fun () ->
     Unix.getaddrinfo node service []
     |> List.filter_map to_eio_sockaddr_t
+
+    module Process = struct
+      external pidfd_open : int -> Unix.file_descr = "caml_eio_pidfd_open" 
+
+      type t = {
+        process : FD.t;
+        pid : int;
+        mutable hook : Switch.hook option;
+        mutable status : Unix.process_status option;
+      }
+
+      let await_exit t =
+        match t.status with
+          | Some status -> status
+          | None ->
+            await_readable t.process;
+            let status = Unix.waitpid [] t.pid |> snd in
+            Option.iter Switch.remove_hook t.hook;
+            t.status <- Some status;
+            status
+      
+      let spawn ?env ?cwd ?stdin ?stdout ?stderr ~sw prog argv =
+        let paths = Option.map (fun v -> String.split_on_char ':' v) (Sys.getenv_opt "PATH") |> Option.value ~default:[ "/usr/bin"; "/usr/local/bin" ] in
+        let prog = match Eio_unix.Spawn.resolve_program ~paths prog with
+          | Some prog -> prog
+          | None -> raise (Eio.Fs.err (Eio.Fs.Not_found (Eio_unix.Unix_error (Unix.ENOENT, "", ""))))
+        in
+        let pid = Eio_unix.Spawn.spawn ?env ?cwd ?stdin ?stdout ?stderr ~prog ~argv () in
+        let fd = pidfd_open pid in
+        let process = FD.of_unix ~sw ~seekable:false ~close_unix:true fd in
+        let t = { process; pid; hook = None; status = None } in
+        let hook = Switch.on_release_cancellable sw (fun () ->
+          Unix.kill pid Sys.sigkill; ignore (await_exit t)
+        ) in
+        t.hook <- Some hook;
+        t
+
+        let send_signal t i = Unix.kill t.pid i
+    end
 end
 
 external eio_eventfd : int -> Unix.file_descr = "caml_eio_eventfd"
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -903,7 +903,7 @@ module Low_level = struct
       type t = {
         process : FD.t;
         pid : int;
-        mutable hook : Switch.hook option;
+        mutable hook : Switch.hook;
         mutable status : Unix.process_status option;
       }
 
@@ -913,7 +913,7 @@ module Low_level = struct
           | None ->
             await_readable t.process;
             let status = Unix.waitpid [] t.pid |> snd in
-            Option.iter Switch.remove_hook t.hook;
+            Switch.remove_hook t.hook;
             t.status <- Some status;
             status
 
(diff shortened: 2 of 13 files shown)


patricoferris:
Just as a confirmation that using `Spawn` might not be such a bad idea (and adding it to Eio_unix), I added a backend to kqueue using it + an `Eio.Process` module here: https://github.com/patricoferris/eio/commit/98755115bf8e4d2e985f61bb643e5eaceb370260

lib_eio/unix/eio_unix.ml
```diff
@@ -75,3 +75,18 @@ let getnameinfo (sockaddr : Eio.Net.Sockaddr.t) =
   run_in_systhread (fun () ->
     let Unix.{ni_hostname; ni_service} = Unix.getnameinfo sockaddr options in
     (ni_hostname, ni_service))
+
+module Spawn = struct
+  let resolve_program ~paths prog =
+    if not (Filename.is_relative prog) then Some prog
```
smondet, line 81: This condition seems to be missing relative paths that contain a `/`.
(cf. the difference between `prog` and `./prog` in posix shells)

smondet: If that helps I came up with this:

```ocaml
let resolve_program prog =
  (* Cf. https://github.com/ocaml-multicore/eio/pull/435/files *)
  let open Caml in
  let paths =
    Option.map (fun v -> String.split_on_char ':' v) (Sys.getenv_opt "PATH")
    |> Option.value ~default:[ "/usr/bin"; "/usr/local/bin" ]
  in
  let valid p = Sys.file_exists p && not (Sys.is_directory p) in
  if String.exists (function '/' -> true | _ -> false) prog then
    if valid prog then Some prog else None
  else
    let rec loop = function
      | path :: rest ->
          let p = Filename.concat path prog in
          if valid p then Some p else loop rest
      | [] -> None
    in
    loop paths

let%expect_test _ =
  let tmpdir = Caml.Filename.get_temp_dir_name () in
  let resolve_program_and_print p =
    let clean s =
      s
      |> String.substr_replace_all ~pattern:tmpdir ~with_:"${TMPDIR}"
      |> String.substr_replace_all ~pattern:"/usr/bin" ~with_:"${sysbin}"
      |> String.substr_replace_all ~pattern:"/bin" ~with_:"${sysbin}"
    in
    Fmt.pr "%s@ -> %a\n" (clean p)
      Fmt.Dump.(option string)
      (resolve_program p |> Option.map ~f:clean)
  in
  let ( // ) = Caml.Filename.concat in
  let cmd s =
    match Caml.Sys.command s with
    | 0 -> ()
    | n -> Fmt.failwith "%S -> %d" s n
  in
  resolve_program_and_print "echo";
  [%expect {|
    echo
    -> Some "${sysbin}/echo" |}];
  resolve_program_and_print "echooooo";
  [%expect {|
    echooooo
    -> None |}];
  let abs = tmpdir // "resolve_program_and_print.sh" in
  Fmt.kstr cmd "echo bouh > %s" abs;
  resolve_program_and_print abs;
  [%expect
    {|
    ${TMPDIR}/resolve_program_and_print.sh
    -> Some "${TMPDIR}/resolve_program_and_print.sh" |}];
  let rel = "./resolve_program_and_print.sh" in
  Fmt.kstr cmd "echo bah > %s" rel;
  resolve_program_and_print rel;
  [%expect
    {|
    ./resolve_program_and_print.sh
    -> Some "./resolve_program_and_print.sh" |}];
  resolve_program_and_print (Caml.Filename.basename rel);
  [%expect {|
    resolve_program_and_print.sh
    -> None |}];
  Fmt.kstr cmd "rm -f %s %s" abs rel;
  ()
```

talex5: `Filename.is_implicit` is likely what we want here.

lib_eio/unix/eio_unix.ml
```diff
@@ -75,3 +75,18 @@ let getnameinfo (sockaddr : Eio.Net.Sockaddr.t) =
...
+    else 
+    let rec loop = function
+      | path :: rest ->
+        let p = Filename.concat path prog in
+        if Sys.file_exists p then Some p else loop rest
+      | [] -> None
+    in
+    loop paths
```
talex5, line 89: Also `List.find_map` can be used for this.

lib_eio_linux/eio_linux.ml
```diff
@@ -917,6 +917,45 @@ module Low_level = struct
...
+        let prog = match Eio_unix.Spawn.resolve_program ~paths prog with
+          | Some prog -> prog
+          | None -> raise (Eio.Fs.err (Eio.Fs.Not_found (Eio_unix.Unix_error (Unix.ENOENT, "", ""))))
+        in
+        let pid = Eio_unix.Spawn.spawn ?env ?cwd ?stdin ?stdout ?stderr ~prog ~argv () in
+        let fd = pidfd_open pid in
+        let process = FD.of_unix ~sw ~seekable:false ~close_unix:true fd in
+        let t = { process; pid; hook = None; status = None } in
+        let hook = Switch.on_release_cancellable sw (fun () ->
+          Unix.kill pid Sys.sigkill; ignore (await_exit t)
```
talex5, line 952: As with luv, we need to call `Unix.waitpid` exactly once. We can't let the user do it. Probably needs something involving promises here.

Could maybe use `pidfd_send_signal` here? As it is it's not safe, as the process may have already been reaped by another call to `await_exit`.

lib_eio_linux/eio_linux.ml
```diff
@@ -917,6 +917,45 @@ module Low_level = struct
...
     Unix.getaddrinfo node service []
     |> List.filter_map to_eio_sockaddr_t
+
+    module Process = struct
+      external pidfd_open : int -> Unix.file_descr = "caml_eio_pidfd_open" 
+
+      type t = {
+        process : FD.t;
+        pid : int;
+        mutable hook : Switch.hook option;
```
talex5, line 927: You can avoid this option with `Eio.Switch.null_hook`.

lib_eio_linux/eio_linux.mli
```diff
@@ -248,4 +248,29 @@ module Low_level : sig
...
+    val spawn :
+      ?env:Spawn.Env.t ->
+      ?cwd:Spawn.Working_dir.t ->
+      ?stdin:Unix.file_descr ->
+      ?stdout:Unix.file_descr ->
+      ?stderr:Unix.file_descr ->
+      sw:Switch.t ->
+      string ->
+      string list ->
+      t
```
talex5, line 264: Should be made safe like the rest of the API (e.g. using `FD.t`).

lib_eio_linux/eio_stubs.c
```diff
@@ -99,3 +99,15 @@ CAMLprim value caml_eio_getdents(value v_fd) {
...
+
+static int pidfd_open(pid_t pid, unsigned int flags) {
+  return syscall(__NR_pidfd_open, pid, flags);
+}
+
+CAMLprim value caml_eio_pidfd_open(value v_pid) {
+  CAMLparam1(v_pid);
+  int fd;
+  // Returns a file descriptor for the PID with close-on-exec set.
+  fd = pidfd_open(Int_val(v_pid), 0);
```
talex5, line 111: Need to check for errors here.

lib_eio/unix/eio_unix.ml
```diff
@@ -75,3 +75,18 @@ let getnameinfo (sockaddr : Eio.Net.Sockaddr.t) =
...
+  
+  let spawn = Spawn.spawn
```
talex5, line 91: Seems pointless to take a dependency on `spawn` here just to re-export its API.

talex5 (commented):
Comments in-line.

I'd still like proper `clone3` support at some point, but using spawn is fine for now (just not in `eio.unix`).

Note that I'm changing the way FDs work a bit in #440, so you might want to rebase on that before changing anything.

talex5:
Regarding PATH, I'd leave that out of the low-level API (which should just export what the OS provides).

lib_eio_linux/eio_stubs.c
```diff
@@ -99,3 +102,86 @@ CAMLprim value caml_eio_getdents(value v_fd) {
...
+        st = caml_alloc_small(1, TAG_WSIGNALED);
+        Field(st, 0) =
+            Val_int(caml_rev_convert_signal_number(WTERMSIG(status)));
+    }
+    return st;
+}
+
+CAMLprim value caml_eio_pidfd_wait(value v_pidfd) {
+  CAMLparam1(v_pidfd);
+  CAMLlocal1(status);
```
avsm, line 179: minor style thing here: a `v_` prefix tends to indicate a `value` type, so you could call this `v_status` to indicate that `alloc_process_status` fills in this variable with a `value`.

lib_eio_linux/eio_linux.mli
```diff
@@ -248,4 +248,29 @@ module Low_level : sig
...
+      ?stderr:FD.t ->
+      sw:Switch.t ->
+      string ->
+      string list ->
+      t
+    (** Spawns a subprocess. By default all of the optional arguments are inherited from the
+        calling process. If the process has not finished when the switch is released, the process
+        will be sent [Sys.sigkill]. *)
+
+    val await_exit : t -> Unix.process_status
```
avsm, line 269: why not just call this `wait`? It's shorter and has precedent in *wait(2)*.

lib_eio_linux/eio_linux.mli
```diff
@@ -248,4 +248,29 @@ module Low_level : sig
...
+    (** [await_exit t] waits for the process [t] to exit. This blocks the fiber until the process
+        has finished. *)
+
+    val send_signal : t -> int -> unit
```
avsm, line 273: similarly (although it's a bit weirder), this could just be `kill` as it has the same semantics as `Unix.kill`.  Easier to find in the interface if it's got a similar name.

lib_eio_linux/eio_linux.mli
```diff
@@ -248,4 +248,29 @@ module Low_level : sig
   (** [getaddrinfo host] returns a list of IP addresses for [host]. [host] is either a domain name or
       an ipaddress. *)
 
+  module Process : sig
+    type t
+    (** A subprocess *)
+
+    val spawn :
+      ?env:Spawn.Env.t ->
```
avsm, line 256: Process management is a pretty core function of eio.  Given that `spawn` is just a single set of ml/mli/c files, perhaps we should just vendor it in here in Eio_linux and give local assignments to the Env and Working_dir modules, to keep Eio more standalone.

lib_eio_linux/eio_linux.mli
patricoferris: Sounds good to me, done in https://github.com/ocaml-multicore/eio/pull/435/commits/944664cfd161d60d325ab293d801a8cbd653f4b3 (I think it was named that way initially because I exposed the status as a promise to the user, hence `await`).

lib_eio_linux/eio_stubs.c
patricoferris: Yes! https://github.com/ocaml-multicore/eio/pull/435/commits/944664cfd161d60d325ab293d801a8cbd653f4b3

lib_eio_linux/eio_linux.mli
patricoferris: Hmm, I intentionally didn't go with `kill` because it seemed unnecessarily violent (see [this draft](https://datatracker.ietf.org/doc/draft-knodel-terminology/) and it's conclusions on reflecting on the type of metaphors we use) and (to someone with less experience like me) the connection to `Unix.kill` wasn't that obvious (e.g. why does `Unix.kill` need to be passed `Sys.sigkill` (my conclusion, `kill` is poorly named) ?). It also now uses `pidfd_send_signal`, but at the same time if it makes the API easier to use I'm happy to be persuaded if there's consensus that it would be better :))

lib_eio_linux/eio_stubs.c
patricoferris: Yep https://github.com/ocaml-multicore/eio/pull/435/commits/46c19a72dfeca5e7f692e660a1e78d756a666154

lib_eio_linux/eio_linux.mli
patricoferris: https://github.com/ocaml-multicore/eio/pull/435/commits/0685201dd2c0c19d65048e20b3823d83d0466112

lib_eio_linux/eio_linux.ml
patricoferris: https://github.com/ocaml-multicore/eio/pull/435/commits/ab5a1faf142c04c5dd6ddf4807e5b479fd7b26c0

lib_eio_linux/eio_linux.mli
patricoferris: Seems reasonable to me, I think (though I haven't tried it yet) that would also mean switching over to `clone3` would be fairly straightforward too.

lib_eio_linux/eio_linux.mli
avsm: I'm entirely in agreement with the importance of avoiding violent language in systems programming.  I also dislike typing long function names. How about just `Process.signal`?

patricoferris: Agreed, done in https://github.com/ocaml-multicore/eio/pull/435/commits/f6bdf04b1307cd1190f760de4fdd70049f32d7e3

lib_eio_linux/eio_linux.mli
patricoferris: I'm experimenting with vendoring it over in https://github.com/patricoferris/eio/commit/c8a28ebca940d5980fa39f7e47cf33f04796f0bc for the moment

lib_eio_linux/eio_linux.mli
talex5: We don't want to call `wait(2)` here, though. The process should get reaped whether the users calls this or not. This function should just be awaiting the promise.

lib_eio_linux/eio_linux.ml
```diff
@@ -896,6 +896,61 @@ module Low_level = struct
...
+      
+      let spawn ?env ?cwd ?stdin ?stdout ?stderr ~sw prog argv =
+        let paths = Option.map (fun v -> String.split_on_char ':' v) (Sys.getenv_opt "PATH") |> Option.value ~default:[ "/usr/bin"; "/usr/local/bin" ] in
+        let prog = match resolve_program ~paths prog with
+          | Some prog -> prog
+          | None -> raise (Eio.Fs.err (Eio.Fs.Not_found (Eio_unix.Unix_error (Unix.ENOENT, "", ""))))
+        in
+        let stdin = Option.map (FD.to_unix `Peek) stdin in
+        let stdout = Option.map (FD.to_unix `Peek) stdout in
+        let stderr = Option.map (FD.to_unix `Peek) stderr in
```
talex5, line 940: Using `Peek` isn't safe (I'm planning to remove it). Use `FD.use_exn` instead to keep the FD open during the spawn operation.

patricoferris: Thanks!

patricoferris:
W.r.t this comment:

> Regarding PATH, I'd leave that out of the low-level API (which should just export what the OS provides).

We now vendor spawn so have a little more control over what we can do on the C-side of things, did you mean to just mark it as calling `execve` and let the user worry about things like `PATH` ?

talex5:
> We now vendor spawn so have a little more control over what we can do on the C-side of things, did you mean to just mark it as calling execve and let the user worry about things like PATH ?

Yes.

lib_eio_linux/eio_linux.mli
```diff
@@ -248,4 +248,46 @@ module Low_level : sig
...
+      ?env:Env.t ->
+      cwd:Cwd.t ->
+      stdin:FD.t ->
+      stdout:FD.t ->
+      stderr:FD.t ->
+      sw:Switch.t ->
+      string ->
+      string list ->
+      t
+    (** Spawns a subprocess. If the process has not finished when the switch is released, the process
```
avsm, line 281: This diff is looking really good! Just one thing about the documentation though: there is quite a lot of useful information in the underlying [spawn ocamldoc](https://github.com/janestreet/spawn/blob/master/src/spawn.mli#L76) that is worth propagating here, now that the underlying library is no longer exposed.  For example, it answers: "what happens if [env] is None" in that documentation, which is quite important to know.

lib_eio_linux/eio_linux.mli
```diff
@@ -248,4 +248,46 @@ module Low_level : sig
...
+        will be sent [Sys.sigkill]. [spawn] uses [execve] which means the program will {e not} be searched
+        for in the [PATH] environment variable. *)
+
+    val wait : t -> Unix.process_status
+    (** [wait t] waits for the process [t] to exit. This blocks the fiber until the process
+        has finished. *)
+
+    val signal : t -> int -> unit
+    (** [signal t s] send a signal [s] to process [t]. For example to stop a process you could use
```
avsm, line 290:
```suggestion
    (** [signal t s] sends a signal [s] to process [t]. For example, to terminate a process use
```

confusingly, SIGSTOPing a process is different from SIGTERMing a process :-)

lib_eio_linux/eio_linux.mli
```diff
@@ -248,4 +248,46 @@ module Low_level : sig
...
+      string ->
+      string list ->
+      t
+    (** Spawns a subprocess. If the process has not finished when the switch is released, the process
+        will be sent [Sys.sigkill]. [spawn] uses [execve] which means the program will {e not} be searched
+        for in the [PATH] environment variable. *)
+
+    val wait : t -> Unix.process_status
+    (** [wait t] waits for the process [t] to exit. This blocks the fiber until the process
+        has finished. *)
```
avsm, line 287:
```suggestion
    (** [wait t] blocks the fiber until process [t] has terminated, either by
          exiting or receiving a signal. *)
```

talex5:
The spawn library has a single function that does a lot of stuff, and different platforms will want to provide different features. I'm wondering if it would be better to make it more modular.

The basic problem we have is that we can't call OCaml code after forking. However, we can make a list of C functions to call. Here's an experiment that defines a `Fork_action.t` and has spawn take a list of those:

  https://github.com/ocaml-multicore/eio/compare/main...talex5:eio:spawn?expand=1

e.g. (I put this in eio_posix, but it's demonstrating the Linux-specific `execveat` and `fchdir` actions here):

```ocaml
open Eio.Std

module L = Eio_posix.Low_level

let () =
  Eio_posix.run @@ fun _env ->
  Switch.run @@ fun sw ->
  let exe = L.openat ~sw ~mode:0 "/bin/ls" L.Open_flags.(rdonly + excl) in
  let dir = L.openat ~sw ~mode:0 "/etc" L.Open_flags.(rdonly + directory + excl) in
  L.Fd.use_exn "exec" exe @@ fun exe ->
  L.Fd.use_exn "chdir" dir @@ fun dir ->
  let pid = L.spawn ~sw L.Fork_action.[
      fchdir dir;
      execat exe [| "ls"; "-l" |];
    ] in
  traceln "spawned PID %d" pid
```

Probably some of this could go in Eio_unix and be shared between the various backends. This would all be a private API for backends; we can present something higher-level to the user.

avsm:
`Fork_actions` could also be expanded to include privilege dropping syscalls for the subprocess too, which is rather neat. E.g. a call to [`Fork_action.pledge`](https://man.openbsd.org/pledge.2) in a future Eio_openbsd :-)

talex5:
> The example is a simple case that I think (at least in my mind) is easier to understand/code with something like what we have i.e. spawn ~cwd:dir "ls"

Yes, I'd expect the high-level API to build this up for you.

> but we have no nice way to say chdir; exec; chdir; exec inside the same fork.

You can't do anything after `exec` (well, unless it fails, I guess). But things that might be useful include:

- Reassigning FDs.
- Dropping privileges (`pledge` etc, as Anil mentioned), which each platform does differently.
- Setting up container namespaces.
- Changing directory (with `chdir` or `fchdir`).
- Maybe progress group stuff? Signal masks?
- Execing with `execve` or `execveat`.

And being able to do this with `fork`, `vfork` or `clone3`.

> I'm really just thinking out loud at this point but it might be nice to be able to do something like `pipe (exec foo) (exec bar)` although again that could probably be built on top if this anyway...

I think that would be built on top. This API is just for stuff that can only happen after forking, where we can't use OCaml.

> Not quite sure where to go from here, experiment with this API first porting more of the spawn code to this modular approach?

Yes, I think so. I can try to clean it up over the next couple of days. Do you want to try putting the higher-level API on top after that?

RyanGibb:
I currently have a problem where there are some syscalls I need to make after forking but before execing. This is to [set up a PTY for the child process](https://github.com/RyanGibb/ocaml-exec-shell/blob/eedf2b5105f2455d651e42c76fb43a8fc2abc7b6/src/main.ml#L61). I've got a [branch](https://github.com/RyanGibb/ocaml-exec-shell/blob/eio-processes/src/main.ml) using this PR at the moment, but without the PTY setup. While writing a C wrapper could work, I think @talex5's proposal would provide a good solution for this.

patricoferris:
> Yes, I think so. I can try to clean it up over the next couple of days. Do you want to try putting the higher-level API on top after that?

Sounds good to me :))

patricoferris:
Closing in favour of #472 :))

PR #436, opened 2023-02-03
Add more Eio.Condition documentation

talex5:
This adds a discussion of conditions to the README and provides examples using them to handle signals.

Closes #374, #321, #301.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,9 @@ Eio replaces existing concurrency libraries such as Lwt
   * [Example: Concurrent Cache](#example-concurrent-cache)
   * [Streams](#streams)
   * [Example: Worker Pool](#example-worker-pool)
-  * [The Rest: Mutex, Semaphore and Condition](#the-rest-mutex-semaphore-and-condition)
+  * [Mutexes and Semaphores](#mutexes-and-semaphores)
+  * [Conditions](#conditions)
+  * [Example: Signal handlers](#example-signal-handlers)
 * [Design Note: Determinism](#design-note-determinism)
 * [Provider Interfaces](#provider-interfaces)
 * [Example Applications](#example-applications)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1291,6 +1292,186 @@ let release () =
...
++x is now zero
+- : unit = ()
+```
+
+Note that we need a loop in `await_zero`.
+This is needed because it's possible that another fiber might set it to zero
+and then set it to something else before the waiting fiber resumes.
+
+The above version is not safe to share across domains, because `await_zero` relies on the value of `x` not changing
+after it reads it but before `await_no_mutex` registers itself with the condition.
```
haesbaert, line 1339: "after it is read" maybe ?

README.md
```diff
@@ -1291,6 +1292,186 @@ let release () =
...
+- : unit = ()
+```
+
+Conditions are more difficult to use correctly than e.g. promises or streams.
+In particular, it is easy to miss a notification due to `broadcast` getting called before `await`.
+However, they can be useful if used carefully.
+
+### Example: Signal handlers
+
+On Unix-type systems, processes can react to *signals*.
```
haesbaert, line 1390: I think this holds for windows as well, no clue to what extend though.

README.md
```diff
@@ -1291,6 +1292,186 @@ let release () =
...
+         (* We could cancel the loading fiber now, in case it's still running,
+            but in this example we just wait for it to finish by itself. *)
+      )
+      (fun () ->
+         traceln "Reading configuration ('kill -SIGHUP %d' to reload)..." (Unix.getpid ());
+         Eio_unix.sleep 2.0;    (* Simulate loading config *)
+         traceln "Finished reading configuration";
+      )
+  done
+```
```
haesbaert, line 1467: I was wonder if it might be worth pointing the user how he can then communicate outside that the configuratoin changed and was reloaded, this assumes the configuration is stored in some global reference, but that's probably too much, so it's fine as it is.

README.md
talex5: I don't think so; if you try to ignore SIGPIPE on Windows then it raises an exception, for example (see #420).

README.md
talex5: Yeah, too many `it`s here. I replaced it with ``after `x` is read``.

README.md
talex5: I added some comments to the full example, but just factored this out as `load_config ()` in the README, as it's not very relevant to handling signals.

PR #437, opened 2023-02-05
Update changelog

talex5:


talex5:
I did a test commit disabling the crowbar and dscheck tests, which allows the other archs to build. Then only s390x failed. Removing the eio_luv tests then passed (since s390x uses eio_linux by default, which works).

Issue #434, opened 2023-02-02
Replace libuv


talex5:
libuv is not suitable for an Eio backend. Problems include:

- Resources are attached to loops, which doesn't work if you want to share them between domains (#386).
- It interferes with signals (#400), requiring a work-around.
- It is unmaintained (#433, #335).
- It interferes with other effect handlers (#427).
- File operations can't be cancelled (#373, #357).
- It provides a second abstraction layer (e.g. chooses a backend for your platform, then libuv chooses another one). It would be simpler to have only one OS abstraction.
- It doesn't support many useful features (e.g. #366, `openat`).

The main reason to use it was to get Windows support for free. However, that doesn't fully work (https://github.com/aantron/luv/issues/142) and we hope to provide a native Windows backend instead.

At least on POSIX systems, we should be able to provide a much simpler Eio_posix backend.

patricoferris:
In case it is useful I have a mostly working (but needs more testing and of course reviewing!) [kqueue](https://github.com/anuragsoni/kqueue-ml)-based backend https://github.com/patricoferris/eio/tree/kqueue/lib_eio_kqueue. I'm a little more confident in this backend for macOS than GCD at the moment and we might get BSDs out of it too (although kqueue implementations vary between them all it seems)

tmcgilchrist:
Out of interest @patricoferris what is the issue with using macOS GCD? Are there technical limitations or is it a matter of developer time?

patricoferris:
> Out of interest @patricoferris what is the issue with using macOS GCD? Are there technical limitations or is it a matter of developer time?

Good question! I've been hacking away without really completing both a kqueue-based backend and a GCD backend whilst reading quite a bit about both, and here's my opinionated thoughts on that matter:

 - From the outside it would seem GCD is not the best option for something like macOS -- I think it can still make sense for some applications (e.g. iOS) but after scouring many a forum, this is the general picture I'm getting from devs working in the Apple ecosystem. Have a read of this forum discussion https://forums.swift.org/t/task-safe-way-to-write-a-file-asynchronously/54639/ (I find it quite compelling that SwiftNIO did not choose GCD for macOS, they do have an extension to support more Apple devices that does use GCD for some things https://github.com/apple/swift-nio-extras).
[...]

PR #448, opened 2023-02-23
Initial Eio_posix backend

talex5:
This adds a new backend for POSIX systems. This is intended to be used as the fallback backend on most platforms, replacing libuv (see #434).

This isn't quite ready, but I thought some people might like to look it at early. Especially @patricoferris, as it may be relevant to his kqueue work.

The main commit uses the `poll` library, which has kqueue support, but a later commit replaces this with @haesbaert's new `iomux` library. However, this might not work on macos.

This is on top of #447. Use `make test_posix` to run the tests.

TODO before merging:
- [x] Release iomux or vendor it properly (not as a submodule). Sort out opam deps.
- [x] Add CI to test this (all ocaml-ci platforms are Linux at the moment).
- [x] Add `pread[v]`, `pwrite[v]`.
- [x] Better `readdir` support.
- [x] Add support for opening with `O_NOFOLLOW`.
- [x] Add `openat`, `mkdirat`, etc.

Other things to work on after merging this initial support:
- Fix errors from `getaddrinfo` (see #351).
- Add a systhread pool of workers so that e.g. file operations never block the main thread.
- Decide whether/how to make this more flexible, so that e.g. most of this can be reused with kqueue systems.
- Ensure we yield on IO to prevent starvation.


haesbaert:
:cow: moo I'm back today, I did release a iomux 0.1 just before leaving, but that won't work on macos cause, believe it or not, it doens't have ppoll(2). I'll make a new release today that makes it fallback nicely to poll(2) on macos.

lib_eio_posix/eio_posix_stubs.c
```diff
@@ -0,0 +1,107 @@
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/eventfd.h>
```
patricoferris, line 3:
```suggestion
```

I think this can be deleted as it isn't used (nor available on macOS).

lib_eio_posix/eio_posix_stubs.c
```diff
@@ -0,0 +1,107 @@
...
+
+CAMLprim value caml_eio_posix_getrandom(value v_ba, value v_off, value v_len) {
+  CAMLparam1(v_ba);
+  ssize_t ret;
+  ssize_t off = (ssize_t)Long_val(v_off);
+  ssize_t len = (ssize_t)Long_val(v_len);
+  do {
+    void *buf = Caml_ba_data_val(v_ba) + off;
+    caml_enter_blocking_section();
+    ret = getrandom(buf, len, 0);
```
patricoferris, line 34: `getrandom` is not available on macOS it would seem, there's `getentropy` and `arc4random` which look like they may be a good fit?

lib_eio_posix/low_level.ml
```diff
@@ -0,0 +1,157 @@
...
+
+let rec connect fd addr =
+  try
+    Fd.use_exn "connect" fd (fun fd -> Unix.connect fd addr)
+  with
+  | Unix.Unix_error (EINTR, _, _) -> connect fd addr   (* Just in case *)
+  | Unix.Unix_error ((EAGAIN | EWOULDBLOCK | EALREADY | EINPROGRESS), _, _) ->
+    await_writable fd;
+    match Fd.use_exn "connect" fd Unix.getsockopt_error with
+    | None -> connect fd addr
```
patricoferris, line 65: This part, on macOS, is failing the networking tests. If I understand it correctly, wouldn't an `EINPROGRESS` plus waiting for the FD to be writeable mean by the time we get to the match statement the error would be cleared and the socket would be connected meaning this second connect would fail with:

```
Eio.Io Unix_error (Socket is already connected, "connect", "")
```

patricoferris (commented):
This great and is mostly working on macOS (M1, 13.2). Apart from the inline comments, in the README there is a little test for copying from stdin to stdout:

```ocaml
let () =
    Eio_main.run @@ fun env ->
    Eio.Flow.copy
      (Eio.Stdenv.stdin env)
      (Eio.Stdenv.stdout env);;
```

this is currently making MDX hang when testing the README.

lib_eio_posix/low_level.ml
haesbaert: You're correct, I think the test is wrong. There shouldn't be a second connect, if SO_ERROR is cleared then it's all done and there's nothing else to do (says Mr.Stevens).
It probably only works on the linux case because a local connection is allowed to never return EINPROGRESS and finish immediately.

lib_eio_posix/eio_posix_stubs.c
haesbaert: ack, getentropy() would be the thing. in fact getrandom() was designed after getentropy

lib_eio_posix/low_level.ml
talex5: Makes sense (the man-page is a bit vague about it). What about the other cases (`EAGAIN | EWOULDBLOCK | EALREADY`)?

The Linux man-page says:
> EINPROGRESS The socket is nonblocking and the connection cannot be completed immediately.  (UNIX domain sockets failed with EAGAIN  instead.) [ then talks about waiting and checking for errors ]

But the EAGAIN case doesn't say what you should do in that case.

haesbaert: On a TCP socket you shouldn't get EAGAIN/EWOULDBLOCK (IIRC you can get on an AF_UNIX/SOCKSTREAM).

Since now you don't issue a secondary connect you should also never see an EALREADY. 

**But**, I think you're missing the EINTR case, which should be handled the same way as a blocking, which means EINTR does not cancel a connect so you should wait for writeable and check the error.

lib_eio_posix/low_level.ml
```diff
@@ -60,10 +60,11 @@ let rec connect fd addr =
   with
   | Unix.Unix_error (EINTR, _, _) -> connect fd addr   (* Just in case *)
```
haesbaert, line 61: This EINTR should be handled like EINPROGRESS below, and I think you can zap EAGAIN | EWOULDBLOCK | EALREADY

lib_eio_posix/low_level.ml
talex5: OK. Since we're always nonblocking, I guess we shouldn't normally see EINTR, but I'll treat it the same anyway.

talex5:
> in the README there is a little test for copying from stdin to stdout [...] this is currently making MDX hang when testing the README.

I guess stdin isn't `/dev/null` for some reason? I marked that bit as not needing to be run, since it's not supposed to output anything anyway.

talex5:
@haesbaert I made some minor changes to the iomux bits, which looked safe to me:

- I think there's no need to mark `eventfd_r` for reading after each event, as it should still be set.
- I looks like you don't need to set all entries to invalid at the start, as `Poll` does it anyway.

lib_eio_posix/eio_posix_stubs.c
```diff
@@ -0,0 +1,178 @@
...
+  CAMLparam1(v_path);
+  char *path;
+  int ret;
+  caml_unix_check_path(v_path, "mkdirat");
+  path = caml_stat_strdup(String_val(v_path));
+  caml_enter_blocking_section();
+  ret = mkdirat(Int_val(v_fd), path, Int_val(v_perm));
+  caml_leave_blocking_section();
+  caml_stat_free(path);
+  if (ret == -1) uerror("mkdirat", v_path);
```
avsm, line 142: are you sure that *free(2)* won't clobber errno? It was a bit [muddy in the past](https://sourceware.org/bugzilla/show_bug.cgi?id=17924) but might be the case now.  To be safe, should save errno right after the function call before any other functions.  But calling free might be safe nowadays...

talex5: Interesting. https://github.com/ocaml/ocaml/blob/trunk/otherlibs/unix/mkdir.c seems to assume errno won't change.

haesbaert: :fearful: I had no idea glibc could do that, that's quite surprising.

avsm: malloc and free are pretty complex beasts on Linux, but the errno-setting happens in pretty specific cases (usually when using the mmap mode).  Irritatingly hard to predict, but triggers a more often in virtualised platforms where there are a lot of VMA games going in with the virtual memory subsystem.

avsm:
@talex5 I took the liberty of pushing e19bc4a912fcb9ca8050578e94901f6411667df8 to your branch, which fixes the build on OpenBSD for me (and seems to work, at a quick spin).  Incidentally, the `luv` build is also broken on OpenBSD, which I won't bother debugging since this will replace it.

lib_eio_posix/sched.ml
```diff
@@ -0,0 +1,348 @@
...
+    Hashtbl.remove t.fd_map fd
+  ) else (
+    Poll.set_index t.poll fdi fd flags;
+    if fdi > t.poll_maxi then
+      t.poll_maxi <- fdi
+  )
+
+let resume t node =
+  t.active_ops <- t.active_ops - 1;
+  let k : unit Suspended.t = Lwt_dllist.get node in
```
avsm, line 130: Should we also just vendor in `Lwt_dllist` and give it a reasonable module name that is independent of Lwt?  I made a mistake of preserving the prefix when I originally pulled it out, and it's a little confusing to see a reference to Lwt at the heart of Eio when there is in fact none!

lib_eio_posix/domain_mgr.ml
```diff
@@ -0,0 +1,76 @@
...
+      | Eio_unix.Private.Get_monotonic_clock -> Some (fun k -> continue k (Time.mono_clock : Eio.Time.Mono.t))
+      | Eio_unix.Private.Socket_of_fd (sw, close_unix, fd) -> Some (fun k ->
+          Unix.set_nonblock fd;
+          let fd = Fd.of_unix ~sw ~blocking:false ~close_unix fd in
+          continue k (Flow.of_fd fd :> Eio_unix.socket)
+        )
+      | Eio_unix.Private.Socketpair (sw, domain, ty, protocol) -> Some (fun k ->
+          let a, b = Unix.socketpair ~cloexec:true domain ty protocol in
+          Unix.set_nonblock a;
+          Unix.set_nonblock b;
```
avsm, line 37: if these nonblocks raise an exception, do we leak the socketpair? (it'll raise the Unix_error over to the continuation, but the socketpair will linger).

talex5: Actually, it will crash the event loop if that happens (`continue` is only used on success). I can't think of any reason why `set_nonblock` should fail, but we should probably handle `socketpair` failing.

lib_eio_posix/eio_posix_stubs.c
talex5: I've now wrapped the free function as `caml_stat_free_preserving_errno` to avoid that.

lib_eio_posix/sched.ml
talex5: Perhaps we should rename the `lwt_dllist` library itself (and provide `Lwt_dllist` as a compatibility alias)?

If we're going to change it in Eio, it should be a separate issue because we already depend on it elsewhere.

lib_eio_posix/domain_mgr.ml
talex5: Updated error handling is in https://github.com/ocaml-multicore/eio/pull/448/commits/3f3aab43388f1218b0df567234b67ef39c4a776e

haesbaert:
Finally coming around to review this, sorry for the delay, if you can believe, I had no internet this morning.
Commited two small nits. 
 - No need to conditionally include stdlib 53e108059bab15a20e3d775dbf1b079705d2dd93
 - Missing ; in tests 99611861a8c5aca3b9139583a801f3152aff1b80

haesbaert:
> @haesbaert I made some minor changes to the iomux bits, which looked safe to me:
> 
> * I think there's no need to mark `eventfd_r` for reading after each event, as it should still be set.
> * I looks like you don't need to set all entries to invalid at the start, as `Poll` does it anyway.

Ack, I've changed it to invalidate always, that was just way too dangerous.
I was just wondering, I see you always use `==` to compare a `Unix.file_desc`, why is that ?

haesbaert:
I had accidentally force-pushed here, I removed, I'm keeping my additional changes here:
https://github.com/ocaml-multicore/eio/compare/main...haesbaert:eio:eio-posix

lib_eio_posix/eio_posix_stubs.c
```diff
@@ -0,0 +1,190 @@
...
+  errno = saved;
+}
+
+CAMLprim value caml_eio_posix_getrandom(value v_ba, value v_off, value v_len) {
+  CAMLparam1(v_ba);
+  ssize_t ret;
+  ssize_t off = (ssize_t)Long_val(v_off);
+  ssize_t len = (ssize_t)Long_val(v_len);
+  do {
+    void *buf = Caml_ba_data_val(v_ba) + off;
```
haesbaert, line 41:
```suggestion
    void *buf = (uint8_t *)Caml_ba_data_val(v_ba) + off;
```

lib_eio_posix/eio_posix_stubs.c
```diff
@@ -0,0 +1,190 @@
...
+
+/* Fill [iov] with pointers to the cstructs in the array [v_bufs]. */
+static void fill_iov(struct iovec *iov, value v_bufs) {
+  int n_bufs = Wosize_val(v_bufs);
+  for (int i = 0; i < n_bufs; i++) {
+    value v_cs = Field(v_bufs, i);
+    value v_ba = Field(v_cs, 0);
+    value v_off = Field(v_cs, 1);
+    value v_len = Field(v_cs, 2);
+    iov[i].iov_base = Caml_ba_data_val(v_ba) + Long_val(v_off);
```
haesbaert, line 63:
```suggestion
    iov[i].iov_base = (uint8_t *)Caml_ba_data_val(v_ba) + Long_val(v_off);
```

lib_eio_posix/test/poll.md
```diff
@@ -0,0 +1,22 @@
+```ocaml
+# #require "eio_posix";;
+```
+
+```ocaml
+open Eio.Std
+```
+
+## Closing an FD removes it from epoll
```
haesbaert, line 9:
```suggestion
## Closing an FD removes it from the multiplexer
```

lib_eio_posix/test/poll.md
```diff
@@ -0,0 +1,22 @@
...
+```ocaml
+# Eio_posix.run @@ fun _env ->
+  Switch.run (fun sw ->
+     let r, w = Eio_unix.pipe sw in
+     Eio_unix.await_writable (Eio_unix.FD.peek w)
+  );
+  Switch.run (fun sw ->
+     let r, w = Eio_unix.pipe sw in
+     Eio_unix.await_writable (Eio_unix.FD.peek w)
+  );;
```
haesbaert, line 20: duplicated ?

lib_eio_posix/sched.ml
haesbaert: I'd leave it as a Homage to Lwt :dancers:

lib_eio_posix/eio_posix.ml
```diff
@@ -0,0 +1,48 @@
...
+  fs : Eio.Fs.dir Eio.Path.t;
+  cwd : Eio.Fs.dir Eio.Path.t;
+  secure_random : Eio.Flow.source;
+  debug : Eio.Debug.t;
+>
+
+let run main =
+  (* SIGPIPE makes no sense in a modern application. *)
+  Sys.(set_signal sigpipe Signal_ignore);
+  Domain_mgr.run_event_loop main @@ object (_ : stdenv)
```
haesbaert, line 36: Nice that the main domain goes through the same path as the manager !

eio_main.opam
```diff
@@ -10,11 +10,13 @@ doc: "https://ocaml-multicore.github.io/eio/"
 bug-reports: "https://github.com/ocaml-multicore/eio/issues"
 depends: [
   "dune" {>= "3.0"}
-  "eio_linux" {= version & os = "linux"}
   "mdx" {>= "1.10.0" & with-test}
-  "eio_luv" {= version}
+  "eio_linux" {= version & os = "linux"}
+  "eio_posix" {= version & os != "windows"}
```
haesbaert, line 15: is there no `os_type` here like dune ?

lib_eio_posix/domain_mgr.ml
haesbaert: I just did a quick check in the OpenBSD kernel, I'd say it's pretty safe to assume you can expect set_nonblock to be ok, as long as we know what the file descriptor is. There are some weird cases like trying to set_nonblock on a kqueue fd will fail since there is no ioctl handler.

talex5:
> Missing ; in tests

Note that MDX no longer requires this (see https://github.com/realworldocaml/mdx/pull/398).

>  I was just wondering, I see you always use == to compare a Unix.file_desc, why is that ?

It's only one place (`if fd == t.eventfd_r`). I don't remember why it's like that (might have been a typo, or maybe a previous version was comparing `Fd.t` rather than `Unix.file_descr`).

> I had accidentally force-pushed here, I removed, I'm keeping my additional changes here:
https://github.com/ocaml-multicore/eio/compare/main...haesbaert:eio:eio-posix

I don't see any commits there. It's OK to push to this branch (no need to force-push). We can squash before merging if needed.

lib_eio_posix/flow.ml
```diff
@@ -0,0 +1,92 @@
...
+      gid     = ust.st_gid   |> Int64.of_int;
+      rdev    = ust.st_rdev  |> Int64.of_int;
+      size    = ust.st_size  |> Optint.Int63.of_int64;
+      atime   = ust.st_atime;
+      mtime   = ust.st_mtime;
+      ctime   = ust.st_ctime;
+    }
+  with Unix.Unix_error (code, name, arg) -> raise @@ Err.wrap code name arg
+
+let write_bufs fd bufs =
```
haesbaert, line 30: Not particular to this change but I realize we might want to consider exporting a `single_write` like `single_read`, the reason for that is that if any of the Flow write fails, the user has no way to know how many bytes have been copied before getting the exception.

lib_eio_posix/fd.ml
```diff
@@ -0,0 +1,60 @@
...
+type t = {
+  fd : Rcfd.t;
+
+  (* stdin, stdout and stderr are blocking, and so need special care.
+     For these, we first wait for them to become e.g. readable and then hope
+     that the read doesn't block. This may fail if multiple fibers try to read
+     at the same time. We could check that it's still readable after being
+     resumed, but that still won't work if multiple domains read at the same
+     time. Same problem for writes. *)
+  blocking : bool;
```
haesbaert, line 14: I think it might be worth considering making stdin nonblocking, and being careful to restore the blockiness on fork. 
This way at least we guarantee it's safe (even if silly) to have two fibers waiting on input.
I'd not touch stderr or stdout, because users will still end up using Printf directly, but stdin we might consider assuming "this is ours". At any rate would be nice to see what others are doing, maybe the nonblock->restore dance ?

haesbaert (commented):
So this is a partial review, I'm still missing the harder bits, but this is a bit too much to do in one go.

lib_eio_posix/sched.ml
haesbaert: I'd live it as a Homage to Lwt :dancers:

lib_eio_posix/eio_posix_stubs.c
```diff
@@ -0,0 +1,190 @@
...
+CAMLprim value caml_eio_posix_renameat(value v_old_fd, value v_old_path, value v_new_fd, value v_new_path) {
+  CAMLparam2(v_old_path, v_new_path);
+  int old_path_len = caml_string_length(v_old_path);
+  int new_path_len = caml_string_length(v_new_path);
+  char *old_path;
+  char *new_path;
+  int ret;
+  caml_unix_check_path(v_old_path, "renameat-old");
+  caml_unix_check_path(v_new_path, "renameat-new");
+  old_path = caml_stat_alloc(old_path_len + new_path_len + 2);
```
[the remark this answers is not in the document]
talex5: > I couldn't find any guarantees that it's safe to assume this.

https://v2.ocaml.org/manual/intfc.html#ss:c-block-access says:
> String_val(v) returns a pointer to the first byte of the string v, with type const char *. This pointer is a valid C string: there is a null byte after the last byte in the string.

BTW, the way it does this is quite interesting. The block header just stores the number of words, so for strings it uses the last byte in the block as the number of `'\0'` padding bytes, the first of which acts as the C terminator. If there are no padding bytes, then the "zero padding bytes" byte acts as the terminator!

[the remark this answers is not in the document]
talex5: > I'd be more comfortable with two calls to `caml_stat_strdup`

Then if the second one fails we have to remember to free the first one. With a single allocation, we can just let it raise an exception and forget about it.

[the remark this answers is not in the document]
haesbaert: > > I couldn't find any guarantees that it's safe to assume this.
> 
> https://v2.ocaml.org/manual/intfc.html#ss:c-block-access says:
> 
> > String_val(v) returns a pointer to the first byte of the string v, with type const char *. This pointer is a valid C string: there is a null byte after the last byte in the string.
> 
> BTW, the way it does this is quite interesting. The block header just stores the number of words, so for strings it uses the last byte in the block as the number of `'\0'` padding bytes, the first of which acts as the C terminator. If there are no padding bytes, then the "zero padding bytes" byte acts as the terminator!



> > I couldn't find any guarantees that it's safe to assume this.
> 
> https://v2.ocaml.org/manual/intfc.html#ss:c-block-access says:
> 

Seems I have to improve my search manual foo, I was Ctrl+f String_val and somehow didn't catch that.

> > String_val(v) returns a pointer to the first byte of the string v, with type const char *. This pointer is a valid C string: there is a null byte after the last byte in the string.
> 
> BTW, the way it does this is quite interesting. The block header just stores the number of words, so for strings it uses the last byte in the block as the number of `'\0'` padding bytes, the first of which acts as the C terminator. If there are no padding bytes, then the "zero padding bytes" byte acts as the terminator!

lib_eio_posix/test/poll.md
talex5: No, this was necessary for epoll; I've added a comment explaining what this test is for.

eio_main.opam
talex5: They're the ones from `opam var`, so it doesn't look like it.

lib_eio_posix/flow.ml
talex5: Yes; feel free to open a separate issue for that (it affects all backends).

lib_eio_posix/fd.ml
talex5: Lwt does the same thing (waits and then does the read): https://github.com/ocsigen/lwt/blob/cc05e2bda6c34126a3fd8d150ee7cddb3b8a440b/src/unix/lwt_unix.cppo.ml#L656

But it does it in a systhread, so it's not a big problem. We can do the same once eio_posix uses systhreads (after this initial PR is merged).

polytypic:
There was mention that some of this might/should also work on Windows.  For testing on Windows, there is an example of how to use setup-ocaml to setup a working 5.0 Windows CI [here](https://github.com/dra27/ocaml5-ci-demo).

lib_eio_posix/eio_posix.ml
```diff
@@ -0,0 +1,48 @@
...
+    method stdin = (Flow.of_fd Low_level.Fd.stdin :> <Eio.Flow.source; Eio_unix.unix_fd>)
+    method stdout = (Flow.of_fd Low_level.Fd.stdout :> <Eio.Flow.sink; Eio_unix.unix_fd>)
+    method stderr = (Flow.of_fd Low_level.Fd.stderr :> <Eio.Flow.sink; Eio_unix.unix_fd>)
```
polytypic, line 39: Is it intentional / desirable that each call of stdin/stdout/stderr allocates a new flow object?

lib_eio_posix/eio_posix_stubs.c
```diff
@@ -0,0 +1,178 @@
...
+
+  r = preadv(Int_val(v_fd), iov, n_bufs, Int63_val(v_offset));
+  if (r < 0) uerror("preadv", Nothing);
+
+  CAMLreturn(Val_long(r));
+}
+
+CAMLprim value caml_eio_posix_pwritev(value v_fd, value v_bufs, value v_offset) {
+  CAMLparam2(v_bufs, v_offset);
+  ssize_t r;
```
polytypic, line 109: Hmm… Is there a reason to prefer ancient C style with declarations at the beginning of a block?

I would recommend against that unless there is some really good reason for it.

lib_eio_posix/eio_posix.ml
talex5: The wrapper is stateless (essentially it's just a tuple `(flow_ops, fd)`), but I guess it would be more logical to return the same one each time. Ideally we'd use `lazy` here, but that doesn't work on multicore.

I've pushed a commit that creates the wrappers once in all cases.

lib_eio_posix/eio_posix_stubs.c
haesbaert: Historically C99 declarations didn't get that much adoption in the open source world (especially the UNIX world) because portability was important, I don't think this holds in 2023 and I'd be surprised if there is any issue. Having said that, I dislike declarations in the middle of a block, I like the discipline of having to declare beforehand.

lib_eio_posix/low_level.ml
```diff
@@ -0,0 +1,200 @@
...
+let read fd buf start len =
+  if Fd.is_blocking fd then await_readable fd;
+  Fd.use_exn "read" fd @@ fun fd ->
+  do_nonblocking Read (fun fd -> Unix.read fd buf start len) fd
+
+let write fd buf start len =
+  if Fd.is_blocking fd then await_writable fd;
+  Fd.use_exn "write" fd @@ fun fd ->
+  do_nonblocking Write (fun fd -> Unix.write fd buf start len) fd
+
```
haesbaert, line 46: Did you consider starvation ?
If you have a fiber that always reads/writes and it's always available, since now they don't even trap into the scheduler they can starve everything else.
In Oslo I was build an IO greed mechanism, basically you have to force them to yield either at every IO or at some other heuristic.

lib_eio_posix/low_level.ml
```diff
@@ -0,0 +1,200 @@
...
+let await_writable fd =
+  Fd.use_exn "await_writable" fd @@ fun fd ->
+  Sched.enter @@ fun t k ->
+  Sched.await_writable t k fd
+
+let rec do_nonblocking ty fn fd =
+  try fn fd with
+  | Unix.Unix_error (EINTR, _, _) -> do_nonblocking ty fn fd    (* Just in case *)
+  | Unix.Unix_error((EAGAIN | EWOULDBLOCK | EINPROGRESS), _, _) ->
```
haesbaert, line 29: hmm I don't think EINPROGRESS should be here, is it just defensive programming or I'm missing something else ?

lib_eio_posix/low_level.ml
talex5: Yeah, it's one of the "Other things to work on after merging this initial support" items:

> Ensure we yield on IO to prevent starvation.

I want to add the systhread stuff first, because that will fix some of the places by itself.

haesbaert: :+1:

lib_eio_posix/low_level.ml
talex5: I think you're right - it's only useful for `connect`, which has its own code for this.

lib_eio_posix/eio_posix_stubs.c
talex5: Yeah, let's keep style preferences out of the review (I do prefer the newer style, but most other OCaml code seems to be written this way).

haesbaert (approved):
I think this all looks fine.

lib_eio_posix/eio_posix_stubs.c
polytypic: The reason C code used to be written that way was that it was required.

Nowadays C supports having declarations anywhere that a statement can be used.  This allows one to make it so that all variables are initialized with a proper value.  It reduces the chance of making programming errors such as using uninitialized variables or having type mismatches (variable declared with one type then later initialized with another).  It also promotes the use of more variables (rather than e.g. reusing a variable for multiple different purposes) and even using variables as in FP such that they are only initialized once and never written, which can further help to avoid programming errors.  It also makes it easier to read the code as the scopes of variables can be narrower and be seen more clearly.

haesbaert: If we wanna catch more errors, which is always good, I'd suggest adding proper warnings to the compilation flags, the ones that dune select are absurdly spartan (as it's inherited from ocamlc_cflags, we basically build without any warning enabled, standard flags are these:
`ocamlc_cflags: -O2 -fno-strict-aliasing -fwrapv -pthread -fPIC`

PR #449, opened 2023-02-24
buf_read : add bigstring version of take_while/take_while1

bikallem:
This PR adds the bigstring version of `take_while/take_while1` so that we don't allocate strings if we are primarily only working with `Cstruct.t`

### DIFF
--- a/lib_eio/buf_read.ml
+++ b/lib_eio/buf_read.ml
@@ -297,19 +297,23 @@ let count_while p t =
   try aux 0
   with End_of_file -> t.len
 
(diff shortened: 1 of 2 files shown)


talex5 (changes requested):
`Cstruct.of_bigarray` reuses the bigarray's storage, which is no good (it's already marked as "consumed" by the time the function returns).

Also it shouldn't be called "bigstring" if it's actually a cstruct.

How are you intending to use this?

bikallem:
Closing this since I no longer need this.

PR #457, opened 2023-03-08
Use dune.3.7.0

patricoferris:
This is a PR to show the diff needed to bump everything to `dune.3.7.0` in order to get #405 working. Looks like:

 - Luv (even if it is on its way out) `File`s never make use of the `close_unix` param 
 - The mock backend doesn't need the runq to be mutable
 - A `Buf_write.t` has a CTF id which we would presumably like to use one day but at the moment don't so I've removed it (perhaps suppressing the warning there might be useful and ultimately actually using it would be good too).

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,4 +1,4 @@
-(lang dune 3.0)
+(lang dune 3.7)
 (name eio)
 (formatting disabled)
 (generate_opam_files true)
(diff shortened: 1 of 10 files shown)


talex5 (approved):
Thanks!

PR #460, opened 2023-03-10
Add Fiber.fork_seq

talex5:
This is useful for writing generator functions. e.g.

```ocaml
Switch.run @@ fun sw ->
let seq = Fiber.fork_seq ~sw (fun yield ->
    for i = 1 to 3 do
      traceln "Yielding %d" i;
      yield i
    done
  ) in
Seq.iter (traceln "Got: %d") seq
```

### DIFF
--- a/lib_eio/core/eio__core.mli
+++ b/lib_eio/core/eio__core.mli
@@ -250,6 +250,36 @@ module Fiber : sig
       This is just a convenience wrapper around {!fork}.
       If [fn] raises an exception then the promise is resolved to the error, but [sw] is not failed. *)
 
+  val fork_seq : sw:Switch.t -> (('a -> unit) -> unit) -> 'a Seq.t
+  (** [fork_seq ~sw fn] creates (but does not start) a new fiber to run [fn yield].
+
+      Requesting the next item from the returned sequence resumes the fiber until it
+      calls [yield x], using [x] value as the next item in the sequence. If [fn]
+      returns without producing a value then the result is {!Seq.Nil} (end-of-sequence).
+
+      The returned sequence can be consumed safely from another domain.
+      [fn] itself always runs in the domain that called [fork_seq].
+
+      Example:
+      {[
+        Switch.run @@ fun sw ->
+        let seq = Fiber.fork_seq ~sw (fun yield ->
+            for i = 1 to 3 do
+              traceln "Yielding %d" i;
+              yield i
+            done
+          ) in
+        Seq.iter (traceln "Got: %d") seq
+      ]}
+
+      If [fn] raises an exception then the consumer receives it.
+      If the consumer cancels while awaiting a value, the producer is cancelled when
+      it next calls [yield].
+      It is an error to request two items at once, or to request items out of sequence.
+
+      @param sw When the switch finishes, the fiber is cancelled (if still running).
+                Attempting to read from the sequence after this raises an exception. *)
+
   val fork_daemon : sw:Switch.t -> (unit -> [`Stop_daemon]) -> unit
   (** [fork_daemon] is like {!fork} except that instead of waiting for the fiber to finish,
       the switch will cancel it once all non-daemon fibers are done.
--- a/lib_eio/core/fiber.ml
+++ b/lib_eio/core/fiber.ml
@@ -1,3 +1,5 @@
+[@@@alert "-unstable"]
+
 type _ Effect.t += Fork : Cancel.fiber_context * (unit -> unit) -> unit Effect.t
 
 let yield () =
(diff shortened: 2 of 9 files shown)


lib_eio/core/eio__core.mli
```diff
@@ -250,6 +250,36 @@ module Fiber : sig
       This is just a convenience wrapper around {!fork}.
       If [fn] raises an exception then the promise is resolved to the error, but [sw] is not failed. *)
 
+  val fork_seq : sw:Switch.t -> (('a -> unit) -> unit) -> 'a Seq.t
+  (** [fork_seq ~sw fn] creates (but does not start) a new fiber to run [fn yield].
+
+      Requesting the next item from the returned sequence resumes the fiber until it
+      calls [yield x], using [x] value as the next item in the sequence. If [fn]
```
avsm, line 257: I like this function, but is `yield` the best name for the iterator function within the closure? It's a little confusing since it's nothing like `Fibre.yield` really.  Something like `push` or `dispense` (the Seq terminology) might be clearer as the function name in the example.

lib_eio/core/eio__core.mli
```diff
@@ -250,6 +250,36 @@ module Fiber : sig
...
+              yield i
+            done
+          ) in
+        Seq.iter (traceln "Got: %d") seq
+      ]}
+
+      If [fn] raises an exception then the consumer receives it.
+      If the consumer cancels while awaiting a value, the producer is cancelled when
+      it next calls [yield].
+      It is an error to request two items at once, or to request items out of sequence.
```
avsm, line 278: Worth specifying the exact exception raised here? Not clear from a quick read of the diff if Failure or Invalid_argument (or either) might be raised on this error. Ideally it would be Invalid_argument I think.

lib_eio/core/fiber.ml
```diff
@@ -284,3 +286,108 @@ let with_binding var value fn =
...
+   while suspending itself. The consumer passes in its own [enqueue] function. They run alternatively
+   like this, switching between the [Ready] and [Running] states.
+
+   To finish, the coroutine fiber can set the state to [Finished] or [Failed],
+   or the client can set the state to [Client_cancelled].
+*)
+
+(* Note: we could easily generalise this to [('in, 'out) coroutine] if that was useful. *)
+type 'out coroutine =
+  [ `Init
```
avsm, line 303: Wouldn't this be slightly more efficient without using poly variants and just normal ones? Not using subtyping anywhere I think.

lib_eio/core/eio__core.mli
talex5: I think `yield` is the standard term here (see https://en.wikipedia.org/wiki/Generator_(computer_programming)).

e.g. in Python:

```python
def count():
    for i in range(1, 4):
        yield i

print(list(count()))
```

It's `yield` in JavaScript too: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield

`yield ()` is a bit like `Fiber.yield ()` in that it suspends the caller. The difference is that `yield` is yielding to a particular consumer, who controls when it resumes, whereas `Fiber.yield ()` immediately joins the run queue.

lib_eio/core/eio__core.mli
talex5: There isn't necessarily an exception. It uses `Seq.of_dispenser`, which doesn't check whether you use it in order or not. If you happen to make a second call before the first has finished then you get `Invalid_argument`, but you can't rely on that because it might have finished before that and will then work. So, it's just a programming error.

lib_eio/core/eio__core.mli
avsm: I guess I'm dissatisfied that we have an entirely different terminology ("dispenser") in the Seq module, so we're not being consistent _within_ the OCaml/eio documentation. And the use of `yield` being overloaded (both control flow surrender, and argument-driven resumption) in other languages is a perennial source of confusion, so I'm not sure that's a good guideline to follow.

I just wanted to voice my vague angst here; I don't think I have a clearer alternative apart from "dispense".

lib_eio/core/fiber.ml
avsm: Ah yes, I'd missed that case of `enqueue`; thanks!

Minor nit: iirc there's a modest efficiency benefit in pattern matching with normal variants as well, since the tags are more predictable. E.g.

```
type t =
| Alice | Bob | Charlie | Dan

let fn1 f = function
  | Alice -> 1
  | Bob -> 2
  | Charlie -> 3
  | Dan -> 4

let fn2 f = function
  |`Alice -> 1
  |`Bob -> 2
  |`Charlie -> 3
  |`Dan -> 4
```
gives

```
% ocamlc -dlambda t.ml
(setglobal T!
  (let
    (fn1/272 =
       (function f/274 param/275[int] : int
         (switch* param/275
          case int 0: 1
          case int 1: 2
          case int 2: 3
          case int 3: 4))
     fn2/276 =
       (function f/278 param/279[int] : int
         (if (>= param/279 482771474) (if (>= param/279 884917024) 1 3)
           (if (>= param/279 3403313) 4 2))))
    (makeblock 0 fn1/272 fn2/276)))
```

talex5:
```ocaml
[@@@warning "-A"]

type t =
| Alice | Bob | Charlie | Dan

let fn1 = function
  | Alice -> 1
  | Bob -> 2
  | Charlie -> 3
  | Dan -> 4

let fn2 = function
  |`Alice -> 1
  |`Bob -> 2
  |`Charlie -> 3
  |`Dan -> 4

let fn1 = Sys.opaque_identity fn1
let fn2 = Sys.opaque_identity fn2

let time fn =
  let t0 = Unix.gettimeofday () in
  for i = 1 to 100_000 do
    fn ()
  done;
  let t1 = Unix.gettimeofday () in
  Printf.printf "%.3f ms\n" (1000. *. (t1 -. t0))

let () =
  print_endline "";
  time (fun () -> ignore (fn1 Alice));
  time (fun () -> ignore (fn2 `Alice));
  time (fun () -> ignore (fn1 Alice));
  time (fun () -> ignore (fn2 `Alice));
```
Any difference seems like noise:
```
0.359 ms
0.349 ms
0.353 ms
0.345 ms
```

avsm: Indeed, you won't notice it much with small type definitions -- it's only an issue when the pattern matching code in the compiler can't generate an efficient jump table for a large pattern match. (I've run into this when using OCaml code generators in the past, when it's convenient for the generator to just emit polymorphic variants, but going the extra mile for normal variants makes a difference).

PR #461, opened 2023-03-15
eio_posix: initial support for subprocesses

talex5:
To spawn a child executable on Unix, the parent `fork`s a copy of itself, then has the child copy set up the environment and `execve` the new program.

However, we cannot run any OCaml code in the forked child process. This is because `fork` only duplicates its own domain. To the child, it appears that all other domains have stopped responding and if it tries to e.g. perform a GC then the child process will hang.

Therefore, the `fork` and all child actions need to be written in C. To keep this flexible, this PR adds a new `fork_action` type and lets the user provide a list of such actions to perform.

Some of this support is in Eio_unix as it should be possible to share it with Eio_linux.

At present, there are three actions (`chdir`, `fchdir` and `execve`). It should be fairly easy to plug in new ones (e.g. for renumbering file descriptors, dropping privileges, execveat, etc). Different backends may provide different functions.

In future, it might be good to support `vfork` too, as well as `clone3` on Linux. The SIGCHLD stuff could also be massively simplified on systems with process descriptors, but I think that's only Linux and FreeBSD at the moment (and Eio_linux can do its own thing anyway).


lib_eio_posix/eio_posix_stubs.c
```diff
@@ -185,3 +187,17 @@ CAMLprim value caml_eio_posix_renameat(value v_old_fd, value v_old_path, value v
...
+  pid_t child_pid;
+
+  child_pid = fork();
+  if (child_pid == 0) {
+    eio_unix_run_fork_actions(Int_val(v_errors), v_actions);
+  } else if (child_pid < 0) {
+    uerror("fork", Nothing);
+  }
+
+  CAMLreturn(Val_long(child_pid));
```
haesbaert, line 202: I'm surprised this holds, since a pid_t might not fit on an int. The standard library does the same though.

lib_eio_posix/eio_posix_stubs.c
```diff
@@ -185,3 +187,17 @@ CAMLprim value caml_eio_posix_renameat(value v_old_fd, value v_old_path, value v
...
 }
+
+CAMLprim value caml_eio_posix_spawn(value v_errors, value v_actions) {
+  CAMLparam1(v_actions);
+  pid_t child_pid;
+
+  child_pid = fork();
+  if (child_pid == 0) {
+    eio_unix_run_fork_actions(Int_val(v_errors), v_actions);
+  } else if (child_pid < 0) {
```
haesbaert, line 198:
```suggestion
  } else if (child_pid == -1) {
```

haesbaert: Exactly, PIDs can't be negative and fork cannot return anything below -1.

lib_eio_posix/eio_posix_stubs.c
haesbaert: Interesting, I remember Stevens mentioning that the correct practice whas to for example cast a pid_t to (long long) in order to print it safely, since it was impossible to know its size.

lib_eio/unix/fork_action.c
```diff
@@ -0,0 +1,100 @@
...
+
+void eio_unix_fork_error(int fd, char *fn, char *buf) {
+  try_write_all(fd, fn);
+  try_write_all(fd, ": ");
+  try_write_all(fd, buf);
+}
+
+static char **make_string_array(int errors, value v_array) {
+  int n = Wosize_val(v_array);
+  char **c = malloc(sizeof(char *) * (n + 1));
```
haesbaert, line 47:
```suggestion
  char **c = calloc(sizeof(char *), (n + 1));
```
Unlikely to ever overflow, but would avoid the idiom: https://undeadly.org/cgi?action=article&sid=20060330071917

talex5: Yes, it can only overflow if the input array is already impossibly large, but makes sense to avoid this in general.

lib_eio_posix/eio_posix_stubs.c
talex5: If it never returns anything below -1 then this code change makes no difference. And if it does, then the original version was safer.

haesbaert: it was just more "precise" :smile:

patricoferris (commented):
Looks good! I did a little bit of stress testing on macOS and nothing seemed to break! Would standard output redirection be another action?

lib_eio_posix/eio_posix.mli
```diff
@@ -22,3 +22,5 @@ val run : (stdenv -> 'a) -> 'a
 
 module Low_level = Low_level
 (** Low-level API for making POSIX calls directly. *)
+
+module Process = Process
```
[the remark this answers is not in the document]
talex5: Yes, this was a bit of a hack because Process depends on Low_level. I've moved it now, by making Children a separate file instead.

lib_eio_posix/process.mli
```diff
@@ -0,0 +1,40 @@
...
+  val chdir : string -> t
+  (** [chdir path] changes directory to [path]. *)
+
+  val fchdir : Fd.t -> t
+  (** [fchdir dir] changes directory to [dir]. *)
+end
+
+val spawn : sw:Switch.t -> Fork_action.t list -> t
+(** [spawn ~sw actions] forks a child process, which executes [actions].
+    The last action should be {!Fork_action.execve}.
```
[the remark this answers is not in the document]
talex5: I've added a comment to `execve` that it only makes sense to put it last.

lib_eio_posix/process.ml
```diff
@@ -0,0 +1,158 @@
...
+  let rec reap () =
+    Atomic.set need_wait false;
+    match Unix.(waitpid [WNOHANG] (-1)) with
+    | pid, status -> report_child_status pid status; reap ()
+    | exception Unix.Unix_error (EINTR, _, _) -> reap ()
+    | exception Unix.Unix_error (ECHILD, _, _) ->  ()
+
+  let rec reap_nonblocking () =
+    if Mutex.try_lock lock then (
+      reap ();
```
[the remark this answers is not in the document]
talex5: I hope not. It's not clear whether it would be safe to release the mutex in that case, as we're probably in a bad state anyway.

lib_eio/unix/fork_action.c
```diff
@@ -0,0 +1,100 @@
...
+  }
+  _exit(1);
+}
+
+static void try_write_all(int fd, char *buf) {
+  int len = strlen(buf);
+  while (len > 0) {
+    int wrote = write(fd, buf, len);
+
+    if (wrote <= 0)
```
avsm, line 31: it's worth checking for EINTR here isn't it? Reasonable chance that if there's an error from a forked child, that some signal is coming back to the parent process and interrupting things.

talex5: This code runs in the child. Possibly we should be masking signals there though.

talex5:
> Would standard output redirection be another action?

I've made a branch with some experimental support for inheriting FDs here: https://github.com/talex5/eio/compare/spawn...talex5:eio:inherit-fds?expand=1

RyanGibb:
This allows fixing the issue described at https://github.com/ocaml-multicore/eio/pull/435#issuecomment-1466833565 by writing custom fork actions wrapping `dup2` and `ioctl`:  https://github.com/RyanGibb/ocaml-exec-shell/compare/eedf2b...main#diff-5f310ca7290aedf0d4ea03186711c91b09aa4f211b0c4f3956260ea10ab98c18

How would you feel about exposing this to and documenting for users? (I.e. not putting it behind `Eio_unix.Private`)

RyanGibb:
> If we want to make this public, it will need a bit of work to put the public header files where other packages can find them.

That makes sense, I did have to duplicate some declarations and definitions from `fork_action.h`.

> we could just add `setsid` and setting the tty actions to Eio itself.

I'll try and create a separate PR doing that.

PR #465, opened 2023-03-21
Prepare to split eio_linux into separate files

talex5:
The goal here is to allow the code to be split into multiple files more easily, and to make it more similar to the structure of eio_posix.

- Store the configuration parameters in a record.
- Rename `run` as `run_event_loop` and move `stdenv` out of it.
- Use `Multiple` exception (to minimise use of logging).
- Reduce the number of effects used (`enter` works for lots of things).
- Split out core effects (that only depend on the scheduler) from the extras.

To make review easier, this PR doesn't actually do the split. I'll make another PR for that later.

I tried the `bench_fd` and `bench_yield` benchmarks before and after and there is no noticeable difference.

### DIFF
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -599,29 +599,27 @@ module Low_level = struct
     let result = enter enqueue_noop in
     if result <> 0 then raise (unclassified_error (Eio_unix.Unix_error (Uring.error_of_errno result, "noop", "")))
 
-  type _ Effect.t += Sleep_until : Mtime.t -> unit Effect.t
-  let sleep_until d =
-    Effect.perform (Sleep_until d)
+  let sleep_until time =
+    enter @@ fun t k ->
+    let job = Zzz.add t.sleep_q time k in
+    Fiber_context.set_cancel_fn k.fiber (fun ex ->
+        Zzz.remove t.sleep_q job;
+        enqueue_failed_thread t k ex
+      )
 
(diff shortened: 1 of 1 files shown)


lib_eio_linux/eio_linux.ml
```diff
@@ -1484,15 +1456,47 @@ let rec run : type a.
...
+          let w = (flow (FD.of_unix ~sw ~seekable:false ~close_unix:true w) :> <Eio.Flow.sink; Eio.Flow.close; Eio_unix.unix_fd>) in
+          continue k (r, w)
+        )
+      | _ -> None
+  } in
+  run_sched ~extra_effects st main arg
+
 let run ?queue_depth ?n_blocks ?block_size ?polling_timeout ?fallback main =
+  let config = config ?queue_depth ?n_blocks ?block_size ?polling_timeout () in
+  let stdenv = stdenv ~run_event_loop:(run_event_loop ?fallback:None config) in
```
avsm, line 1499: Why's the `:None` needed after the `?fallback` here?

avsm:
This all looks good to me after eyeballing the changes (not run them). Especially the reduction on reliance on `Logs`.

Might it be worth also considering the role of `eio_utils` when harmonising Eio_linux and Eio_posix? There's a slightly odd collection of data structures and the Suspended thread definition in there.  We can't quite pull out Zzz into a separate since it depends on Suspended...

talex5:
> Might it be worth also considering the role of eio_utils when harmonising Eio_linux and Eio_posix?

The idea of eio.utils is that it's for code shared between backends but that doesn't depend on Unix.

PR #466, opened 2023-03-21
eio_linux: split into multiple files

talex5:
This is the follow-up to #465. It's just moving code around.


avsm (approved):

talex5:
Running that test in a loop, it also fails on `main`.

PR #468, opened 2023-03-21
Condition: fix the example in the docstring

avsm:
The interface changed in 49c22696d50ed20b9ff68c1b1bc6804f85ab99d6 to make Mutex.t a required argument to Condition.await


talex5 (approved):
Thanks!

PR #469, opened 2023-03-22
eio_linux: drain ring on exit

### LINKED ISSUE #467: eio_linux: race cancelling jobs and exiting at the same time
When the main loop exits, it checks that no uring operations are still in progress. Normally this will be the case because we don't allow a fiber to resume until its operation is complete, and the event loop doesn't exit until all fibers have done so.

However, cancel operations don't stop and wait. So I think this can happen:

1. Fiber A starts an operation.
2. Fiber B asks to cancel it and finishes.
3. Fiber A's operation completes by itself and fiber A finishes.
4. We try to end the loop, but get an error because the cancel operation hasn't yet been processed.

I think this is the cause of https://github.com/ocaml-multicore/eio/pull/466#issuecomment-1478028998.

We should probably just wait for cancel operations to complete the same way we do for other operations.

This issue was previously hidden because the exception was reported using `Log.warn`, which silently drops messages by default. #465 changed this to raise an exception instead.

talex5:
Normally, all operations should have finished by the time we exit because we don't exit until all fibers have finished, and a fiber can't finish while an operation is in progress. However, this is not the case for cancellation operations, which may still be active in rare cases. So drain any remaining CQEs at exit.

Other options here are to allow cancellation operations to block (which they used to do, but that caused other problems), or to have the main operation wait for its cancellation too (but that's tricky and affects the fast-path).

Fixes #467.

### DIFF
--- a/lib_eio_linux/sched.ml
+++ b/lib_eio_linux/sched.ml
@@ -498,6 +498,26 @@ let config ?(queue_depth=64) ?n_blocks ?(block_size=4096) ?polling_timeout () =
 
 external eio_eventfd : int -> Unix.file_descr = "caml_eio_eventfd"
 
+let exit_uring uring =
+  (* Normally, all operations should have finished by the time we exit because
+     we don't exit until all fibers have finished, and a fiber can't finish while
+     an operation is in progress. However, this is not the case for cancellation
+     operations, which may still be active in rare cases. Drain them here. *)
+  let rec aux errors =
+    if Uring.active_ops uring = 0 then errors
+    else (
+      match Uring.wait ~timeout:1.0 uring with
+      | None -> errors
+      | Some { data = runnable; result = _ } ->
+        match runnable with
+        | Cancel_job -> aux errors
+        | _ -> aux (errors + 1)
+    )
+  in
+  let unexpected_jobs = aux 0 in
+  Uring.exit uring;
+  if unexpected_jobs > 0 then Fmt.failwith "%d spare CQEs found on uring at exit!" unexpected_jobs
+
 let no_fallback (`Msg msg) = failwith msg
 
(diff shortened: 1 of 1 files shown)


lib_eio_linux/sched.ml
```diff
@@ -498,6 +498,26 @@ let config ?(queue_depth=64) ?n_blocks ?(block_size=4096) ?polling_timeout () =
...
 
+let exit_uring uring =
+  (* Normally, all operations should have finished by the time we exit because
+     we don't exit until all fibers have finished, and a fiber can't finish while
+     an operation is in progress. However, this is not the case for cancellation
+     operations, which may still be active in rare cases. Drain them here. *)
+  let rec aux errors =
+    if Uring.active_ops uring = 0 then errors
+    else (
+      match Uring.wait ~timeout:1.0 uring with
```
avsm, line 509: I'm not so sure about a 1s timeout here.  I could imagine an inflight operation taking longer than that (slow disk or net).  Is there any use in having a timeout here at all? It might be better just to leave the process hanging if there's something weird going on with Uring requests not completing.  It won't be deadlocked, since in theory the kernel can unwedge the process by pushing a completion event...

avsm: I'm just wary of magic timeout numbers (the 1s).   Looping indefinitely with a log seems better. In a stress test, it's kind of hard to spot log entries with errors (although there would be a non-zero exit code here), but quite easy to spot hanging processes taking up all the room on the machine...

talex5:
Closing this in favour of #470. The main loop isn't supposed to finish until `active_ops` is zero, but it exits early if the root fiber raises an exception, which is wrong. Once that's fixed, this PR is unnecessary.

PR #470, opened 2023-03-23
eio_linux: wait for uring to finish before exiting

### LINKED ISSUE #467: eio_linux: race cancelling jobs and exiting at the same time

talex5:
If a fiber raises an exception then the mainloop exits immediately. This normally indicates a bug in Eio, since spawned fibers have exception handlers that pass the error to the switch instead. However, that wasn't the case for the root fiber.

If the root fiber raised an exception after submitting a cancel operation then we would try to exit the ring too early, resulting in:

    Invalid_argument("exit: 1 request(s) still active!")

Fixes #467.

### DIFF
--- a/lib_eio_linux/sched.ml
+++ b/lib_eio_linux/sched.ml
@@ -470,15 +470,18 @@ let run ~extra_effects st main arg =
             Switch.on_release sw (fun () ->
                 Fd.close st.eventfd
               );
-            result := Some (
-                Fiber.first
-                  (fun () -> main arg)
-                  (fun () -> monitor_event_fd st)
-              )
+            Fiber.fork_daemon ~sw (fun () -> monitor_event_fd st);
+            match main arg with
+            | x -> result := Some (Ok x)
+            | exception ex ->
+              let bt = Printexc.get_raw_backtrace () in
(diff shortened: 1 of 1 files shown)


avsm (approved):
Looks like the right fix to me.

Slightly inconsistent with how we deal with `None` returns in the posix backend (which is more of a deadlock warning) https://github.com/ocaml-multicore/eio/blob/main/lib_eio_posix/sched.ml#L360

PR #472, opened 2023-03-24
eio_linux: add subprocess support

talex5:
This combines @patricoferris's code from #435 with the new fork actions from #461.

It needs some review and cleaning up, but might be useful if someone wants to have another go at the high-level API (#330), now that all backends support subprocesses.

I wanted to use `clone3`, but for now it's just using plain `fork` combined with `pidfd_open`. We might need to check that Docker's security policy doesn't block `clone3`. The first commit is just moving the `fork` support from `eio_posix` to `eio_unix` to share it.

It uses `pidfd_send_signal` to signal child processes, which requires `#define CAML_INTERNALS`. I think that's fine, but if not we could instead use plain `Unix.kill`, with `Rcfd` ref-counting to ensure we don't signal the wrong process.

This also cleans up the `Eio_unix.Private.Pipe` handling, to match eio_posix. The same should be done for the other handlers.

- [ ] Split out pipe changes to another PR and fix sockets too.
- [ ] Consider not reporting an exception if you signal a process that's already finished.


avsm:
The clone3/seccomp issue was fixed last year (https://github.com/moby/moby/pull/42836) but need to check thats been rolled out into production Docker instances.

talex5:
Hmm, it seems they didn't actually allow `clone3` in Docker; they just changed it to return `ENOSYS` instead of `EPERM`.

PR #473, opened 2023-03-25
Process API Redux

patricoferris:
An updated version of #330 built on top of #472 -- this is more to inform #472 and any other changes we want to make to the low-level process API. I'm not convinced by f7a982f6ec684c135fa0449088d8c4f7402b11be just yet, it was more of an experiment than anything else. 

I've also half-disabled the `libuv` backend just to get it out of the way.

Some bits to do:

 - Expose environment variables in `Eio.Process`
 - *Actually* work out the best way to support passing in flows that are not backed by FDs.

### DIFF
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -16,7 +16,16 @@ let status proc = proc#status
 let signal proc = proc#signal
 
 class virtual mgr = object
-  method virtual spawn : sw:Switch.t -> ?cwd:Fs.dir Path.t -> stdin:Flow.source -> stdout:Flow.sink -> stderr:Flow.sink -> string -> string list -> t
+  method virtual spawn : 'a 'b 'c.
+    sw:Switch.t ->
+    ?cwd:Fs.dir Path.t ->
+    stdin:(#Flow.source as 'a) ->
+    stdout:(#Flow.sink as 'b) ->
+    stderr:(#Flow.sink as 'c) ->
+    string ->
+    string list ->
+    t
 end
 
-let spawn ~sw t ?cwd ~stdin ~stdout ~stderr cmd args = t#spawn ~sw ?cwd ~stdin ~stdout ~stderr cmd args
+let spawn ~sw (t:#mgr) ?cwd ~(stdin:#Flow.source) ~(stdout:#Flow.sink) ~(stderr:#Flow.sink) cmd args = 
+  t#spawn ~sw ?cwd ~stdin ~stdout ~stderr cmd args
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -5,14 +5,33 @@ let pp_status ppf = function
   | Signaled i -> Format.fprintf ppf "Signalled %i" i
   | Stopped i -> Format.fprintf ppf "Stopped %i" i
 
+type Exn.err += E of status
+
+let err e = Exn.create (E e)
+
+let () =
+  Exn.register_pp (fun f -> function
+    | E e ->
+      Fmt.string f "Process ";
+      pp_status f e;
+      true
+    | _ -> false
+  )
+
 class virtual t = object
   method virtual pid : int
-  method virtual exit_status : status
+  method virtual await : status
   method virtual signal : int -> unit
 end
 
 let pid proc = proc#pid
-let exit_status proc = proc#exit_status
+let await proc = proc#await
+
+let await_exn proc =
+  match proc#await with
+  | Exited 0 -> ()
+  | status -> raise (err status)
+
 let signal proc = proc#signal
 
 class virtual mgr = object
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -1,0 +1,22 @@
+type status = Exited of int | Signaled of int | Stopped of int
+
+let pp_status ppf = function
+  | Exited i -> Format.fprintf ppf "Exited %i" i
+  | Signaled i -> Format.fprintf ppf "Signalled %i" i
+  | Stopped i -> Format.fprintf ppf "Stopped %i" i
+
+class virtual t = object
+  method virtual pid : int
+  method virtual status : status
+  method virtual signal : int -> unit
+end
+
+let pid proc = proc#pid
+let status proc = proc#status
+let signal proc = proc#signal
+
+class virtual mgr = object
+  method virtual spawn : sw:Switch.t -> ?cwd:Fs.dir Path.t -> stdin:Flow.source -> stdout:Flow.sink -> stderr:Flow.sink -> string -> string list -> t
+end
+
+let spawn ~sw t ?cwd ~stdin ~stdout ~stderr cmd args = t#spawn ~sw ?cwd ~stdin ~stdout ~stderr cmd args
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -16,16 +16,19 @@ let exit_status proc = proc#exit_status
 let signal proc = proc#signal
 
 class virtual mgr = object
-  method virtual spawn : 'a 'b 'c.
+  method virtual spawn :
     sw:Switch.t ->
     ?cwd:Fs.dir Path.t ->
-    stdin:(#Flow.source as 'a) ->
-    stdout:(#Flow.sink as 'b) ->
-    stderr:(#Flow.sink as 'c) ->
+    stdin:Flow.source ->
+    stdout:Flow.sink ->
+    stderr:Flow.sink ->
     string ->
     string list ->
     t
 end
 
-let spawn ~sw (t:#mgr) ?cwd ~(stdin:#Flow.source) ~(stdout:#Flow.sink) ~(stderr:#Flow.sink) cmd args = 
-  t#spawn ~sw ?cwd ~stdin ~stdout ~stderr cmd args
+let spawn ~sw (t:#mgr) ?cwd ~stdin ~stdout ~stderr cmd args = 
+  t#spawn ~sw ?cwd cmd args
+    ~stdin:(stdin :> Flow.source)
+    ~stdout:(stdout :> Flow.sink)
+    ~stderr:(stderr :> Flow.sink)
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -7,12 +7,12 @@ let pp_status ppf = function
 
 class virtual t = object
   method virtual pid : int
-  method virtual status : status
+  method virtual exit_status : status
   method virtual signal : int -> unit
 end
 
 let pid proc = proc#pid
-let status proc = proc#status
+let exit_status proc = proc#exit_status
 let signal proc = proc#signal
 
 class virtual mgr = object
--- a/lib_eio/process.mli
+++ b/lib_eio/process.mli
@@ -2,18 +2,25 @@ type status = Exited of int | Signaled of int | Stopped of int
 
 val pp_status : Format.formatter -> status -> unit
 
+type Exn.err += E of status
+
+val err : status -> exn
+(** [err e] is [Eio.Exn.create (E e)] *)
+
 class virtual t : object
     method virtual pid : int
-    method virtual exit_status : status
+    method virtual await : status
     method virtual signal : int -> unit
 end
 
 val pid : #t -> int
 (** The process ID *)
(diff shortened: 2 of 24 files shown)


tests/process.md
```diff
@@ -0,0 +1,160 @@
...
+    Switch.run @@ fun sw ->
+    let t = Eio.Process.spawn ~sw ~stdout ~stdin:env#stdin ~stderr:env#stderr process "/usr/bin/echo" [ "echo"; "Hello" ] in
+    Process.status t
+  in
+  match run () with
+    | Exited 0 ->
+      Eio.Path.(with_open_in (fs / filename)) @@ fun flow ->
+      let buff = Buffer.create 128 in
+      Eio.Flow.copy flow (Eio.Flow.buffer_sink buff);
+      Buffer.contents buff
```
talex5, line 77:
```suggestion
      Eio.Path.load (fs / filename)
```

tests/process.md
```diff
@@ -0,0 +1,160 @@
...
+    | _ -> failwith "Subprocess didn't exit cleanly!";;
```
talex5, line 78: We could probably do with a helper function in lib_eio for checking the exit status. `Process.await_exn` or something?

lib_eio_linux/eio_linux.ml
```diff
@@ -286,11 +286,73 @@ let net = object
...
+  | None ->
+    let r, w = Low_level.pipe ~sw in
+    Some (flow r), w
+  | Some fd -> None, fd
+
+let process_mgr = object
+  inherit Eio.Process.mgr
+  
+  method spawn ~sw ?cwd ~(stdin : #Eio.Flow.source) ~(stdout : #Eio.Flow.sink) ~(stderr : #Eio.Flow.sink) prog args = 
+    let chdir = Option.to_list cwd |> List.map (fun (_, s) -> Process.Fork_action.chdir s) in
```
talex5, line 322: You can't ignore the base directory here (`s` is relative to that). It needs to use `fchdir` in that case.

talex5 (commented):
Thanks - looking good so far!

patricoferris:
Thanks for the review @talex5, I see the #472 is merged, I'll get round to the corrections and rebase on the weekend :))

avsm:
The number of users of Eio/Windows is probably zero, given the lack of a stable OCaml 5.x release for it. Perhaps we should just remove it for Eio 0.9 and ease the path for this PR, and focus all efforts on getting the IOCP backend (#125) over the line?

lib_eio/process.mli
```diff
@@ -0,0 +1,39 @@
...
+        string list ->
+        t
+end
+(** A process manager capable of spawning new processes. *)
+
+val spawn : sw:Switch.t -> #mgr -> ?cwd:Fs.dir Path.t -> stdin:#Flow.source -> stdout:#Flow.sink -> stderr:#Flow.sink -> string -> string list -> t
+(** [spawn ~sw mgr ?cwd ~stdin ~stdout ~stderr cmd args] creates a new subprocess that is connected to the
+    switch [sw]. A process will be stopped when the switch is released.
+    
+    You must provide a standard input and outputs that are backed by file descriptors and
```
avsm, line 38:
```suggestion
    [stdin], [stderr] and [stdout] must be backed by file descriptors.
```

lib_eio/process.ml
```diff
@@ -0,0 +1,31 @@
+type status = Exited of int | Signaled of int | Stopped of int
```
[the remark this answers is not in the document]
patricoferris: Would something like 

```ocaml
type exit_status = [ `Exited of int | `Signaled of int ]

type status = [ exit_status | `Stopped of int ] 
```

be useful? If the `Stopped` case can sometimes happen in the exit case the `exit_status` method would still have to handle it. If it can never happen I suppose we can `assert false` it?

lib_eio/process.ml
```diff
@@ -0,0 +1,31 @@
...
+    stdin:(#Flow.source as 'a) ->
+    stdout:(#Flow.sink as 'b) ->
+    stderr:(#Flow.sink as 'c) ->
+    string ->
+    string list ->
+    t
+end
+
+let spawn ~sw (t:#mgr) ?cwd ~(stdin:#Flow.source) ~(stdout:#Flow.sink) ~(stderr:#Flow.sink) cmd args = 
+  t#spawn ~sw ?cwd ~stdin ~stdout ~stderr cmd args
```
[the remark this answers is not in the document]
patricoferris: Ah thanks that's much cleaner https://github.com/ocaml-multicore/eio/pull/473/commits/419c5fb67237d8a5fe21b9c73e9420b2c0095b9e

lib_eio_linux/eio_linux.ml
patricoferris: Hmm, I've given it a quick go in https://github.com/ocaml-multicore/eio/pull/473/commits/9f2a5ed9e89739a243e0518817b16e560278d486 which seems to be working but wasn't sure if that's right (or the cleanest way) to do it

tests/process.md
patricoferris: Good idea -- do you think instead of the `val exit_status` we should provide `val await : t -> exit_status` and `val await_exn : t -> unit` ?

lib_eio/process.mli
patricoferris: Thanks, hopefully with the pipe changes I added in https://github.com/ocaml-multicore/eio/pull/473/commits/d553598aeebf293c7ba8f884a17dc32eaeaafe2e this can be completely removed as we check if there's an FD and if not we pipe into the sources and sinks (that code could do with a review I think)

patricoferris:
> Perhaps we should just remove it for Eio 0.9 and ease the path for this PR, and focus all efforts on getting the IOCP backend (https://github.com/ocaml-multicore/eio/issues/125) over the line?

This sounds sensible. It should probably be removed in a separate PR, for now in https://github.com/ocaml-multicore/eio/pull/473/commits/f4024231a56ce5f2b38e0166498d1aff63e61acc I've brought it back but just stubbed out process support.

patricoferris:
https://github.com/ocaml-multicore/eio/pull/473/commits/5afdbd4ac8c41972204fbc77b98a784787d9a3dc added `Eio.Stdenv.stdio` to allow getting `stdin, stdout, stderr` together in one call. It changed the `exit_status` name to `await` and `await_exn` (perhaps this should be just `wait` and `wait_exn`).

patricoferris:
The `Fiber.fork` calls in `Process.spawn` weren't quite right. In https://github.com/ocaml-multicore/eio/pull/473/commits/d6eec9f235431fb9c8dcd2207c0b56d95b85392f I removed them and this seems to fix the issues, the only concern I have is that the call to `Process.spawn` might yield now depending on what kind of `stdin/stdout` you pass to the call, is that okay ?

lib_eio_posix/process.ml
```diff
@@ -0,0 +1,72 @@
...
+        0, Fd.to_rcfd stdin_fd, `Blocking;
+        1, Fd.to_rcfd stdout_fd, `Blocking;
+        2, Fd.to_rcfd stderr_fd, `Blocking;
+      ];
+      execve prog ~argv:(Array.of_list args) ~env:[||]
+    ] in
+    let with_actions cwd fn = match cwd with
+      | Some ((dir, path) : Eio.Fs.dir Eio.Path.t) -> (
+        match Eio.Generic.probe dir Fs.Posix_dir with
+        | None -> fn actions
```
talex5, line 49: If the cwd isn't a native directory then it should raise an exception instead of ignoring it.

lib_eio_posix/process.ml
```diff
@@ -0,0 +1,72 @@
...
+          fn (Process.Fork_action.fchdir cwd :: actions)
+      )
+      | None -> fn actions
+    in
+    let proc =
+      with_actions cwd @@ fun actions ->
+      process (Process.spawn ~sw actions)
+    in
+    Option.iter (fun stdin_w ->
+      Eio.Flow.copy stdin stdin_w;
```
talex5, line 62: You need to fork off a another fiber to handle this (I see you had it that way earlier and then removed it for some reason).

Failing to fork will work OK for short messages, where the kernel will buffer them, but will deadlock for longer ones. It's also a problem for interactive programs (e.g. hooked up to GUI). Imagine the program wants to output a long set of terms and conditions and then ask for confirmation. Once the stdout pipe is full, it will stop and wait for us to read it, but we're still here waiting for the user to finish sending the input (which they won't do because they haven't seen any output yet).

tests/process.md
```diff
@@ -0,0 +1,181 @@
...
+  let echo = resolve_program "echo" in
+  let p =
+    Eio.Process.spawn proc ~sw ~stdin:env#stdin ~stdout:dst ~stderr:env#stderr echo [ "echo"; "Hello, world" ]
+  in
+  let _ : Process.status = Process.await p in
+  Buffer.contents buf
+- : string = "Hello, world\n"
+```
+
+Changing directory
```
talex5, line 171: Needs a test for the confined case too.

talex5 (commented):
Looking good! We also need support for environment variables (at the moment, it always clears the environment). I think it's fine to make passing an environment optional and default to inheriting the current one.

talex5:
> The Fiber.fork calls in Process.spawn weren't quite right.

What was the problem?

> The number of users of Eio/Windows is probably zero, given the lack of a stable OCaml 5.x release for it. Perhaps we should just remove it for Eio 0.9 and ease the path for this PR, and focus all efforts on getting the IOCP backend (https://github.com/ocaml-multicore/eio/issues/125) over the line?

I'm planning to remove eio_luv right after the 0.9 release. Keeping it in 0.9 might make things slightly easier for people who used it directly and want to move to eio_posix, or who want to compare the performance between them.

lib_eio_posix/process.ml
patricoferris: Very quickly, it seemed that there was a race between the process exiting and the standard input/output being read/written. This very reliably fails in the old fork version: 

```ocaml
open Eio
open Eio.Std

let run env =
  let buf = Buffer.create 128 in
  Switch.run @@ fun sw ->
  let stdout = Flow.buffer_sink buf in
  let child =
    Process.spawn ~sw env#process_mgr ~stdin:env#stdin ~stdout ~stderr:env#stderr "/bin/echo" [ "echo"; "Hello" ]
  in
  Process.await_exn child;
  let contents = Buffer.contents buf in
  if contents = "Hello\n" then () else failwith contents

let () =
  Eio_main.run @@ fun env ->
  for i = 0 to 1000 do
    Eio.traceln "Process number %i" i;
    run env
  done
```

But your comment is right!

talex5: Yes, you need to wait for stdout to be copied to the buffer. Just waiting for the process to exit isn't sufficient. This works (but is easy to get wrong):
```ocaml
let run env =
  let buf = Buffer.create 128 in
  Switch.run (fun sw ->
      let stdout = Flow.buffer_sink buf in
      let child =
        Process.spawn ~sw env#process_mgr ~stdin:env#stdin ~stdout ~stderr:env#stderr "/bin/echo" [ "echo"; "Hello" ]
      in
      Process.await_exn child;
    );
  let contents = Buffer.contents buf in
  if contents = "Hello\n" then () else failwith contents
```
Perhaps we should just expose the pipes and let the user do the reading, as Lwt does? It's a bit of a pain with the types (which is why Lwt needs so many functions). Maybe GADTs would help (so if you pass `~stdout:Pipe` then you can do `Process.stdout child` to get the flow).

Then we can add convenience functions on top (like `popen`) that collect the output for you, rather than hard-coding that behaviour in each backend.

https://docs.python.org/3/library/subprocess.html might be a good source of ideas.

talex5:
I had a go at some more changes in my branch (at https://github.com/talex5/eio/tree/proc-redux). I think there's enough API design work still to do here that it's better to release Eio 0.9 without a cross-platform process API. Then we can spend a few weeks trying different APIs and get something merged for Eio 0.10 instead.

talex5:
Rebased and updated as #499.

patricoferris:
Thanks! I'll review the process PR later today :)) Closing in favour of it!

PR #480, opened 2023-04-04
Mint exclusive IDs

TheLortex:
Another take on https://github.com/ocaml-multicore/eio/pull/368 to generate unique IDs in a multicore context.

The proposal is to encode the domain ID in the generated ID, and have one counter per domain (using domain local storage), so that there's no contention on the ID counter. As it its implemented, it requires knowing the maximum number of IDs. It is currently hardcoded to 128, but the value is not specified. 

An alternative solution for that is to create a fixed array of atomic counters, say of size N, and map domain ID to the counter `ID mod N`. It requires the usage of atomic variables to support the situation where the number of domains is greated than N, but reduces contention as multiple counters are used.

### DIFF
--- a/lib_eio/core/ctf.ml
+++ b/lib_eio/core/ctf.ml
@@ -28,11 +28,24 @@ end
 
 type id = int
 
-let last_id = ref 0
+let id_chunk_size = 1024
+
+let next_id_chunk = Atomic.make 0
+
+let next_id_key =
+  Domain.DLS.new_key (fun () -> Atomic.fetch_and_add next_id_chunk id_chunk_size)
 
 let mint_id () =
-  incr last_id;
-  !last_id
+  let next_id_local = Domain.DLS.get next_id_key in
+  let next_id_local_succ =
+    if ((next_id_local + 1) mod id_chunk_size) = 0 then
+      (* we're out of local IDs *)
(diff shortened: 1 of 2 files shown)


talex5:
This won't work. Although there can only be 128 domains at a time, the IDs keep increasing:

```ocaml
for i = 1 to 130 do
  let d = Domain.spawn ignore in
  Printf.printf "%d\n" (Domain.get_id d :> int)
done;;
```
Prints `128`, `129`, `130`.

We can instead do this the way OCaml allocates object IDs, using blocks of 1024:
https://github.com/ocaml/ocaml/blob/8746939b08d84bafd50630892048ed512fc711ff/runtime/obj.c#L270

TheLortex:
Oh that looks like a nice solution indeed, thanks for the pointer. I'll update the implementation.

TheLortex:
@talex5 I have added an implementation inspired from ocaml object ids. It took the same value of `1024` for the chunk size but it can be adjusted.

talex5 (approved):
Thanks!

(I just renamed `next_id` to `next_id_chunk` (since it's not the next ID now) and got the test to demonstrate the discontinuity.)

TheLortex:
Great, thanks for the improvements.

PR #482, opened 2023-04-05
use raw system call of getrandom for glibc version before 2.25

zenfey:
'getrandom' was added to glibc in version 2.25. Older versions of glibc would fall back to use raw system call to correctly compile eio.

### DIFF
--- a/lib_eio_linux/eio_stubs.c
+++ b/lib_eio_linux/eio_stubs.c
@@ -4,7 +4,9 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/eventfd.h>
+#if __GLIBC__ > 2 || __GLIBC_MINOR__ > 24
 #include <sys/random.h>
+#endif
 #include <sys/syscall.h>
 #include <limits.h>
 #include <errno.h>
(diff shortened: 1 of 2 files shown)


talex5 (approved):
Seems reasonable. Non-glibc systems will use the syscall directly, too.

zenfey:
> Seems reasonable. Non-glibc systems will use the syscall directly, too.

 You're right and I wasn't aware of non-glibc systems. The original purpose of this PR is mainly for eio to build on CentOS 7 with linux kernel ahead of 3.17 with glibc earlier than 2.25. Thanks for review : )

avsm:
This looks reasonable, but need to keep an eye on the musl situation in the future since [vDSO support for getrandom was discussed on lklm last year](https://lore.kernel.org/lkml/YuktqQS7Rb0IbJNh@zx2c4.com/T/).  There's also no downside that I can see to just always using `syscall` here -- the [glibc function](https://github.com/bminor/glibc/blob/master/sysdeps/unix/sysv/linux/getrandom.c) is currently just a single-line invocation of `syscall` as well.

talex5:
Thanks!

PR #487, opened 2023-04-14
Fix Condition.await bug when cancelling

talex5:
`Condition.await` needs to re-acquire the lock when cancelled. However, if another fiber is holding the lock then this operation must wait and would itself get cancelled.

Reported by @polytypic.

It's a little annoying that we have to do a non-cancellable wait during cancellation, since typically we only need the lock so we can call unlock on it again immediately. I did try adding a `wait_until` function that handles all the locking, which allows simplifying this. However, we would need multiple versions of it to cover the various ways of using a lock (`use_rw`, `use_ro`, and `~protected`), which seems a bit messy.

### DIFF
--- a/lib_eio/condition.ml
+++ b/lib_eio/condition.ml
@@ -2,6 +2,9 @@ type t = Broadcast.t
 
 let create () = Broadcast.create ()
 
+let lock_protected m =
+  Cancel.protect (fun () -> Eio_mutex.lock m)
+
 let await_generic ?mutex t =
   match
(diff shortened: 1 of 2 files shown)


polytypic:
The fix seems correct.  On quick look the failures don't seem related?

talex5:
Yes, `[WARNING] Failed to allocate 4096 byte fixed buffer` means the build machine was a bit overloaded. Normally Eio would fall back to using a regular buffer, but that test is specifically for checking that the fixed buffer works.

PR #489, opened 2023-04-17
Document how to use Domainslib from Eio

talex5:
/cc @kayceesrk

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ Eio replaces existing concurrency libraries such as Lwt
   * [Async](#async)
   * [Lwt](#lwt)
   * [Unix and System Threads](#unix-and-system-threads)
+  * [Domainslib](#domainslib)
 * [Best Practices](#best-practices)
   * [Switches](#switches-1)
   * [Casting](#casting)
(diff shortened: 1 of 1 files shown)


polytypic (approved):

talex5:
I've expanded the example a bit to show how to start the main loop, due to some confusion in https://discuss.ocaml.org/t/interaction-between-eio-and-domainslib-unhandled-exceptions/11971.

README.md
```diff
@@ -1578,6 +1579,36 @@ The [Eio_unix][] module provides features for using Eio with OCaml's Unix module
...
+
+For certain compute-intensive tasks it may be useful to send work to a pool of [Domainslib][] worker domains.
+You can resolve an Eio promise from non-Eio domains (or systhreads), which provides an easy way to retrieve the result.
+For example:
+
+<!-- $MDX skip -->
+```ocaml
+let pool = Domainslib.Task.setup_pool ~num_domains:2 ()
+
+let fib n = ... (* Some Domainslib function *)
```
patricoferris, line 1592: Do you think it makes sense to stress that you _cannot_ perform a call that may perform an (Eio) effect in the Domainslib async task because that may be run in a different domain that no longer has the effect handler installed?

talex5: Good idea. I've added a bit of text saying you can mostly only use functions from the appropriate library in each domain.

README.md
```diff
@@ -1578,6 +1579,43 @@ The [Eio_unix][] module provides features for using Eio with OCaml's Unix module
...
+```ocaml
+open Eio.Std
+
+let pool = Domainslib.Task.setup_pool ~num_domains:2 ()
+
+let fib n = ... (* Some Domainslib function *)
+
+let run_in_pool fn x =
+  let result, set_result = Promise.create () in
+  let _ : unit Domainslib.Task.promise = Domainslib.Task.async pool (fun () ->
```
polytypic, line 1598: I initially assumed that the `send` that happens internally with `Task.async` would be safe to call from non-domainslib domains, but as I'm now looking into adding POC domain local await support to domainslib, I realize that this works somewhat accidentally.

When you create the pool

```ocaml
let pool = Domainslib.Task.setup_pool ~num_domains:2 ()
```

the requsted `n=2` domains are started.  Additionally an internal "multi channel" is created with `n+1` capacity.  The `+1` reserves an extra internal "channel" for the domain that calls `Task.run`.

When `Task.async` is called, it looks up the internal channel for the current domain.  If no channel has already been allocated for the current domain, it then allocates a channel for the domain from the reserved `n+1`.  This works as long as `Task.async` is only called from a single additional domain.  If you had two (or more) Eio domains, then only one of those could to call `Task.async`.

talex5: Interesting. But doesn't `Task.run` have the same limitation, then?

polytypic: Yes, I think it has the same limitation.

At any rate, I'm looking into how to potententially change that to be more domain safe.

PR #490, opened 2023-04-17
Initial windows scheduler

patricoferris:
As suggested in #125, this is just the mock backend scheduler placed into the Windows IOCP backend so we can try to do this incrementally. I'm seeing if I can get a reliable (if horribly hacked together) Github CI setup for Windows and OCaml 5 too.

### DIFF
--- a/lib_eio_windows/eio_windows.ml
+++ b/lib_eio_windows/eio_windows.ml
@@ -1,3 +1,23 @@
-(* Can base this on the eio_posix directory structure.
-   See HACKING.md for instructions on creating a new backend. *)
-let run _main = failwith "TODO: Windows support."
+(* WIP backend for Windows using IOCP *)
+
+let run main =
+   let stdenv = object
+      method stdin = failwith "Not implemented"
+      method stdout = failwith "Not implemented"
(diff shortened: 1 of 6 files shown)


talex5:
Seems like a reasonable place to start, especially if it lets us get Windows CI running.

It might be better to base sched.ml on a simplified version of the Eio_linux one, though (e.g. with `Uring.wait` replaced with a `failwith "TODO"`). Then we'd already have the structure for polling for IO, doing wakeups, etc. (assuming uring is somewhat similar to iocp)

avsm:
Agreed. My concern is that the IOCP bindings still need some work, and I think a slower Windows backend is fine for the eio 1.0, followed by an IOCP version in a subsequent release.

patricoferris:
I have this _nearly_ working and will open a new PR (and close this one once I have), just some uncompilable bits being brought in from `eio.unix` where I need some of the features (e.g. FD stuff) but not others (e.g. fork actions)

talex5:
Ah. I guess most of `fork_action.c` needs wrapping in a `#ifdef` then.

patricoferris:
New PR over at #497

PR #491, opened 2023-04-19
Unify Eio_linux.FD and Eio_posix.Fd as Eio_unix.Fd

talex5:
Now that eio_luv is gone, there is no need for different backends to have different types for wrapped file descriptors. This means we can use wrapped FDs in more places, avoiding races (before, any API shared between backends had to use `Unix.file_descr`).

This also allows e.g. `Eio_unix.pipe` to return resources with an `Fd.t` attached explicitly in the type. It also avoids duplicating the fork actions in each backend.

In the first commit the old APIs are mostly still present (using wrappers defined in terms of the new API), and the tests still use them. The second commit updates the tests to the new API and marks the old APIs as deprecated.

The Linux backend needs to know if FDs are seekable, while the POSIX one wants to know if they're blocking, so the unified version tracks both but only does the query if needed.

More API improvements should be possible as a result of this, such as adding an `Eio_unix.Stdenv.t`, where resources have FDs in their types, but that can be a separate PR. That should also be useful for the subprocess PR, allowing us to inherit generic FDs, not just flows.

### DIFF
--- a/lib_eio/unix/dune
+++ b/lib_eio/unix/dune
@@ -4,5 +4,5 @@
  (foreign_stubs
   (language c)
   (include_dirs include)
-  (names fork_action))
+  (names fork_action stubs))
  (libraries eio unix threads mtime.clock.os))
--- a/lib_eio/unix/eio_unix.ml
+++ b/lib_eio/unix/eio_unix.ml
@@ -1,3 +1,16 @@
+[@@@alert "-unstable"]
+
+module Fd = Fd
+
+module Resource = struct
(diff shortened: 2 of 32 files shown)


talex5:
Will merge this afternoon if there are no objections.

anmonteiro:
What's the best way to get a `Eio.Flow.{source,sink}` from a `Eio_unix.Fd.t`?

talex5:
@anmonteiro note that it's `Eio_unix.Net.import_socket_stream` now.

PR #492, opened 2023-04-19
Document mirage `Ipaddr` conversion

RyanGibb:
Enabled by https://github.com/mirage/ocaml-ipaddr/pull/117

### DIFF
--- a/lib_eio/net.mli
+++ b/lib_eio/net.mli
@@ -67,6 +67,14 @@ module Ipaddr : sig
     'a
   (** [fold ~v4 ~v6 t] is [v4 t] if [t] is an IPv4 address, or [v6 t] if it's an IPv6 address. *)
 
+  (** {2 Interoperability}
+
+  To convert to or from OCaml Unix addresses, use {!Eio_unix.Ipaddr}.
+
+  To interoperate with the {{:https://opam.ocaml.org/packages/ipaddr/} ipaddr} library:
+  - [Ipaddr.to_octets ipaddr_ip |> Eio.Net.Ipaddr.of_raw]
+  - [Ipaddr.of_octets_exn (eio_ip :> string)] *)
+
   val of_raw : string -> v4v6
   (** [of_raw addr] casts [addr] to an IP address.
       @raise Invalid_argument if it is not 4 or 16 bytes long. *)


lib_eio/net.mli
```diff
@@ -27,7 +27,12 @@ type Exn.err += E of error
 val err : error -> exn
 (** [err e] is [Eio.Exn.create (Net e)] *)
 
-(** IP addresses. *)
+(** IP addresses.
+
+  In order to interoperate with the {{:https://opam.ocaml.org/packages/ipaddr/} ipaddr} library we can use it's octet conversion functions.
```
patricoferris, line 32:
```suggestion
  In order to interoperate with the {{:https://ocaml.org/p/ipaddr/} ipaddr} library we can use its octet conversion functions.
```

I think linking to OCaml.org might be preferable and more future proof?

lib_eio/net.mli
```diff
@@ -27,7 +27,12 @@ type Exn.err += E of error
...
+  That is, to convert to an [Eio.Ipaddr] we can use [Ipaddr.to_octets ip |> Eio.Net.Ipaddr.of_raw].
```
patricoferris, line 33:
```suggestion
  That is, to convert to an {! Eio.Net.Ipaddr.t} we can use [Ipaddr.to_octets ip |> Eio.Net.Ipaddr.of_raw].
```

That should link to the `t` in the docs :)) (I think)

patricoferris (commented):
Thanks! This is much cleaner than what we had to do before ^^

talex5 (commented):
I suggest putting it as a section with `of_raw`, e.g.

```ocaml
  (** {2 Interoperability}

      To convert to or from OCaml Unix addresses, use {!Eio_unix.Ipaddr}.

      To interoperate with the {{:https://opam.ocaml.org/packages/ipaddr/} ipaddr} library:
      - [Ipaddr.to_octets ipaddr_ip |> Eio.Net.Ipaddr.of_raw]
      - [Ipaddr.of_octets_exn (eio_ip :> string)] *)

  val of_raw : string -> v4v6
  (** [of_raw addr] casts [addr] to an IP address.
      @raise Invalid_argument if it is not 4 or 16 bytes long. *)
```

lib_eio/net.mli
RyanGibb: Thanks patrick, that makes sense!

lib_eio/net.mli
RyanGibb: Ah yes, I'm not too familiar with OCaml documentation syntax. Thanks!

RyanGibb:
Thanks for the suggestion @talex5, I've applied it now

avsm (approved):

talex5:
Thanks!

PR #494, opened 2023-04-24
Add support for domain local await

polytypic:
This PR adds [domain local await](https://github.com/ocaml-multicore/domain-local-await) support to Eio.  Domain local await allows blocking abstractions to work across different schedulers.

I've also made a similar [PR to Domainslib](https://github.com/ocaml-multicore/domainslib/pull/107).  If and when both Eio and Domainslib merge the support for domain local await, it should be possible to use [upcoming version of kcas](https://github.com/ocaml-multicore/kcas/pull/32) to [easily communicate between Eio and Domainslib domains](https://discuss.ocaml.org/t/interaction-between-eio-and-domainslib-unhandled-exceptions/11971/10?u=polytypic).  Perhaps [lockfree](https://github.com/ocaml-multicore/lockfree/pull/68) could also adopt use of domain local await until we get some official blocking mechanism.


talex5:
Interesting. I guess this doesn't even need any changes in Eio. You could just run any Eio event loop and then wrap Domain_mgr to install this handler in all new domains, right?

lib_eio/core/dla.ml
```diff
@@ -0,0 +1,34 @@
...
+      | `Awaiting b -> Broadcast.resume_all b
+      | _ -> ()
+  and await () =
+    match Atomic.get state with
+    | `Released -> ()
+    | `Awaiting _ -> failwith "await called twice"
+    | (`Cancelable | `Protected) as was ->
+        let b = Broadcast.create () in
+        if Atomic.compare_and_set state was (`Awaiting b) then
+          let resume ctx enqueue =
```
talex5, line 20: `resume` seems like a bad name. This is called when we suspend.

lib_eio/core/dla.ml
```diff
@@ -0,0 +1,34 @@
...
+    if Atomic.get state != `Released then
+      match Atomic.exchange state `Released with
+      | `Awaiting b -> Broadcast.resume_all b
+      | _ -> ()
+  and await () =
+    match Atomic.get state with
+    | `Released -> ()
+    | `Awaiting _ -> failwith "await called twice"
+    | (`Cancelable | `Protected) as was ->
+        let b = Broadcast.create () in
```
talex5, line 18: `Broadcast` seems a bit heavy-weight for this. It allows any number of suspenders, but it looks like this code only ever has at most one (given the `failwith` above). Can't we just use `state` to hold the `enqueue` function?

polytypic: I changed the PR to use domain local await that will hopefully soon be available from opam and simplified the implementation to just use `enqueue`.

polytypic:
> Interesting. I guess this doesn't even need any changes in Eio. You could just run any Eio event loop and then wrap Domain_mgr to install this handler in all new domains, right?

Yes, I think for many schedulers it would be possible to provide domain local await outside of the scheduler by implementing it in terms of the public API of the scheduler and then installing it within a fiber running on the scheduler.  However, for better UX / DX it is preferable to have schedulers provide the support directly.  Of course, DLA is experimental at this stage and it is quite possible that it might be replaced with some official standard blocking mechanism, but I think that for the short term it would be great to be able have various libraries that need blocking to be able to cooperate through it.

lib_eio/core/dla.ml
```diff
@@ -0,0 +1,14 @@
...
+  let release () =
+    if Atomic.get state != `Released then
+      match Atomic.exchange state `Released with
+      | `Awaiting enqueue -> enqueue (Ok ())
+      | _ -> ()
+  and await () =
+    if Atomic.get state != `Released then
+      Suspend.enter @@ fun _ctx enqueue ->
+      if not (Atomic.compare_and_set state `Init (`Awaiting enqueue)) then
+        enqueue (Ok ())
```
talex5, line 12: The cancellation support seems to have gone missing. I would expect something like this:
```suggestion
      Suspend.enter @@ fun ctx enqueue ->
      let awaiting = `Awaiting enqueue in
      if Atomic.compare_and_set state `Init awaiting then (
        Cancel.Fiber_context.set_cancel_fn ctx (fun ex ->
            if Atomic.compare_and_set state awaiting `Released then (
              enqueue (Error ex)
            )
          )
      ) else (
        enqueue (Ok ())
      )
```

polytypic: Hmm... Is the `cancel_fn` always cleared automatically in case the fiber is enqueued normally?

polytypic: LGTM

talex5 (approved):
Looks good (I just squashed and rebased it).

PR #497, opened 2023-04-27
Posix-based windows implementation

patricoferris:
This is a revised version of #490 (as part of #125). As mentioned in the previous PR, for an Eio 1.0 we're looking to have functional support for windows in a rigorous manner. To that end this PR implements a Windows backend using a mostly unchanged `Eio_posix` backend (directly copied over) with the following modifications and points of note:

 - Filesystem operations (mkdir, open etc.) are not currently supported.
 - Subprocess support is not currently supported either.
 - The "polling" mechanism has been changed to use `Unix.select` I started looking into adding a Windows backend to `iomux` then I stumbled across [poll](https://github.com/anuragsoni/poll) and the use of Wepoll. I tried to simply use `WSApoll` but of course that completely doesn't work for anything but sockets, so to minimise the diff and just get something working I went with `Unix.select` which I believe should just work for files and sockets a like and is pretty well tested.
 - I ran into some issues with MDX so I couldn't just copy the tests over from `Eio_main` tests. I'll look into that, it seemed to be something about a dll and flexlink or something.
 - It uses this PR for `Unix_cstruct` https://github.com/mirage/ocaml-cstruct/pull/302 -- if this can't be merged, we could probably vendor the stubs. We still need other functions implemented like `pread` and `pwrite` which is a bit awkward because of the differences in how the FD current offset is changed compared to Unix systems... 

This is an initial draft, I want to go over the Unix.select stuff and of course do more testing. Maybe then a follow-up PR can come in to add FS operations and the other missing pieces.


samwgoldman:
[This issue](https://github.com/ocaml/ocaml/issues/7665) is relevant here. Unix.select works well with sockets on Windows, but the emulation for non-sockets is not 100%.

lib_eio_windows/children.mli
```diff
@@ -0,0 +1,14 @@
+(** Keep track of child processes and respond to SIGCHLD. *)
```
talex5, line 1: It seems unlikely we need this module on Windows.

lib_eio_windows/dune
```diff
@@ -1,5 +1,15 @@
 (library
  (name eio_windows)
  (public_name eio_windows)
+ (library_flags :standard -ccopt -lbcrypt)
  (enabled_if (= %{os_type} "Win32"))
- (libraries eio eio.utils fmt))
+ (foreign_stubs
+  (language c)
+  (flags :standard -D_LARGEFILE64_SOURCE)
+  (include_dirs ../lib_eio/unix/include)
```
talex5, line 9: Probably don't need this.

lib_eio_windows/eio_windows.ml
```diff
@@ -1,3 +1,53 @@
...
+  cwd : Eio.Fs.dir Eio.Path.t;
+  secure_random : Eio.Flow.source;
+  debug : Eio.Debug.t;
+>
+
+let run main =
+  (* SIGPIPE makes no sense in a modern application. *)
+  (* These don't work on Windows. *)
+  (* Sys.(set_signal sigpipe Signal_ignore); *)
+  (* Sys.(set_signal sigchld (Signal_handle (fun (_:int) -> Children.handle_sigchld ()))); *)
```
talex5, line 37: Let's get rid of them then :-)

lib_eio_windows/eio_windows.mli
```diff
@@ -0,0 +1,24 @@
+(** Fallback Eio backend for POSIX systems. *)
```
talex5, line 1: Comment needs updating.

lib_eio_windows/fd.ml
```diff
@@ -0,0 +1,16 @@
+(* Deprecated *)
```
talex5, line 1: No need for deprecated modules in new code.

lib_eio_windows/fd.mli
```diff
@@ -0,0 +1,49 @@
+(** A safe wrapper for {!Unix.file_descr}. *)
```
talex5, line 1: Don't need this file (it's all deprecated).

lib_eio_windows/low_level.ml
```diff
@@ -0,0 +1,282 @@
...
+
+let pipe ~sw =
+  let unix_r, unix_w = Unix.pipe ~cloexec:true () in
+  let r = Fd.of_unix ~sw ~blocking:false ~close_unix:true unix_r in
+  let w = Fd.of_unix ~sw ~blocking:false ~close_unix:true unix_w in
+  Unix.set_nonblock unix_r;
+  Unix.set_nonblock unix_w;
+  r, w
+
+module Process = struct
```
talex5, line 221: I guess we can just delete this module.

lib_eio_windows/sched.ml
```diff
@@ -0,0 +1,364 @@
...
+  let cleanup () =
+    Unix.close eventfd_r;
+    let was_open = Rcfd.close eventfd in
+    assert was_open
+  in
+  let poll = { to_read = FdMap.empty; to_write = FdMap.empty } in
+  let fd_map = Hashtbl.create 10 in
+  let t = { run_q; poll; poll_maxi = (-1); fd_map; eventfd; eventfd_r;
+            active_ops = 0; need_wakeup = Atomic.make false; sleep_q } in
+  let eventfd_ri : int = Obj.magic eventfd_r in
```
talex5, line 244: I don't think file descriptors are ints on Windows.

talex5 (commented):
Thanks! Are some of the examples able to run on Windows (`hello`, or even `net`?).

Deleting everything marked as deprecated would shrink it a bit.

CI is failing to find Eio. Probably just needs `opam pin --with-version=dev`.

patricoferris:
> Thanks! Are some of the examples able to run on Windows (hello, or even net?).

Yep both work :)) And yes, I was a little too copy-paste happy and can remove all the bits that are not needed shortly.

With regards to CI, I'm trying to get it working over on a different branch to avoid noise here https://github.com/patricoferris/eio/actions/runs/4829634872/jobs/8604900630 -- there are just lots of blockers to having OCaml 5 and windows including:

 - Fixes to ocplib-endian https://github.com/OCamlPro/ocplib-endian/pull/26
 - The patches are needed for ocamlbuild https://github.com/ocaml/ocamlbuild/issues/316 (so far I've cheekily installed the duniverse version)
 - Ideally we'd have David's patches upstreamed too https://github.com/dra27/opam-repository/tree/windows-5.0

These are all probably prerequisites before an Eio_windows release?

avsm:
Where's the dependency on ocplib-endian coming from? We removed it from cstruct a while back, and it shouldn't be necessary these days.

lib_eio_windows/sched.ml
avsm: definitely not an integer; https://github.com/ocaml/ocaml/blob/trunk/otherlibs/unix/unixsupport.h#L43

patricoferris: Yep, that's a little embarrassing ^^" Will fix soon

patricoferris:
> Where's the dependency on ocplib-endian coming from?

Crowbar I think https://ocaml.org/p/crowbar/latest

polytypic:
> I ran into some issues with MDX so I couldn't just copy the tests over from Eio_main tests. I'll look into that, it seemed to be something about a dll and flexlink or something.

On Windows OCaml uses [flexdll / flexlink](https://github.com/ocaml/flexdll) for DLLs.  So, on Windows, if you want to be able to load a C library dynamically, e.g. to use it from the OCaml toplevel, utop, or mdx, for example, then that library needs to be compiled (with  `__declspec(dllimport)` annotations where needed) and linked (using `flexlink`) in a special way to support it.

I ran into this problem when trying to run tests with the luv backend.  See here for [an ugly hack](https://github.com/aantron/luv/issues/142#issuecomment-1402144498) with which I managed to actually be able to build luv to support dynamic loading on Windows.

lib_eio_windows/sched.ml
patricoferris: The incorrect use of `Obj.magic` has been removed in https://github.com/ocaml-multicore/eio/pull/497/commits/3ab48eaf0e30a04f243aa991977c1b96ac55afb9 -- I'm now relying on compare `Unix.file_descr` for the set of thing to check to read and write.

patricoferris:
Thanks @polytypic, I think that explains some of the errors I was seeing (which I've disabled I think). I'm not sure yet what's causing the new mdx errors in the CI (the stackoverflows with pretty bizarre line numbers).

patricoferris:
Hmm are we going to be bitten here by `FD_SETSIZE` in select ? https://github.com/ocaml/ocaml/blob/23dab79a4e42856aa33816b9c79c3d4d79959cb9/otherlibs/unix/select_win32.c#L942

> The default size FD_SETSIZE (currently 1024) is somewhat smaller than the current kernel limit to the number of open
     files.  However, in order to accommodate programs which might potentially use a larger number of open files with select, it is possible to increase this size within a program by providing a larger definition of FD_SETSIZE before the inclusion of <sys/types.h>

patricoferris:
@polytypic just for extra clarity this was the kind of error I was seeing with MDX -- it looks very similar to the one you had with Luv only but not quite the same maybe.

```
+Cannot load required shared library dllcstruct_unix_stubs.
+Reason: C:/OCaml64/home/PatrickFerris/.opam/5.0/lib/stublibs\dllcstruct_unix_st
ubs.dll: flexdll error: cannot relocate caml_failwith RELOC_REL32, target is too
 far: fffffffd958da134  ffffffff958da134.
````

**Comment by polytypic** [comment]
> `flexdll error: cannot relocate caml_failwith RELOC_REL32, target is too
 far: fffffffd958da134  ffffffff958da134.`

Yes, I saw those kinds of errors also.  The functions that may/need to be dynamically loaded should be marked with `__declspec(dllimport)`.

**Comment by avsm** [comment]
We shouldn't have to manually annotate declspec(dllimport) in user stubs like cstruct these days. The full background is in https://github.com/ocaml/ocaml/pull/1633, but I'm not sure what the current state of those tools are as Cygwin64 itself has been a moving target.

File: lib_eio/unix/fork/dune [hunk 9]
```diff
@@ -0,0 +1,4 @@
+(library
```
talex5, line 1: Does this need to be split out, or just a left-over change?

lib_eio/unix/fork_action.c
```diff
@@ -5,10 +5,14 @@
 #include <errno.h>
 
 #include <caml/mlvalues.h>
+#include <caml/unixsupport.h>
 
 #include "fork_action.h"
 
 void eio_unix_run_fork_actions(int errors, value v_actions) {
+  #ifdef _WIN32
```
talex5, line 13: You can probably #ifdef out the whole function. It isn't called from OCaml code.

lib_eio_windows/eio_windows_stubs.c
```diff
@@ -0,0 +1,90 @@
...
+}
+
+CAMLprim value caml_eio_windows_pwritev(value v_fd, value v_bufs, value v_offset)
+{
+  uerror("Pwritev is not supported on windows yet", Nothing);
+}
+
+CAMLprim value caml_eio_windows_openat(value v_dirfd, value v_pathname, value v_flags, value v_mode)
+{
+  uerror("Readv is not supported on windows yet", Nothing);
```
talex5, line 69:
```suggestion
  uerror("openat is not supported on windows yet", Nothing);
```

lib_eio_windows/fs.ml
```diff
@@ -0,0 +1,184 @@
...
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *)
+
+(* This module provides (optional) sandboxing, allowing operations to be restricted to a subtree.
+
+   For now, sandboxed directories use realpath and [O_NOFOLLOW], which is probably quite slow,
+   and requires duplicating a load of path lookup logic from the kernel.
+   It might be better to hold a directory FD rather than a path.
+   On FreeBSD we could use O_RESOLVE_BENEATH and let the OS handle everything for us.
+   On other systems we would have to resolve one path component at a time. *)
```
talex5, line 23: Comment needs updating.

lib_eio_windows/sched.ml
```diff
@@ -0,0 +1,364 @@
...
+          Lf_queue.push t.run_q IO;                   (* Re-inject IO job in the run queue *)
+          next t
+        )
+      )
+
+let with_sched fn =
+  let run_q = Lf_queue.create () in
+  Lf_queue.push run_q IO;
+  let sleep_q = Zzz.create () in
+  (* Pipe's on Windows cannot be nonblocking through the OCaml API. *)
```
talex5, line 233:
```suggestion
  (* Pipes on Windows cannot be nonblocking through the OCaml API. *)
```

talex5 (approved):
I left a few minor comments in-line, but basically good to merge once CI is passing. Currently, it says:
```
Error: Library "eio_windows" in _build/default/lib_eio_windows is hidden
```
Probably https://github.com/ocaml/dune/issues/5505 again. The solution is to use `(* -*- tuareg -*- *)` mode in the dune test file and disable the tests when not on Windows (see https://github.com/ocaml-multicore/eio/pull/475/files).

lib_eio/unix/fork/dune
patricoferris: Left-over before the ifdefs thanks!

lib_eio/unix/fork_action.c
```diff
@@ -9,10 +9,9 @@
 
 #include "fork_action.h"
 
+#ifdef _WIN32
+#else
```
talex5, line 13:
```suggestion
#ifndef _WIN32
```

talex5 (commented):
> I'm not sure yet what's causing the new mdx errors in the CI (the stackoverflows with pretty bizarre line numbers
```
[mdx] Fatal error: File "ctf.md", lines 33570790-33570791: Stack overflow
```
Does it work locally on Windows?

Given that `ctf.md` is using almost no features of Eio (just `Eio.Private.Ctf.mint_id`), this seems likely to be an MDX bug. If we can reproduce it without the `#require "eio"`, we can report it there.

It's a good sign that the CI is able to build on Windows though, even if the tests fail. Just checking the build works is probably OK for this PR. We know many of the real tests will fail anyway.

patricoferris:
> Does it work locally on Windows?

It's working locally for me (it would be nice to have a second person verify that). I agree it seems to be something going wrong with MDX and Windows, but only appearing in the CI. I'll see if I can make a demo repo with MDX failing like that to report (and maybe fix ^^") upstream -- we really need mdx to work on Windows in the long run !

patricoferris:
I threw a very simple md file into the tests that has zero dependencies and it also fails with the stackoverflow error see https://github.com/patricoferris/eio/actions/runs/4872160326/jobs/8689998313#step:5:22

talex5:
> So perhaps for this PR we can disable all MDX tests on Windows in the meantime and work to fix the declspec(dllimport) issues?

Sounds good to me!

lib_eio_windows/sched.ml
```diff
@@ -0,0 +1,364 @@
...
+            )
+          | e -> extra_effects.Effect.Deep.effc e
+      }
+  in
+  let result = ref None in
+  let `Exit_scheduler =
+    let new_fiber = Fiber_context.make_root () in
+    fork ~new_fiber (fun () ->
+        result := Some (with_op t main x);
+      )
```
polytypic, line 360: Hmm... Should we install the DLA support here?  (That was recently added to other backends.)

polytypic: See here for the [posix backend version](https://github.com/ocaml-multicore/eio/blob/7ea93f2bc3d5ecc6d4dbf8dc5d6e13e5c8831ff4/lib_eio_posix/sched.ml#L352-L360).

patricoferris: Yep! Hopefully supported in https://github.com/ocaml-multicore/eio/pull/497/commits/9a9c02ba080d63477b5cdcbad5999928bcc59bb0 with a test based on the README one.

polytypic (approved):

lib_eio/unix/fork_action.c
```diff
@@ -67,13 +70,17 @@ CAMLprim value eio_unix_fork_execve(value v_unit) {
 }
 
 static void action_fchdir(int errors, value v_config) {
+  #ifdef _WIN32
+  uerror("Unsupported operation on windows", Nothing);
```
talex5, line 74: We can't allocate OCaml exceptions here (or in the other actions) as we're in the child and anything involving GC isn't going to work.

```suggestion
  eio_unix_fork_error(errors, "action_fchdir", "Unsupported operation on windows");
  _exit(1);
```

talex5:
> I do wonder if (when it is ready) we should merge this into a stable Windows branch instead of the main branch, as it will probably be a pain to cut a release of Eio if this is in there?

We already can't cut a release, because we got rid of the Windows support from libuv and didn't replace it with anything yet.

nojb:
I understand that the POSIX backend for Windows is just a stopgap measure while waiting for a more performant IOCP backend, but just for future reference here's another issue with the emulation in `Unix.select` on Windows that we ran into some time ago: https://github.com/ocaml/ocaml/issues/7935

PR #498, opened 2023-04-30
Move Eio.Stdenv.t to Eio_unix.Stdenv.base

talex5:
The old type wasn't directly useful for anything, since nothing provides exactly this set of resources, and nothing needs to consume exactly this set either. However, it is useful in Eio_unix as the set of resources that a backend needs to provide to be compatible with `Eio_main.run`.

The new type is almost the same for now, except that the standard streams include file descriptors. Other Unix-specific features can be added later, such as support for passing file descriptors over sockets or to child processes.

I called the type `base` because we might want to add a `Stdenv.linux_or_unix` subtype later for basically everything that isn't Windows. Not sure what the name should be, though.

Note that, due to the way row-polymorphism works in OCaml, we can't e.g. have `Net.connect` return an `Eio_unix.socket` subtype of `Eio.Net.socket`. But we can add a new function (e.g. `connect_unix`) that returns a richer socket type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Eio replaces existing concurrency libraries such as Lwt
 * [Best Practices](#best-practices)
   * [Switches](#switches-1)
   * [Casting](#casting)
-  * [Passing Stdenv.t](#passing-stdenvt)
+  * [Passing env](#passing-env)
 * [Further Reading](#further-reading)
 
 <!-- vim-markdown-toc -->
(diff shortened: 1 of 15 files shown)


c-cube:
if I want to ask for "classic" permissions that a non-Eio program would have, what should I use now? I started something with 0.9 and exactly one `stdenv:Eio.Stdenv.t -> …` signature and now I'm a bit lost. I'm not interested in tying it to a particular backend either.

talex5:
If your program requires a "classic" OS then you can just use `Eio_unix.Stdenv.t` instead. It will work with any of the classic backends (eio_linux, eio_posix, eio_windows). It won't work with unikernels, but it wouldn't have done with the old Stdenv.t either.

But probably it would be better to list the things your program actually needs. See: https://github.com/ocaml-multicore/eio#passing-env

PR #499, opened 2023-05-02
Cross-platform subprocess support

talex5:
This is based on @patricoferris's #473, but rebased on top of #498 and with some changes:

- The standard streams are now optional, and default to inheriting from the parent.
- The executable is optional, and searched for in $PATH from the first argument if not given.
- `Process.parse_out` provides an easy way to parse the output of a process.
- The logic for converting flows to pipes is now shared in Eio_unix.

This also splits Eio_unix into separate modules.


lib_eio/process.mli
```diff
@@ -0,0 +1,148 @@
...
+  ?stderr:#Flow.sink ->
+  ?env:string array ->
+  ?executable:string ->
+  string list -> unit
+(** [run] does {!spawn} followed by {!await_exn}, with the advantage that if the process fails then
+    the error message includes the command that failed.
+
+    Note: If [spawn] needed to create extra fibers to copy [stdin], etc, then it also waits for those to finish. *)
+
+val parse_out :
```
patricoferris, line 124: Do you think we should also provide another convenience function that is `parse_out` with `Eio.Buf_read.take_all`, so a `pread` of sorts in [obuilder](https://github.com/ocurrent/obuilder/blob/master/lib/os.ml) speak I think. I just have a feeling most people will want that and it might not be obvious they can achieve that with those two.

lib_eio/process.mli
```diff
@@ -0,0 +1,148 @@
...
+    ?cwd:Fs.dir Path.t ->
+    ?stdin:Flow.source ->
+    ?stdout:Flow.sink ->
+    ?stderr:Flow.sink ->
+    ?env:string array ->
+    ?executable:string ->
+    string list ->
+    t
+end
+(** A process manager capable of spawning new processes. *)
```
patricoferris, line 81: I think (and this is on me from the initial commits) we've used 

 - Process
 - Subprocess
 - Sub-process
 - Child process

in different places to refer to the same thing. Sometimes it is okay, e.g. to describe the _process ID_. But here for example I'm guessing we should say a "manager capable of spawning new **subprocesses**" given that for the function `spawn` we say "creates a new subprocess". Perhaps this doesn't matter too much and the intent is clear?

tests/process.md
```diff
@@ -0,0 +1,159 @@
...
+
+```ocaml
+# run @@ fun mgr env ->
+  let root = env#fs / "/" in
+  Process.run mgr ~cwd:root [ "env"; "pwd" ];;
+/
+- : unit = ()
+```
+
+Changing directory (confined):
```
patricoferris, line 120: I'm pretty confident the code is correct, but do you think we need an example that tries to access a directory it shouldn't ?

patricoferris (approved):
Thanks! This looks great, I like the `parse_out` API.

tests/process.md
talex5: I added a test to check the error is sensible. It probably doesn't matter too much; either the OS is restricting it (in which case checking isn't our problem), or the OS will let the child access things outside of it anyway.

patricoferris: Indeed, thanks!

PR #500, opened 2023-05-03
Run benchmarks with current-bench

Sudha247:
Moves the benchmarks to use continuous benchmarking with [current-bench](https://github.com/ocurrent/current-bench). This PR makes benchmark files emit `ocaml-cb` compliant json format results.

The code needs more clean-up. Opening this draft early to see how the results get displayed in the current-bench front-end.

I've added `yojson` as a test dependency for now. We can maybe move it to a separate bench opam package.


talex5:
The link doesn't work (it goes to https://autumn.ocamllabs.io/ocaml-multicore/eio/pull/500; looks like it should be https://autumn.ocamllabs.io/ocaml-multicore/eio/pull/500/base/main). But then it just says "No data for selected interval".

Sudha247:
Looks like the benchmarks timed out on the default worker - this tends to happen for multicore benchmarks as the server is configured to run on a single core. @punchagan is configuring this to run on a different server that supports multicore benchmarks.

Before it timed out, the status was shown as `Passed` incorrectly. This may or may not be a bug. In any case Puneeth is aware of this and is trying to find the cause.

punchagan:
> The link doesn't work

Yes, that's a bug. We've a fix for this that we hope to merge soon.

Sudha247:
Making it ready for review as we see the graphs - https://autumn.ocamllabs.io/ocaml-multicore/eio/pull/500/base/main/benchmark/Eio?worker=fermat&image=ocaml%2Fopam%3Adebian-ocaml-5.0. Will look into cleaning them up.

talex5:
The graphs are a bit hard to read. We should probably remove the "promotions" ones (they're not very useful).

talex5:
The graphs are a bit hard to read. We should probably remove the "promotions" ones (they're not very useful).

talex5:
Now it says "No data for selected interval" again.

eio.opam
```diff
@@ -24,6 +24,7 @@ depends: [
   "alcotest" {>= "1.4.0" & with-test}
   "dscheck" {>= "0.1.0" & with-test}
   "odoc" {with-doc}
+  "yojson" {>= "2.0.2" & with-test}
```
talex5, line 27: This change needs to go in dune-project. This is the cause of the CI error (though the opam lint is expected to fail anyway):
```
eio.opam: changed after 'dune build @install'!
```

Sudha247:
Rebased on `main`. The CI failures on 32-bit look unrelated, IINM.

talex5:
I don't think the rebase worked. Those 32-bit failures were fixed on `main` last week and `main` is passing CI. The current head commit is bf25e24f9d6a.

Sudha247:
Oops, rebased on the right head now!

talex5 (commented):
I squashed your commits into one, and then added a few more:

- It now uses Yojson to generate the JSON, since we're adding a dependency on it anyway.
- I simplified the benchmarks code so the JSON generation is done from main.
- You can now choose to run only one benchmark.
- I tried to improve the labels and descriptions.
- I changed the dune build so that the benchmarks are attached to `eio_main`.

If that looks OK to you, we can squash and merge. Thanks!

Sudha247:
> Are / characters special in metric names? 

Yes, sorry I forgot to mention that. Paraphrasing what @ElectreAAS told me in a private thread - when we name benchmarks something like `bench_1/sub_plot_1` and `bench_1/sub_plot_2`, they're different lines in the same graph `bench_1`. I'm okay with keeping them as separate graphs too.

Cancel bench is a bit weird indeed. Makes me wonder if `-` is a special character too?

talex5 (approved):
I got the CI passing on Windows. Ready to merge once CI has finished (and looks sensible).

PR #509, opened 2023-05-07
Fs implementation for windows

patricoferris:
A WIP PR built on top of #497 adding support for filesystem operations primarily using `NtCreateFile` which apparently is the closest thing to `openat` that Windows has to offer.

### DIFF
--- a/lib_eio_windows/dune
+++ b/lib_eio_windows/dune
@@ -1,7 +1,7 @@
 (library
  (name eio_windows)
  (public_name eio_windows)
- (library_flags :standard -ccopt -lbcrypt)
+ (library_flags :standard -ccopt -lbcrypt -ccopt -lntdll)
  (enabled_if (= %{os_type} "Win32"))
  (foreign_stubs
   (language c)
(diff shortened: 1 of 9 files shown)


polytypic:
I tried this on my Windows machine and the symlink test seemed to fail.  I assume that is expected?  (I'll take a closer look through the code.)

patricoferris:
Still WIP, I'll be revisiting it this weekend, but thanks for taking a look already!

talex5:
I squashed and rebased this PR on main, which should fix the unrelated CI errors. I'd like to do an Eio 0.10 release, and this is the only remaining issue for it. We could either release without FS support, or merge this as it is and let people fix the missing bits. Any reason not to merge?

patricoferris:
> Any reason not to merge?

Apart from it potentially being a bit broken particularly with the symlinking behaviour and missing functionality like `pread`. But, perhaps a windows backend in the wild will lower the barrier to trying it out and then getting fixes from people?

talex5:
Sounds good. We can document eio_windows as being incomplete.

PR #511, opened 2023-05-08
Don't run Windows discover script on other platforms

talex5:
Otherwise, `dune build` will try to run it and fail.


patricoferris (approved):
Oops should've tried the windows PR on non-windows, thanks!

PR #516, opened 2023-05-12
Add Eio_unix.Net module

### LINKED ISSUE #342: Need Unix.socketpair_datagram
`Unix.socketpair` takes a `Unix.socket_type` argument, allowing you to create a datagram socket. However, it still returns a stream socket. Likewise, `as_socket` always returns a stream socket.

In particular, a zero-length read for a datagram socket does not mean end-of-file (spotted while trying to change the datagram test in network.md to reproduce #340)

talex5:
- Add `Eio_unix.Net` module and move network-related items from `Eio_unix` to there.
- Replace `Eio_unix.socketpair` with `Eio_unix.Net.socketpair_{stream,datagram}`. Previously, it took a type argument (stream or datagram), but always returned a `stream_socket`.
- Move sockaddr conversions to `eio.unix` so they can be shared.
- Implement `getnameinfo` in `eio.unix` to share it by default.
- `Eio_linux.Low_level.send_msg` no longer throws away the number of bytes sent.
- Add `close` to all socket types.
- The destination in `send` is now optional, for unconnected datagram sockets.
- `send` now takes a iovec, not just a single buffer. The stream API allows this and we tested it for datagrams using that API due to the previous type confusion.
- Add `Unix_datagram` socket address type for completeness.
- Add `send_msg` and `recv_msg` stubs to Eio_posix. This avoids converting to bytes, and will be useful to add FD passing later.

I was planning to include FD passing in this FD, but it got too long so I've split that out for later.

Fixes #342.

### DIFF
--- a/lib_eio/net.ml
+++ b/lib_eio/net.ml
@@ -143,6 +143,7 @@ module Sockaddr = struct
 
   type datagram = [
     | `Udp of Ipaddr.v4v6 * int
+    | `Unix of string
   ]
 
   type t = [ stream | datagram ]
@@ -158,9 +159,10 @@ end
 
 class virtual socket = object (_ : #Generic.t)
   method probe _ = None
+  method virtual close : unit
 end
 
-class virtual stream_socket = object
+class virtual stream_socket = object (_ : #socket)
   inherit Flow.two_way
 end
 
@@ -193,7 +195,7 @@ let accept_fork ~sw (t : #listening_socket) ~on_error handle =
 
 class virtual datagram_socket = object
   inherit socket
-  method virtual send : Sockaddr.datagram -> Cstruct.t -> unit
+  method virtual send : ?dst:Sockaddr.datagram -> Cstruct.t list -> unit
   method virtual recv : Cstruct.t -> Sockaddr.datagram * int
 end
 
--- a/README.md
+++ b/README.md
@@ -704,14 +704,13 @@ The `Eio.Io` type is extensible, so libraries can also add additional top-level 
 For example, this HTTP GET function adds the URL to any IO error:
 
 ```ocaml
-# let get ~net ~host ~path =
-    try
-      Eio.Net.with_tcp_connect net ~host ~service:"http" @@ fun _flow ->
-      "..."
-    with Eio.Io _ as ex ->
-      let bt = Printexc.get_raw_backtrace () in
-      Eio.Exn.reraise_with_context ex bt "fetching http://%s/%s" host path;;
-val get : net:#Eio.Net.t -> host:string -> path:string -> string = <fun>
+let get ~net ~host ~path =
+  try
+    Eio.Net.with_tcp_connect net ~host ~service:"http" @@ fun _flow ->
+    "..."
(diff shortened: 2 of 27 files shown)


talex5:
Having tried this out for a bit, this API is a little annoying because we also need unix equivalents to `Net.run_server`, etc. So I think it might be better to make the main Eio types polymorphic instead.

lib_eio/net.ml
```diff
@@ -143,13 +143,16 @@ module Sockaddr = struct
 
   type datagram = [
     | `Udp of Ipaddr.v4v6 * int
+    | `Unix_datagram of string
```
avsm, line 146: I'm still digesting this diff, but not sure why Unix_datagram is named like this in the portable core.  We have unix domain sockets, and potentially Windows named pipes, so should this be a little more abstract?

talex5: It could be. The unix(7) man-page says "The `AF_UNIX` (also known as `AF_LOCAL`) socket family ...", so perhaps `Local_datagram` would work?

I guess the question is whether it's useful to use Unix-domain sockets interchangeably with Windows named pipes. If so, they should share a name. If not, we can add another constructor for those.

avsm: There's a somewhat orthogonal set of properties that each expose. Windows named pipes are unidirectional streams I think.  Is the common property here that some things have a local name, and some are datagrams or streams.

talex5: According to https://learn.microsoft.com/en-us/windows/win32/ipc/pipe-names, a pipe name is in the format `\\ServerName\pipe\PipeName`. If a Linux process wanted to connect to a Windows named pipe on another machine, that would be quite different to connecting to a Unix socket with the same name, so I think they should be separate. Probably something like `Windows_pipe (server, local_name)`.

We could combine `Unix` and `Unix_datagram` I think. The only thing that distinguishes between them is `getnameinfo` to set the `NI_DGRAM` flag, but that isn't useful for Unix-domain sockets anyway.

I wonder if we should separate out Linux abstract socket addresses too. `Unix "/tmp/sock"` refers to a file, but `Unix "\000/tmp/sock"` doesn't. Or perhaps we shouldn't have a `Unix` network address at all, and instead add `Path.bind` and `Path.connect`. They are part of the mount namespace rather than the network namespace.

avsm:
One implication of this change is that `Eio.Net.accept_fork` and other connection handlers now must implement `close` (whereas previously it was just a Flow.two_way).

avsm:
(For reference, this is the diff I applied to bring other libraries uptodate with this change https://github.com/avsm/eeww/commit/dfdd8f248b249254b63bc5f18f8ab172ec94ca0e)

talex5:
Good point. At the moment, `accept_fork` always closes the flow when the handler finishes, so if you close it manually then that will fail, which is why we didn't have close here before. Perhaps we should revert the addition of close here.

avsm:
It would be cleaner to just pass a Flow.two_way I think, and leave closing to the connection handler.

Issue #512, opened 2023-05-09
Add task/worker pool similar to Domainslib task pool


darrenldl:
Thanks for your work first of all, it's been nice to use.

The Domainslib integration as described in README works fine, but working with mutexes then becomes a bit complicated.

- Invoking `Eio.Mutex.lock` etc within a Domainslib (or just non-Eio I guess) domain causes it to stop progressing
  - Not unexpected, but was not straightforward to track down, and Eio data structure documentation doesn't seem warn about this
- Swapping to stdlib's `Mutex` works, but then suffers from the problem of locking the entire domain needlessly if used on the Eio side

Right now it seems that there are enough components to replicate task pool
- Run a few background workers upon pool construction using Fiber.fork and Domain_manager.run
- Construct promises, embed within a closure and pass it onto work stream `(unit -> unit) Eio.Stream.t` (same as domainslib example in readme)

And it would be very nice to be able to stay in Eio domains for parallel worker pools.

I'm happy to draft a PR in next few days if the addition is not deemed problematic.

Sudha247:
Hi @darrenldl, thanks for bringing this up. Would [domain-local-await](https://github.com/ocaml-multicore/domain-local-await) work for you, for domainslib integration?

darrenldl:
Thanks @Sudha247 @talex5 !

> We have an item in the 1.0 roadmap for this (which I've now linked to this issue).

Ah I must have missed it, oops, my bad.

> We might want to make an Eio.Pool abstraction first, and then build worker pools on top of that, but we can prototype one that starts all domains immediately without that.

Could you elaborate on that? I'm confused by what `Eio.Pool` should do by itself - is it a pool of fibers?

talex5:
`Eio.Pool` would be similar to [Lwt_pool](http://ocsigen.org/lwt/5.5.0/api/Lwt_pool) - a generic way to hold resources, creating them as needed (up to a limit). But we can add that later.

darrenldl:
I see, seems like a good skeleton to base the problem on indeed.

patricoferris:
> Eio.Pool would be similar to [Lwt_pool](http://ocsigen.org/lwt/5.5.0/api/Lwt_pool) - a generic way to hold resources, creating them as needed (up to a limit). But we can add that later.

[...]

PR #517, opened 2023-05-15
Add resource pool and task pool

darrenldl:
PR for https://github.com/ocaml-multicore/eio/issues/512

This PR contains a draft for `Pool` and `Task_pool`.

Some choices are not immediately clear to me, so I've just settled on something that seems to work, but probably not optimal. This includes:
- Whether switch should be an argument in places or not
- Naming of functions
- How does Eio cancellation impact promises precisely

### DIFF
--- a/lib_eio/task_pool.ml
+++ b/lib_eio/task_pool.ml
@@ -3,39 +3,49 @@ type 'a task = unit -> 'a
 type runner = unit task -> unit
 
 type t = {
+  sw : Switch.t;
   runners : runner Pool.t;
 }
 
-let create ~max_domains domain_mgr : t =
-  let alloc () : runner =
-    let s : unit task Stream.t = Stream.create 0 in
-    Switch.run (fun sw ->
-        Fiber.fork ~sw (fun () ->
-            Domain_manager.run domain_mgr (fun () ->
-                let rec aux () =
-                  let f = Stream.take s in
+let create ~sw ~max_domains domain_mgr : t =
+  let alloc () : runner * runner Pool.handlers =
+    let s : unit task option Stream.t = Stream.create 0 in
+    Fiber.fork ~sw (fun () ->
+        Domain_manager.run domain_mgr (fun () ->
+            let rec aux () =
+              match Stream.take s with
+              | None -> ()
+              | Some f -> (
                   f ();
                   aux ()
-                in
-                aux ()
-              )
+                )
+            in
+            aux ()
           )
       );
-    (fun (f : unit task) : unit ->
-       let promise, resolver = Promise.create () in
-       Stream.add s (fun () ->
-           f ();
-           Promise.resolve resolver ());
-       Promise.await promise
-    )
+    let runner (f : unit task) : unit =
+      let promise, resolver = Promise.create () in
+      Stream.add s
+        (Some
+           (fun () ->
+              f ();
+              Promise.resolve resolver ()));
+      Promise.await promise
+    in
+    let check _ = true in
+    let dispose _ =
+      Stream.add s None
+    in
+    (runner, { check; dispose })
   in
   {
+    sw;
     runners : runner Pool.t = Pool.create ~alloc max_domains;
   }
 
 let async (t : t) (f : 'a task) =
   let promise, resolver = Promise.create () in
-  Pool.async t.runners (fun runner ->
+  Pool.async ~sw:t.sw t.runners (fun runner ->
       runner (fun () -> Promise.resolve resolver (f ()));
       Promise.await promise
     )
--- a/lib_eio/task_pool.ml
+++ b/lib_eio/task_pool.ml
@@ -1,0 +1,44 @@
+type 'a task = unit -> 'a
+
+type runner = unit task -> unit
+
+type t = {
+  runners : runner Pool.t;
+}
+
+let create ~max_domains domain_mgr : t =
+  let alloc () : runner =
+    let s : unit task Stream.t = Stream.create 0 in
+    Switch.run (fun sw ->
+        Fiber.fork ~sw (fun () ->
+            Domain_manager.run domain_mgr (fun () ->
+                let rec aux () =
+                  let f = Stream.take s in
+                  f ();
+                  aux ()
+                in
+                aux ()
+              )
+          )
+      );
+    (fun (f : unit task) : unit ->
+       let promise, resolver = Promise.create () in
+       Stream.add s (fun () ->
+           f ();
+           Promise.resolve resolver ());
+       Promise.await promise
+    )
+  in
+  {
+    runners : runner Pool.t = Pool.create ~alloc max_domains;
+  }
+
+let async (t : t) (f : 'a task) =
+  let promise, resolver = Promise.create () in
+  Pool.async t.runners (fun runner ->
+      runner (fun () -> Promise.resolve resolver (f ()));
+      Promise.await promise
+    )
+
+let run t f =
+  Promise.await (async t f)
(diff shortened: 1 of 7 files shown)


darrenldl:
Some technical details that I forgot to add during posting:
- `Lwt_pool` uses Stdlib's `Queue`, which does not have blocking wait. As such `Lwt_pool` uses a promise queue/list to wake threads up as resource free up.
- But since Eio's `Stream` has blocking wait, `Pool` does not use a promise queue/list here
- The design of `Pool` is otherwise copying from `Lwt_pool`, with `ref bool` swapped to `bool Atomic.t` for clear signal

darrenldl:
The design copied from `Lwt_pool` is a bit problematic at least for this case, and perhaps would be a limitation in general as well.

Right now, the allocation, dispose, and check functions must all be able to operate independently.
But for the case here, we want to generate at least the dispose function during the allocation,
to notify the loop inside `Domain_manager.run` to stop.

So I think it is better to have the following signature instead for `create`:
```
type 'a handlers = {
  dispose : 'a -> unit;
  check : 'a -> bool;
}

val create : ?init_size:int -> alloc:(unit -> 'a * 'a handlers) -> int -> 'a t
```

lib_eio/task_pool.ml
```diff
@@ -0,0 +1,44 @@
...
+
+type t = {
+  runners : runner Pool.t;
+}
+
+let create ~max_domains domain_mgr : t =
+  let alloc () : runner =
+    let s : unit task Stream.t = Stream.create 0 in
+    Switch.run (fun sw ->
+        Fiber.fork ~sw (fun () ->
```
talex5, line 13: This pattern doesn't do anything useful, because the switch will wait for the fiber to finish.

lib_eio/task_pool.mli
```diff
@@ -0,0 +1,9 @@
+type 'a task = unit -> 'a
+
+type t
+
+val create : max_domains:int -> Domain_manager.t -> t
```
talex5, line 5: There needs to be a switch argument here. Otherwise, any worker domain created can't outlive the `create` or `run` call, which defeats the point of a pool.

talex5 (commented):
The most useful thing would probably be to add a `tests/task_pool.md` demonstrating this working. There's a deterministic mock domain manager in `tests/network.md` (`fake_domain_mgr`) which would likely be useful for that, and should probably be moved into the `eio.mock` library so it can be shared.

darrenldl:
I've adjusted the API and added `tests/task_pool.md`. It doesn't quite give the interleaving output I want, which I presume is due to the mock domain manager only giving one actual domain.

darrenldl:
Okay, added a version that uses the native domain manager and shows the interleaving behaviour I want.

darrenldl:
Spotted a starvation issue in `Pool.async`. Will fix that along with the test failure.

lib_eio/task_pool.mli
```diff
@@ -0,0 +1,13 @@
+type 'a task = unit -> 'a
+
+type t
+
+val create : sw:Switch.t -> max_domains:int -> Domain_manager.t -> t
```
patricoferris, line 5: When I had a look at adding support for a task pool abstraction, I went with baking it directly into the `Domain_manager.t` capability. Users could then request a slice of that resource and get some amount back (I had in my mind that in the future you could add some user-specified way to dynamically scale this). This was in order to ensure we never overcommitted the "parallelism" resource. By allowing people to create multiple pools, I think that means they must ensure the sum of the `max_domains` they pass doesn't exceed a certain limit (presumably `Domain.recommended_domain_count ()`) ? Perhaps we don't want to be in charge of managing the parallel resource in this way ?

darrenldl: Personally I don't think Eio should be making the final decision here, since there are cases where you want OS preemptive scheduling to kick in even for your CPU heavy tasks, regardless of whether they actually run in parallel or not.

But I also don't know how expensive overcommitted domains or idling domains are.

darrenldl:
Right now there are a few directions I'm trying for making sure the resource pool does not leak or lock up in face of allocation and task error:
- Each "request" for resource (e.g. `async`) executes in a loop to try to obtain a resource or allocate when `current_size < max_size`
  - Using `Stream.take_nonblocking` means we lose fairness (presuming `Stream.take` respects the ordering of enqueuing waits), but this can probably be mitigated by some id pool structure
  - Downside is then there can be a lot of pointless wakeups/checks of tasks to see if they can progress
- Create a background thread during `Pool.create` to regularly try to allocate resources when `current_size < max_size`
  - Seems simplest, and can attribute `Switch.t` to fail easily on allocation error
  - Downside is then whether we care about the thread creating the pool being CPU busy, though we probably don't, i.e. users should be very aware of what Eio scheduling being cooperative means
- Handling refilling a pool on completion of task
  - This makes attribution of `Switch.t` to fail messier, though can be worked around by replacing `Stream.take` with a promise stream that some code resolves explicitly (as opposed to using Stream.add)
  - Performance characteristics is a bit muddy, since now a task may not appear to complete as it is trying to refill the pool, but user may expect the `Switch.t` provided to `async` to be ready to clear upon receiving a result

darrenldl:
Reworked the resource pool to use a monitor fiber (the second approach mentioned), and removed use of mutex, which when combined with blocking operations such as `Stream.add` and `Semaphore.acquire` makes deadlock analysis very messy.

Will need some (serious) test suite, but it works for the simple task pool atm.

darrenldl:
@patricoferris 

Seeing the new moonpool package (which uses Thread to preempt computations) and revisiting a [discussion thread](https://discuss.ocaml.org/t/no-domain-maximum-domain-count-in-the-stdlib/11173/7?u=darrenldl) which mentioned overcommitting domains leading to poor GC performance,
I think your suggestion of having a single domain pool behind the task pool is indeed the way to go.

The question is then whether we care about preemption or not within Eio itself.

SGrondin:
I personally quite like the design of `Lwt_pool` so I'm glad to see it ported here. However I don't believe it's necessarily the right foundation for a workpool.

Here are some of my thoughts about workpools in general:

Users know their own workloads (we don't). Library designers have to offer the right parameters for users to tune the workpool to their needs. As such I don't think a workpool should pay attention to `recommended_domain_count`; the users will.

An efficient design is one that fully utilizes each core. For CPU-bound workloads, it means 1 job per thread. For IO-bound it's a lot more. There's also hybrid workloads, with CPU-demanding processing interspersed between blocking IO calls. For those workloads, the right number is `1/p` where `p` is the proportion of each job being CPU-bound.

The user knows the right moment to start and shutdown their workpools. They want the threads to be ready to go by the time the jobs come flying in. Having to spawn threads on the fly (lazily) should be avoided.

Like many other people here, I've written a workpool using Eio. It's based on the principles I've described above. [Link](https://github.com/okTurtles/strings/tree/master/src/dispatcher). If the design proves interesting I'd be happy to submit it as a PR. You can think of it as a workpool of `n` threads times `m` workers.

darrenldl:
My main issue with task pool in this PR is a fiber blocking stops the domain entirely. Instead I would like it to have the behaviour closer to Rust's Tokio, where suspended fiber will be requeued automatically, allowing domain to process another fiber. To this end, I agree using resource pool as task pool being a bit hamfisted.

Whether this task pool behaviour should be built into the scheduler rather than being a separate set of API, I am really uncertain.

darrenldl:
I think it's better to split the resource pool into a separate PR rather than bundling the two. I'll open a new PR just for the resource pool.

PR #522, opened 2023-05-23
Eio_unix: add FD passing

talex5:
`Eio_unix.Net.stream_socket` now provides extra operations for sending and receiving FDs over Unix-domain sockets.

This means that calling e.g. `connect` on an `Eio_unix.Net.t` now returns an `Eio_unix.Net.stream_socket`, not a plain `Eio.Net.stream_socket`. This is a problem for OCaml's row polymorphism, since `#Net.t` only allows networks with extra methods, not sub-typing of existing ones.

Originally, I added new `connect_unix`, etc, methods to get around that, but the problem is that generic functions like `run_server` and `accept_fork` then lose the extra type information.

So instead I added a type parameter to `Eio.Net.t`. I thought this might be a bit messy, but it's not too bad. I only had to update one type annotation in the tests (changing `#Eio.Net.t` to `_ #Eio.Net.t`) and minor changes to capnp-rpc, so I think this is acceptable.

This is needed for the Wayland port to Eio (https://github.com/talex5/ocaml-wayland/tree/eio).


avsm:
This is surprisingly clean, even with the type parameter!  It doesnt add one to the actual socket type, which would be more difficult I think.  Looks good to me...

talex5:
This is now on top of #549.

The types will probably change again before Eio 0.11 due to the variants work, but I think it's best to get the functionality in now, and build the variants work on top of it to make sure that the new system can handle the extra complexity.

talex5:
Actually, I think this will work better as a PR on top of #553.

PR #524, opened 2023-05-24
eio_linux: fallback to fork if clone3 is unavailable

### LINKED ISSUE #514: Code with Eio.Process.spawn fails with “clone3 not implemented”
It's a case of _“it works on my laptop but fails in the CI (AWS + Docker)”_ ;)

In the container all calls fail with:

```default
Unix-error: Function not implemented (function: "clone3", argument: "")
```

This is the host's info:

```
$ uname -a
Linux 770143d8f119 5.15.0-1031-aws #35-Ubuntu SMP Fri Feb 10 02:07:18 UTC 2023 x86_64 GNU/Linux
```

If I replace `Eio_main` with `Eio_posix` it seems to work.

talex5:
`clone3` is blocked by Docker's default security policy.

Also, use `(uintptr_t)` when storing a pointer in a `uint64_t` field, otherwise it doesn't work on 32-bit systems.

Fixes #514.

### DIFF
--- a/lib_eio_linux/eio_stubs.c
+++ b/lib_eio_linux/eio_stubs.c
@@ -8,6 +8,7 @@
 #include <sys/random.h>
 #endif
 #include <sys/syscall.h>
+#include <sys/wait.h>
 #include <limits.h>
 #include <errno.h>
 #include <dirent.h>
@@ -29,6 +30,9 @@
 
 #ifndef SYS_pidfd_send_signal
 # define SYS_pidfd_send_signal 424
+#endif
+#ifndef SYS_pidfd_open
+# define SYS_pidfd_open 434
 #endif
 #ifndef SYS_clone3
 # define SYS_clone3 435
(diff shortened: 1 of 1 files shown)


avsm:
This does leave us in a strange half-way house, since we don't derive any benefit from using `clone3` if the interface always must be able to fallback to fork().  Any reason not to always just use fork?

talex5:
> Any reason not to always just use fork?

I was going to do that, but the fallback path requires some awkward code to handle the case where you spawn a process and then can't get an FD for it (e.g. due to FD limits), so I decided to use the new system when available. Also, I'm still planning to add support for namespaces at some point (it's OK if that doesn't work in a container).

lib_eio_linux/eio_stubs.c
```diff
@@ -148,24 +152,62 @@ CAMLprim value caml_eio_pidfd_send_signal(value v_pidfd, value v_signal) {
...
+   Also, raises exceptions rather then returning an error. */
+static pid_t clone3_with_fallback(struct clone_args *cl_args) {
+  int *pidfd = (int *)(uintptr_t) cl_args->pidfd;
+  pid_t child_pid = syscall(SYS_clone3, cl_args, sizeof(struct clone_args));
+
+  if (child_pid >= 0)
+    return child_pid;		/* Success! */
+
+  if (errno != ENOSYS && errno != EPERM) {
+    uerror("clone3", Nothing);	/* Unknown error */
```
avsm, line 169: This needs to be fixed everywhere, not just here, but I believe the recommended function from OCaml 5 is `caml_uerror` so it's namespaced right. (see ocaml/ocaml@67a4d75cfafa040099cdd322e23464362359af29).

avsm (approved):
A path towards supporting namespaces sounds like an excellent reason to keep trying to use clone3. LGTM

lib_eio_linux/eio_stubs.c
talex5: OK, let's do that in a separate PR.

PR #529, opened 2023-05-27
Allow windows users to run dune build

patricoferris:
This allows someone on windows to run `dune build` from the root of the project and it just builds what should work on Windows. Added that into CI also to protect against breaking it.


avsm (approved):
LGTM

PR #530, opened 2023-05-27
Windows vendor stubs

patricoferris:
Building on #529 

This PR handles most of the external pins that the windows backend needed:

 - The small number of cstruct stubs we needed are now vendored directly in `eio_windows`. We should probably also bring the `send` and `recv` stubs along but `eio_posix` could uses those, so perhaps a follow-up PR for that would be nice.
 - Turns out most of our woes stemmed back to not having [a patch for ocamlbuild from opam-repository-mingw](https://github.com/fdopen/opam-repository-mingw/blob/opam2/packages/ocamlbuild/ocamlbuild.0.14.2/files/ocamlbuild-0.14.2.patch) which I think is the essence of this PR https://github.com/ocaml/ocamlbuild/pull/70 -- though not equal. I have the patch applied to a fork of ocamlbuild and that is now the only pin in the `eio_windows` package. The patch either needs upstreaming to ocamlbuild or ocaml/opam-repository.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,8 +52,7 @@ jobs:
       - run: | 
           opam pin -yn eio.dev .
           opam pin -yn eio_windows.dev .
-          opam install ocamlfind.1.9.5 kcas alcotest mdx crowbar dscheck -y
-          opam install eio eio_windows --deps-only
+          opam install eio eio_windows --deps-only --with-test
       - run: opam exec -- dune build
       - run: opam exec -- dune runtest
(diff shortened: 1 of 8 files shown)


avsm:
Time to ditch that last pin-depends! https://github.com/ocaml/opam-repository/pull/23832

avsm (approved):

talex5 (approved):
Thanks! Should we merge this now, or wait for https://github.com/ocaml/opam-repository/pull/23832?

avsm:
Merge away I say, can do a followup for the last pin-depends later. Need to wait a few days for feedback on the opam-repo one.

PR #531, opened 2023-05-28
support for creating ptys and a login_tty fork_action

avsm:
Add pseudoterminal support and a login_tty fork action (layered over #854 commits)

Adds Eio_unix.Pty for creating pseudoterminals and a login_tty fork action for running a child on one. Pseudoterminals are created with posix_openpt so both ends are close-on-exec.  If available, the terminal-device is opened via TIOCGPTPEER (Linux) and falls back to opening ptsname otehrwise (macOS etc).

login_tty is a fork action that starts a new session, makes the terminal device the controlling terminal, and dups it onto the child's stdin/stdout/stderr. It's exposed via Eio_unix.Process.spawn_unix as a ?login_tty argument.

Based on review feedback in this PR from @talex5, @patricoferris and @RyanGibb.

### DIFF
--- a/lib_eio/unix/fork_action.c
+++ b/lib_eio/unix/fork_action.c
@@ -13,6 +13,11 @@
 #include <string.h>
 #include <errno.h>
 
+#ifndef _WIN32
+#include <sys/ioctl.h>      /* TIOCSCTTY, for login_tty */
+#include <termios.h>
+#endif
+
 #include <caml/mlvalues.h>
 #include <caml/unixsupport.h>
 #include <caml/memory.h>
@@ -302,4 +307,25 @@ static void action_setgid(int errors, value v_config) {
 
 CAMLprim value eio_unix_fork_setgid(value v_unit) {
   return Val_fork_fn(action_setgid);
+}
+
+static void action_login_tty(int errors, value v_config) {
+  #ifdef _WIN32
+  eio_unix_fork_error(errors, "login_tty", ENOSYS);
+  _exit(1);
+  #else
+  int fd = Int_val(Field(v_config, 1));
+  if (setsid() == -1) {
+    eio_unix_fork_error(errors, "setsid", errno);
+    _exit(1);
+  }
+  if (ioctl(fd, TIOCSCTTY, 0) == -1) {
+    eio_unix_fork_error(errors, "ioctl(TIOCSCTTY)", errno);
+    _exit(1);
+  }
+  #endif
+}
+
+CAMLprim value eio_unix_login_tty(value v_unit) {
+  return Val_fork_fn(action_login_tty);
 }
--- a/lib_eio/unix/process.mli
+++ b/lib_eio/unix/process.mli
@@ -23,6 +23,7 @@ module Pi : sig
       ?pgid:int ->
       ?uid:int ->
       ?gid:int ->
+      ?login_tty:Fd.t ->
       env:string array ->
       fds:(int * Fd.t * Fork_action.blocking) list ->
       executable:string ->
@@ -48,6 +49,7 @@ module Make_mgr (X : sig
     ?pgid:int ->
     ?uid:int ->
     ?gid:int ->
+    ?login_tty:Fd.t ->
     env:string array ->
     fds:(int * Fd.t * Fork_action.blocking) list ->
     executable:string ->
@@ -62,6 +64,7 @@ val spawn_unix :
     ?pgid:int ->
     ?uid:int ->
     ?gid:int ->
+    ?login_tty:Fd.t ->
     fds:(int * Fd.t * Fork_action.blocking) list ->
     ?env:string array ->
     ?executable:string ->
@@ -71,7 +74,10 @@ val spawn_unix :
 
     The arguments are as for {!Eio.Process.spawn},
     except that it takes a list of FD mappings for {!Private.Fork_action.inherit_fds}
-    directly, rather than just flows for the standard streams. *)
+    directly, rather than just flows for the standard streams.
+
+    @param login_tty If given, the child starts a new session with this terminal
+                     device as its controlling terminal and stdin/stdout/stderr. *)
 
 val sigchld : Eio.Condition.t
 (** {b If} an Eio backend installs a SIGCHLD handler, the handler will broadcast on this condition.
--- a/lib_eio/unix/pty.c
+++ b/lib_eio/unix/pty.c
@@ -1,0 +1,135 @@
+/* Pseudoterminal support for Eio_unix.Pty. */
+
+#include "primitives.h" /* Defines _GNU_SOURCEfor posix_openpt. */
+
+#include <string.h>
+#include <errno.h>
+
+#ifndef _WIN32
+#include <stdlib.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/ioctl.h>
+#include <termios.h>
+#endif
+
+#include <caml/mlvalues.h>
+#include <caml/alloc.h>
+#include <caml/memory.h>
+#include <caml/fail.h>
+#include <caml/unixsupport.h>
+
+/* Returns [pty_fd], opened close-on-exec. */
+CAMLprim value eio_unix_open_pty(value v_unit) {
+#ifdef _WIN32
+  caml_unix_error(EOPNOTSUPP, "open_pty", Nothing);
+#else
+  CAMLparam1(v_unit);
+
+#if defined(__OpenBSD__)
+  /* OpenBSD posix_openpt accepts only O_RDWR | O_NOCTTY,
+     so request close-on-exec separately via fcntl.
+     https://man.openbsd.org/posix_openpt.3 */
+  int pty_fd = posix_openpt(O_RDWR | O_NOCTTY);
+  if (pty_fd < 0)
+    caml_uerror("posix_openpt", Nothing);
+  if (fcntl(pty_fd, F_SETFD, FD_CLOEXEC) < 0) {
+    int e = errno;
+    close(pty_fd);
+    errno = e;
+    caml_uerror("posix_openpt.fcntl", Nothing);
+  }
+#else
+  int pty_fd = posix_openpt(O_RDWR | O_NOCTTY | O_CLOEXEC);
+  if (pty_fd < 0)
+    caml_uerror("posix_openpt", Nothing);
+#endif  /* __OpenBSD__ */
+
+  CAMLreturn(Val_int(pty_fd));
+#endif  /* _WIN32 */
+}
+
+/* Returns [(tty_fd, name)], opened close-on-exec. */
+CAMLprim value eio_unix_get_pty_peer(value v_pty) {
+#ifdef _WIN32
+  caml_unix_error(EOPNOTSUPP, "TIOCGPTPEER", Nothing);
+#else
+  CAMLparam1(v_pty);
+  CAMLlocal2(v_name, v_ret);
+  int pty_fd = Int_val(v_pty);
+
+  if (grantpt(pty_fd) < 0)
+    caml_uerror("grantpt", Nothing);
+
+  if (unlockpt(pty_fd) < 0)
+    caml_uerror("unlockpt", Nothing);
+
+  {
+#if defined(__OpenBSD__)
+    char *name = ptsname(pty_fd);
+    if (name == NULL)
+      caml_uerror("ptsname", Nothing);
+    v_name = caml_copy_string(name);
+#else
+    char namebuf[256];
+    int err = ptsname_r(pty_fd, namebuf, sizeof namebuf);
+    if (err != 0) {
+      errno = err;
+      caml_uerror("ptsname_r", Nothing);
+    }
+    v_name = caml_copy_string(namebuf);
+#endif
+  }
+
+  int tty_fd;
+#ifdef TIOCGPTPEER
+  tty_fd = ioctl(pty_fd, TIOCGPTPEER, O_RDWR | O_NOCTTY | O_CLOEXEC);
+  if (tty_fd < 0)
+    caml_uerror("ioctl(TIOCGPTPEER)", Nothing);
+#else
+  tty_fd = open(String_val(v_name), O_RDWR | O_NOCTTY | O_CLOEXEC);
+  if (tty_fd < 0)
+    caml_uerror("open", v_name);
+#endif
+
+  v_ret = caml_alloc_tuple(2);
+  Store_field(v_ret, 0, Val_int(tty_fd));
+  Store_field(v_ret, 1, v_name);
+  CAMLreturn(v_ret);
+#endif
+}
+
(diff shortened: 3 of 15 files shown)


RyanGibb:
This looks great! Using `login_tty` is much nicer than manually doing tty setup.

> [ ] do the pty fds need to created as cloexec, and then inherited in a fork_action to the subshell? Right now they are not cloexec, but there is no close Fork_action and so there is too much open on the other side.

I saw that they're manually closed in https://github.com/avsm/melange/blob/e92240e6dc8a440cafa91488a1fc367e2ba57de1/lib/ounix/ounix.ml#L66. Perhaps they should similarly be closed in a `Pty.close` after the fork returns? I think setting cloexec would require manually creating a pty by opening ptmx, and calling ioctl TIOCGPTPEER, etc, as openpty doesn't doesn't support it.

lib_eio/unix/pty.c
```diff
@@ -0,0 +1,87 @@
...
+#include <caml/unixsupport.h>
+
+value eio_unix_open_pty(value v_unit)
+{
+    CAMLparam1 (v_unit);
+    char namebuf[4096]; /* Not PATH_MAX due to portability issues */
+    int i, masterfd, slavefd;
+    CAMLlocal1(v_ret);
+
+    i = openpty(&masterfd, &slavefd, namebuf, NULL, NULL);
```
talex5, line 48: The man-page notes that:
> Nobody knows how much space should be reserved for name.  So, calling openpty() or forkpty() with non-NULL name may  not be secure.

However, `ptsname_r` is Linux-only.

In a multi-domain program, all FDs must be created close-on-exec atomically, or they can leak. https://www.austingroupbugs.net/view.php?id=411 says:
> The function posix_openpt( ) with the O_CLOEXEC flag is necessary to
avoid a data race in multi-threaded applications. Without it, a file
descriptor is leaked into a child process created by one thread in the
window between another thread creating a file descriptor with
posix_openpt( ) and then using fcntl( ) to set the FD_CLOEXEC flag.

However, it's not clear how widely-supported that is.

The man-page for `posix_openpt` notes that you can just open `/dev/ptmx` directly, which would avoid these problems, but require some other manual setup.

lib_eio/unix/pty.ml
```diff
@@ -0,0 +1,16 @@
+type pty = {
+  masterfd : Unix.file_descr;
+  slavefd : Unix.file_descr;
```
talex5, line 3: The public interface should use `Eio_unix.Fd.t`, and attach them to a switch (or maybe two switches?).

talex5 (commented):
Adding `login_tty` seems reasonable. I guess the other option is to add fork actions for creating sessions and setting controlling terminals, but maybe that's less portable. Plumbing it through as an extra argument to `spawn_unix` seems fine.

The `opentty` situation seems like a huge mess, but maybe we just need to document that it only works in single-domain programs (and set close-on-exec quickly after getting the FDs). Or this could be in a separate library.

lib_eio/unix/pty.c
avsm: It seems to come down to using either BSD-style pty handling (openpty) or Unix98-style handling (posix_openpt).  The latter seems very much preferred, and a quick smoke test on macOS, Linux and OpenBSD shows posix_openpt is present there.  I'll have a go at switching this PR to using /dev/ptmx instead -- there's code in portable OpenSSH to use as a guide.

lib_eio/unix/pty.ml
```diff
@@ -0,0 +1,16 @@
+type pty = {
+  masterfd : Unix.file_descr;
```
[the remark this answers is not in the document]
RyanGibb: I looked into this while experimenting with https://github.com/RyanGibb/ocaml-exec-shell. It seems UNIX/POSIX/Linux use the existing terminology and I couldn't find any discussions around changing it. A [relevant thread](https://sourceware.org/pipermail/libc-alpha/2020-July/115792.html) is in the glibc mail archive where a number of alternatives are discussed. I couldn't see a consensus on new terminology, but changing master and slave to pty and tty respectively with 'pseudoterminal device' and 'terminal device' in prose seemed popular.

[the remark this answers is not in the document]
avsm: Using the pty and terminal device terminology is fine by me, of course. I'll do that in the next revision of this diff.

lib_eio/unix/fork_action.c
```diff
@@ -189,3 +192,14 @@ static void action_dups(int errors, value v_config) {
 CAMLprim value eio_unix_fork_dups(value v_unit) {
   return Val_fork_fn(action_dups);
 }
+
+static void action_login_tty(int errors, value v_config) {
+  value v_pty = Field(v_config, 1);
+
+  if (login_tty(Int_val(v_pty)) == -1)
+    dprintf(errors, "action_login_tty Error logging in tty: %s", strerror(errno));
```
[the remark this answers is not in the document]
avsm: fixed in fd5a1db39d61b2eaff851423ad8f80b30cb0e345

lib_eio/unix/pty.c
avsm: switched to posix_openpt in fd5a1db39d61b2eaff851423ad8f80b30cb0e345

lib_eio/unix/pty.ml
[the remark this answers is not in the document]
avsm: fd5a1db39d61b2eaff851423ad8f80b30cb0e345 uses pty/tty now

lib_eio/unix/pty.ml
avsm: Done in fd5a1db39d61b2eaff851423ad8f80b30cb0e345, but just one switch..

avsm:
Rebased now that #854 is merged

avsm:
Bah, this is still failing on FreeBSD

```
--- a/_build/default/tests/pty.md
+++ b/_build/default/tests/.mdx/pty.md.corrected
@@ -74,7 +74,7 @@ Output the child writes to its terminal can be read back from the `pty` end:
in
let r = Eio.Buf_read.of_flow (Pty.source t) ~max_size:1024 in
Eio.Buf_read.lines r |> Seq.take 3 |> List.of_seq;;
-- : string list = ["line 1"; "line 2"; "line 3"]
+- : string list = []

Input can be sent to the child by writing to the `pty` end with {!Pty.sink}.
@@ -114,7 +114,7 @@ With echo left on, the terminal echoes the input back to the `pty` end as well:
Eio.Flow.copy_string "hello\n" (Pty.sink t);
let r = Eio.Buf_read.of_flow (Pty.source t) ~max_size:1024 in
Eio.Buf_read.lines r |> Seq.take 2 |> List.of_seq;;
-- : string list = ["hello"; "reply: hello"]
+- : string list = ["hello"]
```

avsm:
Now fixed thanks to a FreeBSD shell from @mtelvers. The problem was in the tests (FreeBSD discards outstanding pty buffers when closed, Linux seems to persist them).

tests/pty.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+Input can be sent to the child by writing to the `pty` end with {!Pty.sink}.
+Here we turn off the terminal's input echo, send it a line, and read the response:
+
+```ocaml
+# run @@ fun env ->
+  Switch.run @@ fun sw ->
+  let t = Pty.open_pty ~sw () in
+  Eio_unix.Fd.use_exn "tty" (Pty.tty t) (fun fd ->
```
avsm: Design question: should I reexport `Unix.tcgetattr` so we dont need to call Unix? I could do so in Eio_unix, and save an Eio equivalent for when we get Windows support.

lib_eio/unix/fork_action.c
```diff
@@ -303,3 +308,32 @@ static void action_setgid(int errors, value v_config) {
 CAMLprim value eio_unix_fork_setgid(value v_unit) {
   return Val_fork_fn(action_setgid);
 }
+
+static void action_login_tty(int errors, value v_config) {
```
avsm, line 312: The reason we don't actually call libc's logintty(3) here is that it requires `-lutil` on some platforms which would require some discover magic. What the C function does is just 3 function calls though, so they're just invoked directly here.

lib_eio/unix/process.mli
```diff
@@ -71,7 +74,10 @@ val spawn_unix :
 
     The arguments are as for {!Eio.Process.spawn},
     except that it takes a list of FD mappings for {!Private.Fork_action.inherit_fds}
-    directly, rather than just flows for the standard streams. *)
+    directly, rather than just flows for the standard streams.
+
+    @param login_tty If given, the child starts a new session with this terminal
+                     device as its controlling terminal and stdin/stdout/stderr. &*)
```
talex5:
```suggestion
                     device as its controlling terminal and stdin/stdout/stderr. *)
```

lib_eio/unix/fork_action.c
talex5: This is actually convenient, as we don't want to dup the FDs here anyway as it interferes with `inherit_fds`:

- `inherit_fds` might overwrite the `tty` FD before we get to this action (I've changed it to go first now).
- The user might try to set stdin etc manually and should get an error in this case.

I've removed the dupping code from here and changed the backends to use `inherit_fds` for this.

lib_eio/unix/pty.ml
```diff
@@ -0,0 +1,36 @@
+open Eio.Std
+
+type t = {
+  pty : [`Unix_fd | Net.stream_socket_ty] r;
```
talex5: Just to be clearer that it's not really a socket:
```suggestion
  pty : [Types.source_ty | Types.sink_ty] r;
```

lib_eio/unix/pty.ml
```diff
@@ -0,0 +1,36 @@
...
+  name : string;
+}
+
+external create : unit -> Unix.file_descr * Unix.file_descr * string = "eio_unix_open_pty"
+
+let open_pty ~sw () =
+  let pty_fd, tty_fd, name = create () in
+  Unix.set_nonblock pty_fd;
+  let pty = Net.import_socket_stream ~sw ~close_unix:true pty_fd in
+  let tty = Fd.of_unix ~sw ~blocking:true ~seekable:false ~close_unix:true tty_fd in
```
talex5: Just some minor anti-leaking improvements:
- Check the switch isn't finished before creating the FDs.
- Do the simpler `Fd.of_unix` first, as it's less likely to fail.

```suggestion
  Switch.check sw;
  let pty_fd, tty_fd, name = create () in
  Unix.set_nonblock pty_fd;
  let tty = Fd.of_unix ~sw ~blocking:true ~seekable:false ~close_unix:true tty_fd in
  let pty = Net.import_socket_stream ~sw ~close_unix:true pty_fd in
```

tests/pty.md
talex5: Using `Unix` isn't a problem in itself I think, but it might be nice to have a wrapper that takes `Fd.t`.

talex5 (approved):
Looks good.

(I wonder if `Pty.source` and `sink` should be combined to return a bi-directional flow, but maybe they're logically separate?)

lib_eio/unix/pty.ml
avsm: I don't entirely understand why we need to do the manual Switch.check here; wouldnt a canceled switch just unwind via exceptions?

tests/pty.md
avsm: Done in ed5affc61ed8d229210a50b196728fc48a57412d -- note that some run in a systhread to ensure only the current Eio fibre blocks; hope I got all the cases.

lib_eio/unix/pty.ml
talex5: Yes, but the problem is that we have two things to register. The first will get freed but the exception will prevent the second one being registered.

But actually there's a simpler solution here: just split `eio_unix_open_pty` into two functions. That would probably simplify the C code too. I can have a got at that.

talex5: Done in https://github.com/ocaml-multicore/eio/pull/531/changes/d26ab7a7359dcbc4b2cc70460b02366c90916ba9

lib_eio/unix/pty.mli
```diff
@@ -0,0 +1,91 @@
...
+module Tc : sig
+  val getattr : Fd.t -> Unix.terminal_io
+  (** [getattr fd] returns the current terminal attributes of [fd].
+      See {!Unix.tcgetattr}. *)
+
+  val setattr : Fd.t -> Unix.setattr_when -> Unix.terminal_io -> unit
+  (** [setattr fd when_ attr] sets the terminal attributes of [fd].
+
+      With [TCSADRAIN] or [TCSAFLUSH] the change waits for pending output to
+      drain in the current fibre.
```
talex5: I pushed a commit renaming "fibre" to "fiber".

talex5 (approved):

tests/pty.md
talex5: Good idea - I hadn't thought of the blocking aspect!

lib_eio/unix/pty.mli
avsm: One of these days I'll get that right first time...

lib_eio/unix/pty.ml
avsm: That makes sense, thanks!

talex5:
It's failing on OpenBSD:
```
pty.c:54:15: warning: call to undeclared function 'ptsname_r'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
54 |     int err = ptsname_r(pty_fd, namebuf, sizeof namebuf);
```

avsm:
Two separate fixes needed:
- openbsd dont have ptsname_r so opt to make this domain unsafe there 13a9fa49641b4fe5ac67ac791369af9e03812b04; i dont think its worth the trouble of making openpty multidomain safe, but we can add a mutex if necessary
- the openbsd posix_openpt explicitly doesnt want CLOEXEC in the flags, so we need to do that in a second call in b5e11cc9efe2dd00b94984f20e7effb4b0fd7e6c

lib_eio/unix/pty.c
```diff
@@ -0,0 +1,135 @@
...
+  if (unlockpt(pty_fd) < 0)
+    caml_uerror("unlockpt", Nothing);
+
+  {
+    char namebuf[256];
+#if defined(__OpenBSD__)
+    char *name = ptsname(pty_fd);
+    if (name == NULL)
+      caml_uerror("ptsname", Nothing);
+    strlcpy(namebuf, name, sizeof namebuf);
```
talex5: This will silently truncate it in the unlikely case that it's too long; why not just `v_name = caml_copy_string(name);`?

avsm: yes good idea; i thought 256 was the max tty name len but might as well avoid the situation; fixed in  ef5115ba854be5cf8249f15a4f7c7a6380eb2e0b

talex5:
Ready to merge once CI is finished.

PR #532, opened 2023-05-29
Remove pins for windows

patricoferris:
This PR removes the final pins for the windows backend using the PR to opam-repository to add the ocamlbuild patch (thanks @avsm).

### DIFF
--- a/eio_windows.opam
+++ b/eio_windows.opam
@@ -30,7 +30,3 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/ocaml-multicore/eio.git"
-pin-depends: [
-  # With Fdopen's patch, the only thing we need from opam-repo-mingw
-  [ "ocamlbuild.dev" "git+https://github.com/patricoferris/ocamlbuild#b72c0dfdd6f1066d50da4b924b8196237696f66b" ]
-]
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -47,7 +47,7 @@ jobs:
           ocaml-compiler: ocaml.5.0.0,ocaml-option-mingw
           opam-repositories: |
             dra27: https://github.com/dra27/opam-repository.git#windows-5.0
```
avsm, line 49: How far does it get if we remove the dra27 repo here?

patricoferris: Not far I believe, because:

 - We can't create a 5.0.0 compiler switch
 - We need the patches to ocamlfind I think too https://github.com/dra27/opam-repository/tree/windows-5.0/packages/ocamlfind/ocamlfind.1.9.5/files

That is, maybe, all we're getting out of that repo.

avsm: Thanks, I'll look at merging the ocamlfind one tomorrow.  We can get by without a compiler switch with an ocaml-system install possibly.

avsm (approved):

talex5 (approved):
The opam-repository PR has been merged, so I removed the commit hash from this PR.

PR #539, opened 2023-06-01
Buf_write: fix flush returning too early

### LINKED ISSUE #528: Buf_write.flush should wait for the write to complete
`Buf_write.with_flow` runs a fiber that asks the destination flow to copy from the buffer. Each time the flow performs a read, we remove the data from the buffer, completing any flushes waiting for that. However, the data may not have been sent yet. In particular, the user may be waiting for the flush to complete to close or shut down the flow.

Reported in #527.

talex5:
`flush` previously only waited until the data had been taken out of the buffer (i.e. received by the underlying flow), but did not wait for it to be transmitted.

Fixes #528, reported by @cometkim.

### DIFF
--- a/lib_eio/buf_write.ml
+++ b/lib_eio/buf_write.ml
@@ -469,27 +469,21 @@ let rec await_batch t =
       );
     await_batch t
 
-let read_into t buf =
-  let iovecs = await_batch t in
-  let n, _iovecs = Cstruct.fillv ~src:iovecs ~dst:buf in
-  shift t n;
-  n
-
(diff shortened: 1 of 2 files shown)


avsm:
How about extending the sink interface to support short writes instead? We will have to do that at some point, and sources already support short reads.

talex5:
> How about extending the sink interface to support short writes instead?

I'm planning to do that after releasing 0.10 (mainly because there will likely be some other API changes then and I'd rather not make people update twice in a row).

PR #543, opened 2023-06-02
Use `Fmt.Dump.signal` to format signals

talex5:
Thanks to @MisterDA for letting me know about this!

### DIFF
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -6,9 +6,9 @@ type exit_status = [
 type status = [ exit_status | `Stopped of int ]
 
 let pp_status ppf = function
-  | `Exited i -> Format.fprintf ppf "Exited %i" i
-  | `Signaled i -> Format.fprintf ppf "Signalled %i" i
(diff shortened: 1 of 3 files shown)


lib_eio/process.ml
```diff
@@ -7,8 +7,8 @@ type status = [ exit_status | `Stopped of int ]
 
 let pp_status ppf = function
   | `Exited i -> Format.fprintf ppf "Exited %i" i
-  | `Signaled i -> Format.fprintf ppf "Signalled %i" i
-  | `Stopped i -> Format.fprintf ppf "Stopped %i" i
+  | `Signaled i -> Format.fprintf ppf "Signalled %a" Fmt.Dump.signal i
```
MisterDA, line 10: Is it _signaled_ or _signalled_? :)

talex5: Good point. I guess the single-L is to match Unix's `WIFSIGNALED`. I've changed the wording to avoid the problem.

PR #547, opened 2023-06-05
Allow calling close more than once

### LINKED ISSUE #323: file descriptor used after calling close!
- https://github.com/ocaml-multicore/eio/issues/244 seemed like it had a different scope (specific to `accept_fork`), so I opened a new one.

- I'm hitting a case where I sometimes end up calling `Eio.Flow.close` a second time, and that blows up my application
- in cases where it's hard to guarantee that I only call `close` once, would it be reasonable to implement either one or both of the following?
  1. make `Eio.Flow.close` not throw (but perhaps leave the other operations throwing?)
  2. (at least) make the exception raised by `Eio_luv.Handle.get` easier to pattern match against. Right now the pattern needs to be `Invalid_argument "close: handle used after calling close!"` which feels a bit brittle.

talex5:
Previously, some resources raised an exception if you tried to close them more than once, while others ignored it. Document that calling close on a closed resource does nothing.

This is useful for `accept_fork`, which always tries to close the socket at the end, but would like to allow the user to close it sooner if they wish.

Closes #323 (requested by @anmonteiro). It should also allow #537 to be reverted, simplifying #522.

I also moved the definition of `close` to `Generic`, since it's not specific to flows, and documented that it doesn't necessarily cancel in-progress operations.

(it used to be important to avoid double-closes, since there could be a race if two domains closed at once, but that's fixed now)

### DIFF
--- a/lib_eio/generic.mli
+++ b/lib_eio/generic.mli
@@ -9,3 +9,22 @@ val probe : #t -> 'a ty -> 'a option
 (** [probe t feature] checks whether [t] supports [feature].
     This is mostly for internal use.
     For example, {!Eio_unix.FD.peek_opt} uses this to get the underlying Unix file descriptor from a flow. *)
+
+(** {2 Closing}
+
+    Resources are usually attached to switches and closed automatically when the switch
+    finishes. However, it can be useful to close them sooner in some cases. *)
+
+class type close = object
+  method close : unit
+end
+
+val close : #close -> unit
+(** [close t] marks the resource as closed. It can no longer be used after this.
+
+    If [t] is already closed then this does nothing (it does not raise an exception).
+
+    Note: if an operation is currently in progress when this is called then it is not
+    necessarily cancelled, and any underlying OS resource (such as a file descriptor)
+    might not be closed immediately if other operations are using it. Closing a resource
+    only prevents new operations from starting. *)
--- a/lib_eio/core/switch.ml
+++ b/lib_eio/core/switch.ml
@@ -14,6 +14,8 @@ type hook =
 
 let null_hook = Null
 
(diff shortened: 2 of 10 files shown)


lib_eio/generic.mli
```diff
@@ -9,3 +9,22 @@ val probe : #t -> 'a ty -> 'a option
 (** [probe t feature] checks whether [t] supports [feature].
     This is mostly for internal use.
     For example, {!Eio_unix.FD.peek_opt} uses this to get the underlying Unix file descriptor from a flow. *)
+
+(** {2 Closing}
+
+    Resources are usually attached to switches and closed automatically when the switch
+    finishes. However, it can be useful to close them sooner manually in some cases. *)
```
patricoferris, line 16:
```suggestion
    finishes. However, it can be useful to manually close them sooner in some cases. *)
```

patricoferris (commented):
Agreed this would be nice, was there not some discussion about returning a `bool` indicating whether anything had actually happened or not (perhaps it's hard to make that domain safe)? I seem to recall that conversation and just wondering if there were cases where that could be useful, otherwise looks good :))

anmonteiro:
What are your general thoughts on the duality of idempotency in `close` and the required use of capabilities?

Intuitively, it feels to me that if we have capabilities it could be considered a bug to close something you don't own (anymore).

lib_eio/generic.mli
talex5: I got rid of the word `manually` completely. It was awkward in both places.

talex5:
> Intuitively, it feels to me that if we have capabilities it could be considered a bug to close something you don't own (anymore).

Eio's capabilities are not linear, so there's no problem (there are too many concepts called "capabilities"!).

PR #549, opened 2023-06-06
Restore close to socket type

talex5:
Now that double-close is permitted (since #547), we can simplify the types by making all sockets closable.

This reverts commit ccf147474ee3c2c222c2e9c7694b2c69445734c8.


avsm:
Just flagging to @anmonteiro that this will require adaptation in Gluten/H2 when the next eio release comes up. (something like https://github.com/avsm/eeww/commit/dfdd8f248b249254b63bc5f18f8ab172ec94ca0e )

PR #552, opened 2023-06-08
Remove deprecated features

talex5:


avsm (approved):
Prominent Changelog entry on release would be good.

PR #555, opened 2023-06-11
Add `listening_socket#listening_addr`

mefyl:
This adds the ability to retrieve the address a socket is listening on. More than just a convenience, this is required to let the OS pick a random free port by passing `0` and then determine what port was actually picked.

The actual implementation returns `None` if the socket was closed, while the mock always returns `None` - I'm unsure whether this is an issue as I'm only just starting to dip my toes in eio.

### DIFF
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -235,6 +235,10 @@ module Listening_socket = struct
     in
     let flow = (flow client :> _ Eio.Net.stream_socket) in
     flow, client_addr
+
+  let listening_addr fd =
+    Eio_unix.Fd.use_exn "listening_addr" fd
+      (fun fd -> Eio_unix.Net.sockaddr_of_unix_stream (Unix.getsockname fd))
 end
 
 let listening_handler = Eio_unix.Pi.listening_socket_handler (module Listening_socket)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+## v0.14
+
+New features / API changes:
+
+- Add `Net.listening_addr` (@mefyl #555).
+
 ## v0.13
 
 New features / API changes:
--- a/lib_eio/mock/eio_mock.mli
+++ b/lib_eio/mock/eio_mock.mli
@@ -132,8 +132,11 @@ module Net : sig
 
   val on_getnameinfo : t -> (string * string) Handler.actions -> unit
 
-  val listening_socket : string -> listening_socket
-  (** [listening_socket label] can be configured to provide mock connections. *)
+  val listening_socket :
+    ?listening_addr:Eio.Net.Sockaddr.stream -> string -> listening_socket
+  (** [listening_socket label] can be configured to provide mock connections.
+
+      If [listening_addr] is not provided, a dummy value will be reported. *)
 
   val on_accept :
     listening_socket ->
--- a/lib_eio/mock/net.ml
+++ b/lib_eio/mock/net.ml
@@ -97,14 +97,16 @@ type listening_socket = listening_socket_ty r
 module Listening_socket = struct
   type t = {
     label : string;
+    listening_addr : Eio.Net.Sockaddr.stream;
     on_accept : (Flow.t * Eio.Net.Sockaddr.stream) Handler.t;
   }
 
   type tag = [`Generic]
 
-  let make label =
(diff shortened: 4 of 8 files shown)


patricoferris:
Thanks! This looks sensible to me, I think the Eio_windows implementation would be the same as the rest using the `Unix` module.

As with other APIs I think a function wrapping the method call would be useful so people don't have to use `#` if they don't want to. This could then document the semantics of calling `listening_addr`.

talex5 (commented):
Looks reasonable. It's probably better to raise an exception if it's closed (I'd consider it to be a bug to use a socket after closing it, not something you need to write code to check for).

The mock could take a `?listening_addr` argument, with some suitable default dummy value (since most people won't care).

In any case, we should document this behaviour of passing 0 as the port to listen. Linux's `bind` manpage only seems to mention this in the errors section, under `EADDRINUSE`. https://stackoverflow.com/a/1077305/50926 has more information.

mefyl:
> Does the bind to 0 port trick work on macOS? It's not documented there, and I remember encountering issues with this some years ago when trying it for Docker for Desktop...

Looks like it does on some MacOS at least:
```sh
$ uname -a
Darwin mac1gitlabroutineco.local 22.5.0 Darwin Kernel Version 22.5.0: Thu Jun  8 22:22:19 PDT 2023; root:xnu-8796.121.3~7/RELEASE_ARM64_T8103 arm64
$ cat net.c 
#include <stdio.h>
#include <stdlib.h>
#include <netdb.h>

#define handle_error(msg) \
    do { perror(msg); exit(EXIT_FAILURE); } while (0)

int
main(int argc, char *argv[])
{
  int sock = socket(AF_INET, SOCK_STREAM, 0);
  if (sock == -1)
    handle_error("socket");
  struct sockaddr_in addr;
  addr.sin_family = AF_INET;
  addr.sin_addr.s_addr = htonl(INADDR_ANY);
  addr.sin_port = htons(0);
  if (bind(sock, (struct sockaddr *) &addr, sizeof addr) == -1)
    handle_error("bind");
  if (listen(sock, 5) == -1)
    handle_error("listen");
  socklen_t length = sizeof addr;
  if (getsockname(sock,(struct sockaddr *) &addr, &length) == -1) 
    handle_error("getsockname");
  printf("listening on %d\n", ntohs(addr.sin_port));
}
$ gcc net.c -o net
$ ./net
listening on 49176
$ ./net
listening on 49177
$ ./net
listening on 49178
$ ./net
listening on 49179
$ 
```

To be frank I always assumed it was part of POSIX, I'm surprised it's that sparsely documented. I've always used it as the standard way to listen on a random port, especially in test suites that were running on multiples architecture, including macs. But it looks indeed like it's not that clear cut.

mefyl:
May I humbly ask for a status on this ? It's something I rely upon a lot, especially in test suites. I don't mind porting it to the current API if we're willing to accept the feature.

lib_eio_linux/eio_linux.ml
```diff
@@ -235,6 +235,11 @@ module Listening_socket = struct
     in
     let flow = (flow client :> _ Eio.Net.stream_socket) in
     flow, client_addr
+
+  let listening_addr fd =
+    Eio_unix.Fd.use fd
+      (fun fd -> Eio_unix.Net.sockaddr_of_unix_stream (Unix.getsockname fd))
+      ~if_closed:(fun () -> failwith "listening_addr: socket is closed")
```
talex5, line 242:
```suggestion
    Eio_unix.Fd.use_exn "listening_addr" fd
      (fun fd -> Eio_unix.Net.sockaddr_of_unix_stream (Unix.getsockname fd))
```

talex5 (commented):
Looks good, but there are a load of accidental changes to `CHANGES.md` (not sure how it happened; `.gitattributes` should have prevented it).

nojb:
Hello, I recently needed this (see #668); note that this should also support datagram (udp) sockets to be complete.

talex5 (approved):
Looks good, once CI has finished. I squashed and rebased and updated to `use_exn` in the other places.

(Note: GitHub doesn't send notifications when you add commits, so it's helpful to post a comment when a PR is ready for review again.)

@nojb: I agree it would be good to have this for datagram sockets too, but I'm happy for that to be a separate PR.

talex5:
> To be frank I always assumed it was part of POSIX, I'm surprised it's that sparsely documented.

Apparently the bind 0 behaviour has just been added: https://sortix.org/blog/posix-2024/

PR #561, opened 2023-06-19
Raise an exception instead of Failure "execve: Argument list too long"

### LINKED ISSUE #556: "execve: Argument list too long" should not be a `Failure`
Currently, when passing arguments to a program with the new `Eio.Process` API, `E2Big` errors are propagated as `Failure "execve: Argument list too long"`.
Programs using this API should be able to handle such an error. In my case, I am using this in an interpreter, where I would like to attach a call trace to the exception.
The current `Failure` exception effectively makes it impossible to do this robustly, since the error message might change at any point (which is not unreasonable to expect in this case, since the current message is pretty misleading. The real error is that the *total number of bytes* is too large, not that there are too many arguments).

Innf107:
Fix #556

### DIFF
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -13,6 +13,7 @@ let pp_status ppf = function
 type error =
   | Executable_not_found of string
   | Child_error of exit_status
+  | Argument_list_too_long
 
 type Exn.err += E of error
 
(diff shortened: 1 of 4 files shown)


talex5:
This string matching looks a bit fragile (OS/locale dependent). Perhaps we should get `action_execve` to output errno and just match on that (e.g. `execve:7:...`)?

`strerrorname_np` is GNU-specific, but we can use `caml_unix_error_of_code` to convert integer codes to OCaml values.

talex5:
Closing in favour of #854.

PR #566, opened 2023-06-24
Add resource pool

darrenldl:
Split from #517 to focus on just resource pool

### DIFF
--- a/lib_eio/pool.ml
+++ b/lib_eio/pool.ml
@@ -1,0 +1,154 @@
+type 'a check = 'a -> bool
+
+type 'a dispose = 'a -> unit
+
+type 'a handlers = {
+  check : 'a check;
+  dispose : 'a dispose;
+}
+
+let noop_handlers : 'a handlers = {
+  check = (fun _ -> true);
+  dispose = (fun _ -> ());
+}
+
+type 'a ready = 'a * 'a handlers
+
+type 'a alloc = unit -> 'a ready
+
+type 'a t = {
+  max_size : int;
+  alloc_budget : Semaphore.t;
+  alloc : 'a alloc;
+  waiting : 'a ready Promise.u option Stream.t;
+  ready : 'a ready Stream.t;
+  (* Each runner is given a copy of the clear signal
+     to be checked after its run
+
+     The copy to use is replaced after each invocation
+     of [clear], so we can distinguish between
+     clearing of different batch of runs, e.g.
+
+     1. use (clear signal version 0)
+     2. use (clear signal version 0)
+     3. clear
+     4. use (clear signal version 1)
+
+     clear at 3 should apply to use at 1 and 2, but not use at 4
+  *)
+  mutable clear_signal : bool Atomic.t;
+  shutdown : bool Atomic.t;
+}
+
+let start_monitor ~sw (t : 'a t) : unit =
+  let rec aux () =
+    match Stream.take t.waiting with
+    | None -> ()
+    | Some resolver -> (
+        Semaphore.acquire t.alloc_budget;
+        let exn =
+          match Stream.take_nonblocking t.ready with
+          | None -> (
+              match t.alloc () with
+              | x -> (
+                  Promise.resolve resolver x;
+                  None
+                )
+              | exception exn -> Some exn
+            )
+          | Some x -> (
+              Promise.resolve resolver x;
+              None
+            )
+        in
+        match exn with
+        | Some exn -> Switch.fail sw exn
+        | None -> aux ()
+      )
+  in
+  Fiber.fork ~sw aux
+
+let create
+    ~sw
+    ~(alloc : 'a alloc)
+    max_size
+  : 'a t =
+  if max_size <= 0 then (
+    invalid_arg "Pool.create: max_size is <= 0"
+  );
+  let t =
+    {
+      max_size;
+      alloc_budget = Semaphore.make max_size;
+      alloc;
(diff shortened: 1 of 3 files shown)


lib_eio/pool.ml
```diff
@@ -0,0 +1,154 @@
...
+
+     clear at 3 should apply to use at 1 and 2, but not use at 4
+  *)
+  mutable clear_signal : bool Atomic.t;
+  shutdown : bool Atomic.t;
+}
+
+let start_monitor ~sw (t : 'a t) : unit =
+  let rec aux () =
+    match Stream.take t.waiting with
```
patricoferris, line 45: With this current API, a pool must be explicitly `shutdown` in order for the forked monitoring fiber to finish and for the program to move past whatever switch is used to create the pool. I hit this problem right away when using the pool:

```ocaml
open Eio

let () =
  Eio_main.run @@ fun _ ->
  let test_pool () =
    Eio.Switch.run @@ fun sw ->
    let p = Pool.create ~sw ~alloc:(fun () -> (), Pool.noop_handlers) 2 in
    let use () =
      Eio.Fiber.fork ~sw @@ fun () ->
      Pool.use p (fun () -> Eio.traceln "Run!"; Eio_unix.sleep 2.)
    in
    let runs = [ (); (); (); (); (); () ] in
    List.iter use runs;
    (* Pool.shutdown p *)
  in
  test_pool ();
  Eio.traceln "All done"
```

This program hangs without explicitly shutting down the pool. I think this might surprise some people, is there a way to avoid this?

lib_eio/pool.mli
```diff
@@ -0,0 +1,24 @@
...
+
+val create :
+  sw:Switch.t ->
+  alloc:(unit -> 'a * 'a handlers) ->
+  int ->
+  'a t
+
+val use : 'a t -> ('a -> 'b) -> 'b
+
+val async : sw:Switch.t -> 'a t -> ('a -> unit) -> unit
```
patricoferris, line 18: Do you think these are necessary (i.e. do they need to be exposed to the user, I realise they're used internally) ? In my mind, the `use` function is the main one and I can decide to fork the fiber that calls it should I need to. Otherwise we've kinda added `fork` and `fork_promise` with two slightly different names (using `async`) which feels a little cluttered ?

lib_eio/pool.mli
```diff
@@ -0,0 +1,24 @@
...
+
+val async_promise : sw:Switch.t -> 'a t -> ('a -> 'b) -> 'b Promise.or_exn
+
+val clear : 'a t -> unit
```
patricoferris, line 22: I don't quite understand when `clear` would be used, this calls the `dispose` function for the resource, how would people use this in practice? I think a few test cases would help here.

patricoferris (commented):
Thanks @darrenldl!

I've added a few comments, generally on the API rather than the actual implementation for now. I think some MDX tests would help us decide the API too :))

lib_eio/pool.mli
darrenldl: I think I just copied this from Lwt_pool. Indeed not very obvious how to use this in practice - I think removing it might be best.

lib_eio/pool.mli
darrenldl: Ah good point. Leaving the forking responsibilities to users gives a cleaner picture yes. They did feel a bit sloppy and redundant when I was writing it.

lib_eio/pool.ml
darrenldl: This actually caught myself off guard as well, so I agree it is not very good design.

I think `shutdown` can be replaced by polling from Eio switch? I'll have to look at other Eio modules.

darrenldl:
Thanks for feedback! I'll revisit this PR in a few weeks.

lib_eio/pool.ml
patricoferris: I wonder if the pool itself needs to spawn the monitoring fiber, or if it just needs to return a resource when a user calls `use` which would yield when the pool is out of resources. This would avoid needing to pass in this switch.

darrenldl: I think my original intention was to make analysis simple, especially in anticipation of faults in allocation, release etc. I don't know how faithful I've stuck to that intention, and it could be the case that we can leave the responsibilities to `use`.

I'm leaning toward having a monitor for simpler code (in my attempts thus far), but would not mind swapping around approaches.

patricoferris: You might be able to keep most things the same and just use `Fiber.fork_daemon` instead and this wouldn't (I think) be an issue anymore https://ocaml.org/p/eio/latest/doc/Eio/Fiber/index.html#val-fork_daemon

darrenldl: Ah right, it carries the exact behaviour we want for monitor. Thanks!

talex5:
@darrenldl: In #602 I've made a lock-free pool using the Cells module. I think the extra features of this PR (e.g. the monitor fiber) could probably be built on top of that - would that work for your use?

darrenldl:
@talex5 The monitor is just there to make sure if some acquired resources have become unusable after use or an exception occurred during use, the pool is filled back in. But I think the `validate` code and exception handling in your `run_with` fulfills that role already.

I think your version (along with the simpler API) is good - this PR mainly tried to match Lwt pool features, some of which might not be that usable in practice to begin with.

Thanks for adding `Eio.Pool`!

PR #571, opened 2023-07-03
Preserve backtraces across Domain_manager.run

talex5:
`Stdlib.Domain.join` loses the backtrace, so take a copy in the child domain and re-raise in the parent.


avsm:
This seems like its worth reporting in upstream OCaml as well; at least as a documentation issue.

PR #574, opened 2023-07-07
Eio_posix: fix update to watched FDs on cancel

### LINKED ISSUE #572: POSIX backend keeps polling fd after close and ignores POLLNVAL

talex5:
If cancelling removes the last waiter, we need to update the set of FDs being watched.

Fixes #572.

Reported by @quernd

### DIFF
--- a/eio_windows.opam
+++ b/eio_windows.opam
@@ -30,4 +30,4 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/ocaml-multicore/eio.git"
-available: [os-family = "windows"]
+#available: [os-family = "windows"]
--- a/eio_windows.opam.template
+++ b/eio_windows.opam.template
@@ -1,1 +1,1 @@
-available: [os-family = "windows"]
+#available: [os-family = "windows"]
--- a/lib_eio_posix/sched.ml
+++ b/lib_eio_posix/sched.ml
@@ -145,6 +145,7 @@ let resume t node =
 
 (* Called when poll indicates that an event we requested for [fd] is ready. *)
 let ready t _index fd revents =
+  assert (not Poll.Flags.(mem revents pollnval));
   if fd == t.eventfd_r then (
     clear_event_fd t
     (* The scheduler will now look at the run queue again and notice any new items. *)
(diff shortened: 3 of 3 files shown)


anmonteiro:
This also fixes an issue I was seeing where EIO reported successful writes on closed file descriptors.

talex5:
> This also fixes an issue I was seeing where EIO reported successful writes on closed file descriptors.

Closing a file prevents new write operations from starting, but existing ones may be allowed to complete. I don't see why this PR should affect that; the bug will cause an extra wake-up but the write should just get EAGAIN and go back to sleep immediately.

quernd:
I was a bit nervous about the `assert` but it appears the logic is sound. I ran some long-running load tests and the program did not crash or exhibit any pathological behaviour.
Thanks for the quick response 👍

PR #575, opened 2023-07-07
add Eio_unix.Sockopt for setting/getting socket options

### LINKED ISSUE #376: Support retrieving socket options
As a corollary to https://github.com/ocaml-multicore/eio/issues/322, it would be useful to provide `getsockopt()` functionality.

avsm:
There are some socket options that are Linux-specific which are quite handy to have.  This adds a Eio_unix.Sockopt module which works on Eio_unix.Fd.t values.

This is a draft PR to check on the interface: the current PR replaces the need for Unix.setsockopt and has a single simple GADT for the socket options.  Alternatively we could just expose the extended ones in the same style as upstream OCaml as a separate set of types, too.  Opinions welcome.

### DIFF
--- a/lib_eio/net.mli
+++ b/lib_eio/net.mli
@@ -23,7 +23,7 @@ type error =
     (** This is a wrapper for epipe, econnreset and similar errors.
         It indicates that the flow has failed, and data may have been lost. *)
   | Connection_failure of connection_failure
-  | Invalid_option (** Option not supported by backend. *)
+  | Invalid_option (** Socket option not supported. *)
 
 type Exn.err += E of error
 
@@ -138,6 +138,11 @@ module Sockopt : sig
 
   type _ t = ..
 
+  val pp : 'a t Fmt.t
+  val pp_binding : ('a t * 'a) Fmt.t
+
+  (** {2 Common options} *)
+
   type _ t +=
     | SO_DEBUG : bool t         (** Enable socket debugging *)
     | SO_BROADCAST : bool t     (** Permit sending of broadcast messages *)
@@ -156,8 +161,74 @@ module Sockopt : sig
     | SO_RCVTIMEO : float t     (** Receive timeout *)
     | SO_SNDTIMEO : float t     (** Send timeout *)
 
-  val pp : 'a t Fmt.t
-  val pp_binding : ('a t * 'a) Fmt.t
+  (** {2 Linux-specific options} *)
+
+  type _ t +=
+    | TCP_CORK : bool t
+    (** When enabled, partial frames are not sent out.
+        Data is only sent when the option is disabled or the buffer becomes full. *)
+    | TCP_KEEPIDLE : int t
+    (** Time (in seconds) the connection needs to remain idle before TCP starts sending keepalive probes. *)
+    | TCP_KEEPINTVL : int t
+    (** Interval (in seconds) between individual keepalive probes. *)
+    | TCP_KEEPCNT : int t
+    (** Maximum number of keepalive probes TCP should send before dropping the connection. *)
+    | TCP_USER_TIMEOUT : int t
+    (** Maximum time (in milliseconds) that transmitted data may remain unacknowledged
+        before TCP will forcibly close the connection. *)
+    | TCP_MAXSEG : int t
+    (** The maximum segment size for outgoing TCP packets. If set before connection
+        establishment, it also changes the MSS value announced to the other end. *)
+    | TCP_LINGER2 : int option t
+    (** Lifetime of orphaned FIN_WAIT2 sockets, in seconds.
+
+        On set, [None] or [Some 0] requests the system default; [Some n] with
+        [n >= 0] keeps orphaned sockets in FIN_WAIT2 for [n] seconds;
+        negative values raise {!Invalid_argument}.
+
+        On get, [Some n] is the effective timeout and [None] means
+        FIN_WAIT2 lingering has been disabled. *)
+    | TCP_DEFER_ACCEPT : int t
+    (** Allow a listener to be awakened only when data arrives on the socket.
+        Value is the maximum time in seconds to wait for data. *)
+    | TCP_CONGESTION : string t
+    (** Set the TCP congestion control algorithm to be used (e.g., "cubic", "reno").
+        Unprivileged processes are restricted to algorithms in tcp_allowed_congestion_control. *)
+    | TCP_SYNCNT : int t
+    (** Set the number of SYN retransmits that TCP should send before aborting
+        the attempt to connect. Cannot exceed 255. *)
+    | TCP_WINDOW_CLAMP : int t
+    (** Bound the size of the advertised window to this value.
+        The kernel imposes a minimum size. *)
+    | TCP_QUICKACK : bool t
+    (** Enable quickack mode if set or disable if cleared. In quickack mode,
+        acks are sent immediately rather than delayed. This flag is not permanent. *)
+    | TCP_FASTOPEN : int t
+    (** Enable Fast Open (RFC 7413) on the listener socket. The value specifies
+        the maximum length of pending SYNs (similar to backlog in listen). *)
+    | IP_FREEBIND : bool t
+    (** Allow binding to an IP address that is nonlocal or does not (yet) exist.
+        This permits listening on a socket without requiring the underlying
+        network interface to be up. *)
+    | IP_BIND_ADDRESS_NO_PORT : bool t
+    (** Inform the kernel to not reserve an ephemeral port when using bind()
+        with a port number of 0. The port will be chosen at connect() time. *)
+    | IP_LOCAL_PORT_RANGE : (int * int) t
+    (** Set the per-socket local port range as (lower_bound, upper_bound).
+        Both bounds are inclusive and must be in range 0-65535.
+        Use (0, 0) to reset to system defaults. *)
+    | IP_TTL : int t
+    (** Set or retrieve the time-to-live field used in every packet sent from this socket.
+        Valid range is 1-255. *)
+    | IP_MTU : int t
+    (** Retrieve the current known path MTU of the current socket.
+        Only valid for connected sockets. This is a read-only option. *)
+    | IP_MTU_DISCOVER : [`Want | `Dont | `Do | `Probe] t
+    (** Set or receive the Path MTU Discovery setting for a socket.
+        - [`Want]: Use per-route settings (fragment if needed)
+        - [`Dont]: Never do Path MTU Discovery
+        - [`Do]: Always do Path MTU Discovery (reject large datagrams with EMSGSIZE)
+        - [`Probe]: Set DF but ignore Path MTU (for diagnostic tools) *)
 
   (** {2 Registering printers (for backend use only)} *)
 
--- a/lib_eio_linux/eio_stubs.c
+++ b/lib_eio_linux/eio_stubs.c
@@ -16,12 +16,16 @@
 #include <signal.h>
 #include <string.h>
 #include <unistd.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <netinet/tcp.h>
 
 // We need caml_convert_signal_number
 #define CAML_INTERNALS
 
 #include "primitives.h"
 
+#include <caml/fail.h>
 #include <caml/mlvalues.h>
 #include <caml/memory.h>
 #include <caml/alloc.h>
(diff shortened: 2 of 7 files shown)


lib_eio/unix/eio_unix.mli
```diff
@@ -84,6 +84,20 @@ module Stdenv : sig
...
 
+module Sockopt : sig
+  type 'a t = 'a Eio_unix__Sockopt.t =
+   | SO_KEEPALIVE : bool t
+   | SO_REUSEADDR : bool t
+   | SO_REUSEPORT : bool t
+   | TCP_CORK : int t
+   | TCP_KEEPCNT : int t
+   | TCP_KEEPIDLE : int t
+   | TCP_KEEPINTVL : int t
```
anmonteiro, line 95: could you also include TCP_NODELAY?

avsm: Yeah, I'll add the complete set and tests once we decide on the interface.

anmonteiro: On the interface: seems easy to use and I like the GADT approach for the options themselves.

I really like having just a single function to set socket options.

lib_eio/unix/eio_unix.mli
```diff
@@ -84,6 +84,20 @@ module Stdenv : sig
       You can use the functions in {!Eio.Stdenv} to access these fields if you prefer. *)
 end
 
+module Sockopt : sig
```
talex5, line 87: Would probably be better as `Eio_unix.Net.Sockopt`.

lib_eio/unix/eio_unix.mli
```diff
@@ -84,6 +84,20 @@ module Stdenv : sig
...
+  type 'a t = 'a Eio_unix__Sockopt.t =
```
talex5, line 88: Why refer to the `Eio_unix__Sockopt` internal module here?

talex5 (commented):
Seems like socket options could be useful on non-Unix systems too. We could define the type in Eio.Net as an extensible variant and put the common ones there, adding Unix specific ones in Eio_unix.Net, etc. Would have to decide what to do about unsupported options.

avsm:
> Seems like socket options could be useful on non-Unix systems too. We could define the type in Eio.Net as an extensible variant and put the common ones there, adding Unix specific ones in Eio_unix.Net, etc. Would have to decide what to do about unsupported options.

SGTM.  Right now, if a given sockopt isn't defined, it becomes (-1) and a call to it raises EINVAL (similar behaviour to  the OCaml Unix module).  

I'll move the common definitions that work on Windows as well to Eio.Net, and then put the remaining ones in Eio_unix.

lib_eio/unix/eio_unix.mli
avsm: added TCP_NODELAY

lib_eio/unix/eio_unix.mli
avsm: Done in 92a06fe38597fae48c2f520d2d7d070e62cf23a3

lib_eio/unix/eio_unix.mli
avsm: Removed in latest patchset; all one extensible variant now.

talex5:
In #322, @create2000 is asking about helping to finish this. Are we happy with the current design? What needs to happen beyond rebasing it?

It looks to me like some of the options (those not specific to Unix) should move to `Eio.Net` (e.g. `TCP_NODELAY` would also make sense in a unikernel). Possibly all the options should move there; it's OK to talk about an option, even if it's not supported.

Possibly `eio_unix` should define e.g. `type _ t += Unix_bool of Unix.socket_bool_option : bool t`, so all existing OCaml options can be used.

Might need a way to pass a list of options when connecting or binding (since we do that in one step, and maybe it's too late to set some options by then?). Might want some fancy heterogeneous list to hold them, since they have various types.

@create2000 I don't know if it's a good first issue or not. It uses some fancy types and involves C code. On the other hand, it's mostly done!

patricoferris:
Re:design -- my impression is that we should definitely move parts to `Eio.Net` that are portable, and in this instance perhaps all of the options and add the `Unix` version to allow for all existing OCaml options to _just work_. I think we will want to be able to get/set after connecting/binding anyway so this seems like a reasonable chunk of work to have in a single PR (basically what is already here, shuffled around a little)?

Perhaps a follow up PR can work to support options being set at the point of connecting/binding (as you mentioned, this seems very related to #713)?

avsm:
I've pushed a rebase with all the suggestions here. There are loads of Linux and Windows specific ones to add, but I wanted to check if this is version is heading in the right direction first @talex5 @patricoferris

talex5 (commented):
This approach looks good to me.

avsm:
I've added a bunch of Linux specific calls now (skipping multicast for now, and some of the more exotic fast connect and tcp_info ones), and also dropped SO_ACCEPTCON as it's not portable to macOS with no reasonable alternative (have to do a listen and wait for an EINVAL). It can still be passed directly via the Unix passthrough, but it's not much use in Eio where you know the state of your sockets anyway.

lib_eio_linux/eio_stubs.c
```diff
@@ -258,3 +259,26 @@ CAMLprim value caml_eio_sockopt_int_get(value v_fd, value v_level, value v_optio
   if (ret == -1) uerror("getsockopt", Nothing);
   CAMLreturn(Val_int(val));
 }
+
+/* Socket option stubs for string options (e.g. TCP_CONGESTION) */
+CAMLprim value caml_eio_sockopt_string_set(value v_fd, value v_level, value v_option, value v_val) {
+  CAMLparam4(v_fd, v_level, v_option, v_val);
+  const char *str = String_val(v_val);
+  int ret = setsockopt(Int_val(v_fd), Int_val(v_level), Int_val(v_option),
+                       str, strlen(str) + 1);
```
talex5, line 268: Should probably  use `caml_string_length` to avoid problems with zero bytes.

lib_eio_linux/eio_stubs.c
```diff
@@ -258,3 +259,26 @@ CAMLprim value caml_eio_sockopt_int_get(value v_fd, value v_level, value v_optio
...
+
+CAMLprim value caml_eio_sockopt_string_get(value v_fd, value v_level, value v_option) {
+  CAMLparam3(v_fd, v_level, v_option);
+  CAMLlocal1(v_result);
+  char buffer[256];
+  socklen_t len = sizeof(buffer);
+  int ret = getsockopt(Int_val(v_fd), Int_val(v_level), Int_val(v_option),
+                       buffer, &len);
+  if (ret == -1) uerror("getsockopt", Nothing);
+  buffer[len < sizeof(buffer) ? len : sizeof(buffer) - 1] = '\0';
```
talex5, line 281: Likewise, I guess this should use `caml_alloc_initialized_string`.

avsm:
Squashed commits, should be ready for merge.

talex5 (commented):
(only reviewed the mock bit so far)

lib_eio/mock/sockopt.ml
```diff
@@ -0,0 +1,47 @@
...
+  | Eio.Net.Sockopt.SO_RCVLOWAT -> Fmt.str "SO_RCVLOWAT = %d" v
+  | Eio.Net.Sockopt.SO_SNDLOWAT -> Fmt.str "SO_SNDLOWAT = %d" v
+  | Eio.Net.Sockopt.SO_LINGER -> Fmt.str "SO_LINGER = %s" (match v with None -> "None" | Some n -> string_of_int n)
+  | Eio.Net.Sockopt.SO_RCVTIMEO -> Fmt.str "SO_RCVTIMEO = %f" v
+  | Eio.Net.Sockopt.SO_SNDTIMEO -> Fmt.str "SO_SNDTIMEO = %f" v
+  | _ -> "unknown"
+
+let setsockopt : type a. string -> a Eio.Net.Sockopt.t -> a -> unit = fun label opt v ->
+  let opt_desc = pp_sockopt_value opt v in
+  traceln "%s: setsockopt %s" label opt_desc
```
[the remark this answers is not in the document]
avsm: fixed in a44bdb54bd832008721aba1681cc4ea57d69ede2

lib_eio/mock/sockopt.ml
```diff
@@ -0,0 +1,47 @@
...
+  | Eio.Net.Sockopt.SO_REUSEPORT -> traceln "%s: getsockopt SO_REUSEPORT = false" label; false
+  | Eio.Net.Sockopt.SO_SNDBUF -> traceln "%s: getsockopt SO_SNDBUF = 0" label; 0
+  | Eio.Net.Sockopt.SO_RCVBUF -> traceln "%s: getsockopt SO_RCVBUF = 0" label; 0
+  | Eio.Net.Sockopt.SO_TYPE -> traceln "%s: getsockopt SO_TYPE = 0" label; 0
+  | Eio.Net.Sockopt.SO_RCVLOWAT -> traceln "%s: getsockopt SO_RCVLOWAT = 0" label; 0
+  | Eio.Net.Sockopt.SO_SNDLOWAT -> traceln "%s: getsockopt SO_SNDLOWAT = 0" label; 0
+  | Eio.Net.Sockopt.SO_LINGER -> traceln "%s: getsockopt SO_LINGER = None" label; None
+  | Eio.Net.Sockopt.SO_RCVTIMEO -> traceln "%s: getsockopt SO_RCVTIMEO = 0.0" label; 0.0
+  | Eio.Net.Sockopt.SO_SNDTIMEO -> traceln "%s: getsockopt SO_SNDTIMEO = 0.0" label; 0.0
+  | _ -> raise (Failure "Mock getsockopt not implemented")
```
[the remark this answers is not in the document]
avsm: I've left gets logging as I'd like to (in the future) have the mock layer remember the setsockopt values in a heterogenous map. I have lifted out the pp functions to be exposed in the main interface in a44bdb54bd832008721aba1681cc4ea57d69ede2 as that seems obviously useful

avsm:
Rebased to trunk, and some minor fixes also added while re-reviewing it.

avsm:
@talex5 @patricoferris I'm happy with the low-level implementation here, but I'm not sure if the Pi integration is right. We've got a new SOCKET module type now, and I think it won't be too onerous to mock, but I'm not sure...

avsm:
Need to ensure this is compatible with whatever approach we pick for #713 as well

talex5:
I've merged some of this separately as #858 and #859 and rebased the remaining bit (Linux options) on them.

I then added another commit with some suggested changes:

- Move all options to `Eio.Net.Sockopt` so you can try them, even if the backend might not be Linux.
- Simplify code and tests a bit.

lib_eio/net.mli
```diff
@@ -156,8 +161,74 @@ module Sockopt : sig
...
+    (** Enable Fast Open (RFC 7413) on the listener socket. The value specifies
+        the maximum length of pending SYNs (similar to backlog in listen). *)
+    | IP_FREEBIND : bool t
+    (** Allow binding to an IP address that is nonlocal or does not (yet) exist.
+        This permits listening on a socket without requiring the underlying
+        network interface to be up. *)
+    | IP_BIND_ADDRESS_NO_PORT : bool t
+    (** Inform the kernel to not reserve an ephemeral port when using bind()
+        with a port number of 0. The port will be chosen at connect() time. *)
+    | IP_LOCAL_PORT_RANGE : (int * int) t
```
avsm, line 216: This one needs a design opinion: I didn't wrap it in option as I felt that passing 0,0 is fine to reset it, but it can be done (at the cost of more complexity)

talex5: I agree; I don't think it's worth it.

lib_eio_linux/tests/network.md
```diff
@@ -0,0 +1,196 @@
...
+  try_setsockopt client Sockopt.TCP_LINGER2 (Some 110);
+
+  (*
+  try_getsockopt client Sockopt.TCP_CONGESTION;
+  (* Try to set cubic if available - may fail based on system config *)
+  (try
+    try_setsockopt client Sockopt.TCP_CONGESTION "cubic";
+  with _ ->
+    traceln "Could not change TCP_CONGESTION (normal for unprivileged)");
+  *)
```
talex5, line 77: I disabled this test, as the result depends on the host configuration (as noted in the comment).

lib_eio_linux/eio_stubs.c
```diff
@@ -246,3 +249,45 @@ CAMLprim value caml_eio_clone3(value v_errors, value v_actions) {
...
+
+CAMLprim value caml_eio_sockopt_string_get(value v_fd, value v_level, value v_option) {
+  CAMLparam3(v_fd, v_level, v_option);
+  CAMLlocal1(v_result);
+  char buffer[256];
+  socklen_t len = sizeof(buffer);
+  int ret = getsockopt(Int_val(v_fd), Int_val(v_level), Int_val(v_option),
+                       buffer, &len);
+  if (ret == -1) uerror("getsockopt", Nothing);
+  v_result = caml_copy_string(buffer);
```
talex5: I added `strnlen` here - in theory the result might not have a null terminator if there wasn't room in the buffer. Though since `TCP_CA_NAME_MAX` is 16 we'll never hit this problem with the current set of options.

talex5 (approved):

lib_eio_linux/eio_stubs.c
```diff
@@ -246,3 +249,45 @@ CAMLprim value caml_eio_clone3(value v_errors, value v_actions) {
...
+
+CAMLprim value caml_eio_sockopt_string_get(value v_fd, value v_level, value v_option) {
+  CAMLparam3(v_fd, v_level, v_option);
+  CAMLlocal1(v_result);
+  char buffer[256];
+  socklen_t len = sizeof(buffer);
+  int ret = getsockopt(Int_val(v_fd), Int_val(v_level), Int_val(v_option),
+                       buffer, &len);
+  if (ret == -1) uerror("getsockopt", Nothing);
+  v_result = caml_alloc_initialized_string(strnlen(buffer, len), buffer);
```
avsm: Nothing good comes from returning a truncated string; the caller wont be able to use it meaningully. Instead of strnlen, why not check `len` after the getsockopt and raise an exception if it's the same as sizeof(buffer)?

talex5: Yes, could do. Not sure what exception to raise, though.

avsm: Invalid_argument is reasonable here. It's a rare violation that indicates something's quite wrong, and not something expected to be caught and handled.

talex5: OK, done!

Issue #577, opened 2023-07-08
Feature request: Stream select


dermesser:
Thank you very much for your work on eio! I've thoroughly enjoyed working with it, despite being a newcomer to OCaml. Eio feels more advanced than the async frameworks in other languages I've used.

So far, I have been missing one feature: Waiting on multiple streams, which in languages like Go or Rust is called *select*. I've taken a look at the source code, but as a newbie I didn't want to charge ahead with some untenable solution (and besides, any possible implementation appeared non-trivial to me).

I don't have a specific API in mind, but something simple like

```ocaml
val select : ('a Stream.t * ('a -> 'b)) list -> 'b
```

to select between different channels producing items of the same type and permitting per-stream handling, or

```ocaml
val select2 : 'a Stream.t -> 'b Stream.t -> ('a, 'b) Either.t
```

for selecting between exactly two streams may already prove to be useful. I'd love to hear what you think.

anmonteiro:
Can you implement that in terms of `Fiber.first` and `Stream.take`?

dermesser:
> Can you implement that in terms of `Fiber.first` and `Stream.take`?

That's clever! I hadn't thought of this at all, but it obviously makes the implementation trivial. I will try to prepare a PR to see if it is met with interest.

Unfortunately the use case of `select2` described above appears to not be possible, though (it'd need a combination of `Fiber.pair` and `Fiber.first`, IIUC)

talex5:
Something like this needs to be atomic, which `Fiber.first` doesn't support.

The easy solution is to use [kcas](https://github.com/ocaml-multicore/kcas/blob/main/src/kcas_data/queue.mli) queues, which can be composed easily.

However, it would also be nice to have direct support in Eio.Stream too. The general idea is that you create an Atomic to hold the result and then CAS that to the value. There are two implementations of Stream, depending on whether it is zero-capacity or not. For the zero-capacity case (sync.ml) there is already support for rejecting a value (causing the producer to retry with the next client). For the non-zero-capacity case, you have a lock and can just decide not to pop the value, I think.

dermesser:
Thank you for the quick reply! I've been looking around a bit, mostly out of interest, and some questions came up. Keep in mind that it's likely I'm just missing your point here.

> The general idea is that you create an Atomic to hold the result and then CAS that to the value.

[...]

Issue #558, opened 2023-06-17
Feature request: an equivalent to `Lwt.npick`


SGrondin:
### Motivation

Under `Fiber.first`:
> Warning: it is always possible that both operations will succeed (and one result will be thrown away). This is because there is a period of time after the first operation succeeds, but before its fiber finishes, during which the other operation may also succeed.

This enables `Fiber.first` to have a nice, convenient API. 98% of the time, I do not care about the edge case, so the benefits of the API are worth the downside explained in the warning.

But 2% of the time, I do care about the return value of the second fiber that may also have succeeded.

#### `Lwt.pick`
```ocaml
Lwt.pick : ('a Lwt.t) list -> 'a Lwt.t
```
This function is roughly equivalent to `Fiber.any`.
> `Lwt.pick ps` returns a promise that is pending until *one* promise in the list `ps` becomes *resolved*.
When at least one promise in `ps` is resolved, `Lwt.pick` tries to cancel all other promises that are still pending, using `Lwt.cancel`.

#### `Lwt.npick`
```ocaml
Lwt.npick : ('a Lwt.t) list -> ('a list) Lwt.t
```
> `Lwt.npick ps` is similar to `Lwt.pick ps`, the difference being that when multiple promises in `ps` are fulfilled simultaneously (and none are rejected), the result promise is fulfilled with the *list* of values the promises were fulfilled with.
When at least one promise is rejected, `Lwt.npick` still rejects the result promise with the same exception.

### Draft idea

We're optimizing for the 2% case here. Something that behaves just like `Fiber.first`, but without losing the second successful result (when it succeeds """at the same time""").

```ocaml
Fiber.n_first : (unit -> 'a) -> (unit -> 'a) -> [ `Left of 'a | `Right of 'a | `Both of 'a * 'a ]
```
```ocaml
Fiber.n_any : (unit -> 'a) list -> 'a list
```

[The way Lwt accomplishes this](https://github.com/ocsigen/lwt/blob/cc05e2bda6c34126a3fd8d150ee7cddb3b8a440b/src/core/lwt.ml#L2836-L2892) is by scanning the list for other successful promises before resolving the returned promise.

I had a quick look at [the code](https://github.com/ocaml-multicore/eio/blob/a6c0be9affd1793613e5e23bba4258c9f56d2c36/lib_eio/core/fiber.ml#L92-L145) and it seems doable.

Is this a feature the maintainers would welcome?

patricoferris:
[...]

PR #578, opened 2023-07-09
Stream select

dermesser:
(for #577)

This is a proposal for a (simplistic) implementation of `Stream.select`, which permits watching multiple streams and returning the first item yielded by any of them. It works for the intended use case, although at this point there are two weaknesses, both due to the underlying implementation using `Fiber.any`:

* Only streams of the same item type can be selected across (just like `Fiber.any` only works with fiber functions yielding the same type)
* If two or more streams are ready at the same time, all but one elements will be discarded (just like `Fiber.any`). This is documented, but still not desirable.

Thanks for considering this - I'd be happy to hear proposals for making this functionality more general, or just not based on the `Fiber` functions which restrict what can be done here. So please consider it a starting point :-)

### DIFF
--- a/lib_eio/stream.mli
+++ b/lib_eio/stream.mli
@@ -43,7 +43,11 @@ val take_nonblocking : 'a t -> 'a option
 val select: ('a t * ('a -> 'b)) list -> 'b
 (** [select] waits for any stream to have an item available. The item
     is mapped by the provided function and returned. Example:
-    [select [(s1, fun x -> x+1); (s2, fun x -> x+2)] *)
+    [select [(s1, fun x -> x+1); (s2, fun x -> x+2)]
+
+    Warning: as with `Fiber.first`, it is possible that two or more streams
+    yield an item simultaneously, in which case only one item will be
+    returned, and the other items are discarded.*)
 
 val length : 'a t -> int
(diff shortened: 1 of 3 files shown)


lib_eio/stream.mli
```diff
@@ -40,6 +40,15 @@ val take_nonblocking : 'a t -> 'a option
...
     result may already be out-of-date by the time this returns. *)
 
+val select: ('a t * ('a -> 'b)) list -> 'b
+(** [select] waits for any stream to have an item available. The item
+    is mapped by the provided function and returned. Example:
+    [select [(s1, fun x -> x+1); (s2, fun x -> x+2)]
+
+    Warning: as with `Fiber.first`, it is possible that two or more streams
+    yield an item simultaneously, in which case only one item will be
+    returned, and the other items are discarded.*)
```
SGrondin, line 50: In my opinion, this caveat almost defeats the point of the function.
I would love to see `Eio.Stream.select` added, but I think it needs to be done using something like #558

talex5 (changes requested):
Thanks for prototyping this! As @SGrondin says, if we have this then it needs to be atomic (discussion is continuing in the original issue at #577).

PR #581, opened 2023-07-11
eio_posix: fix pread at end-of-file

### LINKED ISSUE #579: Problematic `assert()` in `Eio.File.pread`
I might be wrong, but [this assert](https://github.com/ocaml-multicore/eio/blob/57ace760036b7855efddd3e2a67aeae20288a511/lib_eio/file.ml#L68) appears misguided to me.
```ocaml
(* Current implementation of Eio.File.pread: *)

let pread (t : #ro) ~file_offset bufs =
  let got = t#pread ~file_offset bufs in
  assert (got > 0 && got <= Cstruct.lenv bufs);
  got
```

Imagine I'm reading a file slice by slice, using successive calls to `Eio.File.pread`.

When I reach the end of the file (i.e. when I pass a `~file_offset` equal to the number of bytes in the file), `Eio.File.pread` returns a length of `0` and thus the `assert` fails.

It appears to me that the author expected `t#pread` to raise `End_of_file` to indicate EOF.

Either the `Eio_posix` implementation of `Eio.File.ro#pread` needs to raise `End_of_file`, or the assertion should be
```ocaml
assert (got <= Cstruct.lenv bufs);
```

I might be wrong here! but at the moment I'm having to circumvent `Eio.File.pread` by making direct calls to `t#pread`.

talex5:
Reported by @SGrondin.

Fixes #579.

### DIFF
--- a/lib_eio_posix/flow.ml
+++ b/lib_eio_posix/flow.ml
@@ -71,7 +71,11 @@ let of_fd fd = object (_ : <Eio_unix.Net.stream_socket; Eio.File.rw>)
   method read_methods = []
   method copy src = copy src fd
 
-  method pread ~file_offset bufs = Low_level.preadv ~file_offset fd (Array.of_list bufs)
+  method pread ~file_offset bufs =
+    let got = Low_level.preadv ~file_offset fd (Array.of_list bufs) in
+    if got = 0 then raise End_of_file
+    else got
+
   method pwrite ~file_offset bufs = Low_level.pwritev ~file_offset fd (Array.of_list bufs)
 
   method stat = fstat fd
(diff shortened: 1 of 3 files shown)


lib_eio_posix/flow.ml
```diff
@@ -71,7 +71,11 @@ let of_fd fd = object (_ : <Eio_unix.Net.stream_socket; Eio.File.rw>)
   method read_methods = []
   method copy src = copy src fd
 
-  method pread ~file_offset bufs = Low_level.preadv ~file_offset fd (Array.of_list bufs)
+  method pread ~file_offset bufs =
+    let got = Low_level.preadv ~file_offset fd (Array.of_list bufs) in
```
avsm, line 75: The array/list mismatch in the low-level/high-level interface leads to more allocations, but I can't see how to get rid of that easily given what the underlying syscall expects.

avsm (approved):
LGTM

lib_eio_posix/flow.ml
SGrondin: I know I wouldn't mind the high-level interface exposing array. That way I can choose whether to convert my list to an array or adapt my algorithm to use array.

PR #584, opened 2023-07-12
Eio.Workpool

SGrondin:
For your consideration, this PR is an invitation to discuss the user-facing API and the implementation of a `Eio.Workpool` module.

Let's start with some general "guiding principles" that underpin the current design. I'll gladly make changes to the design if the maintainers invalidate any of these principles.

### Principles
P1: A workpool should not pay attention to `recommended_domain_count`. Users know their own workloads (we don't). The users will tell us how many domains to use.

P2: We need to support `m` concurrent jobs per domain. An efficient design is one that fully utilizes each core. For CPU-bound workloads, it means 1 job per thread. For IO-bound it's a lot more. There's also hybrid workloads, with CPU-demanding processing interspersed between IO calls. For those workloads, the right number is `1/p` where `p` is the proportion of each job being CPU-bound.

P3: The user knows the right moment to start and shutdown their workpools. They want the threads to be ready to go by the time the jobs come flying in. Having to spawn threads on the fly (lazily) should be avoided.

### Tests
We've got them! ~They test durations using monotonic clocks to validate that certain tasks execute concurrently. Despite it being an obvious race condition, the tests are solid and consistent, not flaky at all. I hope they'll behave that way in CI testing too.~ The tests are now fully deterministic using mock clocks, mock domains and the mock backend.

### Caveats
This PR uses `Fiber.n_any` which is only added as part of #587 so it obviously has to wait until the other one is merged.

---
The code is fairly short and I've added plenty of comments to help the reviewers.
I'll trim some comments once it's been reviewed.

Thank you for your time.

### DIFF
--- a/lib_eio/workpool.ml
+++ b/lib_eio/workpool.ml
@@ -1,0 +1,139 @@
+type job = Pack : (unit -> 'a) * ('a, exn) Result.t Promise.u -> job
+
+type action =
+  | Process of job
+  | Quit of {
+      atomic: int Atomic.t;
+      target: int;
+      all_done: unit Promise.u;
+    }
+
+(* Worker: 1 domain/thread
+   m jobs per worker, n domains per workpool *)
+
+type t = {
+  sw: Switch.t;
+  (* The work queue *)
+  stream: action Stream.t;
+  (* Number of domains. Depending on settings, domains may run more than 1 job at a time. *)
+  domain_count: int;
+  (* True when [Workpool.terminate] has been called. *)
+  is_terminating: bool Atomic.t;
+  (* Resolved when the workpool begins terminating. *)
+  terminating: action Promise.t * action Promise.u;
+  (* Resolved when the workpool has terminated. *)
+  terminated: unit Promise.t * unit Promise.u;
+}
+
+let reject (Pack (_, w)) = Promise.resolve_error w (Failure "Workpool.terminate called")
+
+(* This function is the core of workpool.ml.
+   Each worker recursively calls [loop ()] until the [terminating]
+   promise is resolved. Workers pull one job at a time from the Stream. *)
+let start_worker ~limit ~terminating stream =
+  Switch.run @@ fun sw ->
+  let capacity = Semaphore.make limit in
+  let run_job job w =
+    Fiber.fork ~sw (fun () ->
+        Promise.resolve w
+          (try Ok (job ()) with
+           | exn -> Error exn);
+        Semaphore.release capacity )
+  in
+  (* The main worker loop. *)
+  let rec loop () =
+    let actions = Fiber.n_any [ (fun () -> Promise.await terminating); (fun () -> Semaphore.acquire capacity; Stream.take stream) ] in
+    match actions with
+    | [ Process (Pack (job, w)) ] ->
+      (* We start the job right away. This also gives a chance to other domains
+         to start waiting on the Stream before the current thread blocks on [Stream.take] again. *)
+      run_job job w;
+      (loop [@tailcall]) ()
+    | Quit { atomic; target; all_done } :: maybe_job ->
+      List.iter
+        (function
+          | Process job -> reject job
+          | _ -> assert false)
+        maybe_job;
+        (* If we're the last worker terminating, resolve the promise
+           after waiting until the completion of all of this worker's jobs. *)
+      if Atomic.fetch_and_add atomic 1 = target
+      then Switch.on_release sw (Promise.resolve all_done)
+    | _ -> assert false
+  in
+  loop ()
+
+(* Start a new domain. The worker will need a switch, then we start the worker. *)
+let start_domain ~sw ~domain_mgr ~limit ~terminating ~transient stream =
+  let go () =
+    Domain_manager.run domain_mgr (fun () -> start_worker ~limit ~terminating stream )
+  in
+  (* [transient] workpools run as daemons to not hold the user's switch from completing.
+     It's up to the user to hold the switch open (and thus, the workpool)
+     by blocking on the jobs issued to the workpool.
+     [Workpool.submit] and [Workpool.submit_exn] will block so this shouldn't be a problem.
+     Still, the user can call [Workpool.create] with [~transient:false] to
+     disable this behavior, in which case the user must call [Workpool.terminate]
+     to release the switch. *)
+  match transient with
+  | false -> Fiber.fork ~sw go
+  | true ->
+    Fiber.fork_daemon ~sw (fun () ->
+        go ();
+        `Stop_daemon )
+
+let create ~sw ~domain_count ~domain_concurrency ?(transient = true) domain_mgr =
+  let stream = Stream.create 0 in
+  let instance =
+    {
+      sw;
+      stream;
+      domain_count;
+      is_terminating = Atomic.make false;
+      terminating = Promise.create ();
+      terminated = Promise.create ();
+    }
+  in
+  let terminating = fst instance.terminating in
+  for _ = 1 to domain_count do
+    start_domain ~sw ~domain_mgr ~limit:domain_concurrency ~terminating ~transient stream
+  done;
+  instance
+
+let submit_fork ~sw { stream; _ } f =
+  let p, w = Promise.create () in
+  Fiber.fork_promise ~sw (fun () ->
+      Stream.add stream (Process (Pack (f, w)));
+      Promise.await_exn p )
+
+let submit { stream; _ } f =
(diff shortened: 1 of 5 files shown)


SGrondin:
I just made a few tweaks: fixed a race condition and removed the Core-style `~f` named arguments since Eio doesn't normally label lambdas.

SGrondin:
I'm marking it as Ready For Review since I've fixed the race condition (using `Fiber.n_any` which is only available as part of #587  so this current PR obviously has to wait until the other one is merged)

SGrondin:
Is there a way to test this under dscheck? I looked over in tests/dscheck and I didn't see any use of domains or `env`, making me think it may not be realistic to do so.

lib_eio/workpool.mli
```diff
@@ -0,0 +1,37 @@
...
+(** Same as [run] but raises if the job failed. *)
+val run_exn : t -> (unit -> 'a) -> 'a
+
+(** Same as [run] but returns immediately, without blocking. *)
+val run_promise : sw:Switch.t -> t -> (unit -> 'a) -> ('a, exn) result Promise.t
+
+(** Waits for all running jobs to complete, then returns.
+    No new jobs are started, even if they were already enqueued.
+    To abort all running jobs instead of waiting for them, call [Switch.fail] on the Switch used to create this workpool *)
+val terminate : sw:Switch.t -> t -> unit
```
talex5, line 29: Why is there a `sw` argument here?

lib_eio/workpool.ml
```diff
@@ -0,0 +1,145 @@
...
+let run_exn instance f =
+  match run instance f with
+  | Ok x -> x
+  | Error exn -> raise exn
+
+let terminate ~sw ({ terminating = _, w1; terminated = p2, w2; _ } as instance) =
+  if Atomic.compare_and_set instance.is_terminating false true
+  then (
+    (* Instruct workers to shutdown *)
+    Promise.resolve w1 (Quit { atomic = Atomic.make 1; target = instance.domain_count; all_done = w2 });
```
talex5, line 129: This seems a bit over-complicated. I was expecting it to write `instance.domain_count` quit messages to the stream. That would avoid the need for a second channel for the quit message and the `n_any`.

lib_eio/workpool.mli
```diff
@@ -0,0 +1,37 @@
...
+val create :
+  sw:Switch.t ->
+  domain_count:int ->
+  domain_concurrency:int ->
+  ?transient:bool ->
+  #Domain_manager.t ->
+  t
+
+(** Run a job on this workpool. It is placed at the end of the queue. *)
+val run : t -> (unit -> 'a) -> ('a, exn) result
```
talex5, line 18:
```suggestion
val submit : t -> (unit -> 'a) -> ('a, exn) result
```
Makes it a bit more obvious we're using the pool rather than running it.

talex5 (commented):
> Is there a way to test this under dscheck?

This isn't doing any lock-free stuff itself, so dscheck isn't necessary. Instead, you can use the mock domain manager to make the tests deterministic. That currently lives in `network.md`, but it's generally useful and should be moved to the `eio.mock` library.

talex5:
>  Instead, you can use the mock domain manager to make the tests deterministic. That currently lives in network.md, but it's generally useful and should be moved to the eio.mock library.

I've moved it out in #610.

lib_eio/workpool.mli
SGrondin: Good catch, I'm not sure how I missed that.

lib_eio/workpool.ml
```diff
@@ -0,0 +1,145 @@
...
+    (* Reject all present and future queued jobs *)
+    Fiber.fork_daemon ~sw (fun () ->
```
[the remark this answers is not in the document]
SGrondin: Agreed on all points. I'm now using `~sw:instance.sw`, and yeah, while not great I think this background rejection loop is the most sensible approach for now.

lib_eio/workpool.ml
```diff
@@ -0,0 +1,145 @@
...
+        done;
+        `Stop_daemon );
+    (* Wait for all workers to have shutdown *)
+    Promise.await p2 )
+  else (* [Workpool.terminate] was called more than once. *)
+    Promise.await p2
+
+let is_terminating { terminating = p, _; _ } = Promise.is_resolved p
+
+let is_terminated { terminated = p, _; _ } = Promise.is_resolved p
```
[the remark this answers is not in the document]
SGrondin: It's there to allow for uncommon use cases where `terminate` is called before the switch release. My original idea for these 2 functions was to allow for rare use cases without forcing those users to keep track of all this through side channels.
But now I'm starting to think maybe having these 2 functions is a mistake because without them the user would be encouraged to create finer lifetimes instead of reusing the same (overly) long-lived Switch.
What do you think?

lib_eio/workpool.ml
SGrondin: The reason it's this way is due to a combination of factors:
- I don't want the Quit messages to have to wait for the other enqueued jobs to run and complete ahead of them
- As much as possible, calling `terminate` should "immediately" halt starting new jobs.
The obvious solution is to make `terminate` reject all queued jobs before enqueueing Quit messages, but that doesn't work well with the post-termination background rejection loop, because `terminate` becomes both a producer and a consumer while the workers are still consumers. It can be made to work, but the end result was more complex and less predictable.

SGrondin:
Thanks for the thorough review. I've either made the requested change or left a comment/question above. I'm now working on using a mock clock in tests.

tests/workpool.md
```diff
@@ -0,0 +1,338 @@
...
+    Fiber.both
+      (fun () -> Eio.Time.sleep_until clock t1)
+      (fun () ->
+        Fiber.yield ();
+        Fiber.yield ();
+        Fiber.yield ();
+        Fiber.yield ();
+        Fiber.yield ();
+        Fiber.yield ();
+        Fiber.yield ();
```
SGrondin, line 35: Is there a "yield until idle" function? If not, this is the precise number of yields needed...

talex5: There isn't. It needs support in `Eio_mock.Backend`. I was thinking of having a `Backend.run_full` that provides an environment with a mock clock that auto-advances whenever the run-queue is empty.
e.g.

```ocaml
Eio_mock.Backend.run @@ fun env ->
let clock = env#clock in
...
```

SGrondin: That would be fantastic to have! While reading other tests to see how mock clocks were used, I saw multiple opportunities where it could have been used.

wokalski: We also have mixed feelings about those explicit yields in the tests. A purist argument would be that it forces the writer to clearly express the semantics of a function... but in practice it is quite a headache.

lib_eio/workpool.ml
talex5: > I don't want the Quit messages to have to wait for the other enqueued jobs to run and complete ahead of them.

The workers can still check `is_terminating` before running a job and reject the job if it's set. I think that would have the same effect.

SGrondin: After trying it out, I'm remembering why it's not that way.
It's minor but I think it makes a difference.

By using a second channel, we're able to start the background rejection loop immediately. Otherwise, if all workers are fully occupied at the time `terminate` is called, we have to wait for a worker to be available to start rejecting jobs.

An alternative I've also explored is to immediately (T0) reject all queued jobs before enqueueing `n` Quit messages (T1), but that leaves jobs enqueued between T0 and T1 to hang until the background rejection job starts (which can only happen after all workers have exited so their Quit messages don't get dropped). This inconsistent behavior can be patched over by checking the `is_terminating` (`bool Atomic.t`) when submitting a new job but I'm trying to avoid all unnecessary thread coordination in the hot path...

talex5: I think we have to do that anyway, otherwise this could happen:
1. Client checks `is_terminating` and see it's still OK.
2. Pool gets terminated; all workers finish; switch ends.
3. Client adds job to queue.

It's not clear what behaviour we really want for terminate though. e.g. why do we want to reject jobs that were submitted before `terminate` was called? Then there's no way for a client to behave "correctly" (so that it's jobs never get rejected) in the graceful shutdown case.

It would probably be helpful to have an example of a program that needs support for this kind of graceful shutdown (rather than just finishing the switch, which I suspect will cover most real uses). Or possibly we should make an initial version without `terminate` and add it later if/when it's needed.

SGrondin:
Closing in favor of #639

PR #585, opened 2023-07-12
Implement simple atomic stream select

dermesser:
for #577, to supersede #578.

I want to emphasize that this change is based on my likely too superficial understanding of the library. It is likely that there exists a more elegant way to implement it.

* re: https://github.com/ocaml-multicore/eio/issues/577#issuecomment-1632321645 - this is currently worked around by  checking every stream for present items and starting to wait while holding its mutex; if no stream has any items, all streams must have capacity (this is asserted). Maybe there is a hole in my logic, but it sounded fine in my head.
* I had to modify the `Waiters` module slightly, which is what I had hoped to avoid by using multiple fibers previously. The way I modified it makes sense for this kind of application, but it could be that there is a race condition or logical trap I overlooked.
* There are too many in-line comments explaining what the too-long function `select_of_many` does, but maybe it helps you while reviewing it :-)
* Only `Locking` streams are handled at the moment, with an `assert` in place to enforce this. Mostly for the reason that I've wanted to get an initial implementation going before trying to handle everything.

Again, I'm not at all offended if a tighter implementation ends up replacing this PR, but my fingers were itching to try and implement it myself :-)


dermesser:
I'm currently also trying to understand the 0-capacity stream mechanics, and hopefully will come up with a way to (ideally) select out of a mixed set of streams.

lib_eio/stream.ml
```diff
@@ -94,6 +94,45 @@ module Locking = struct
...
+    let add_cancel_fn fn = cancel_fns := fn :: !cancel_fns in
+    let cancel_all () = List.iter (fun fn -> fn ()) !cancel_fns in
+    let wait ctx enqueue (t, f) = begin
+      Mutex.lock t.mutex;
+      (* First check if any items are already available and return early if there are. *)
+      if not (Queue.is_empty t.items)
+      then (
+        cancel_all ();
+        Mutex.unlock t.mutex;
+        enqueue (Ok (f (Queue.take t.items))))
```
talex5, line 109: Another domain could still have set `finished` to true using a different stream by the time this runs.

(also, you can't use `Queue.take` after unlocking)

dermesser: Apologies for missing this, it should have been obvious! Please take another look, it seemed not too difficult to fix.

dermesser:
I've added a small benchmark (f733045) which helped fix two bugs (the two preceding commits).

PR #586, opened 2023-07-19
Add `Process.run ?is_success` to control definition of success

SGrondin:
## Motivation

Some programs (ab)use exit codes in unconventional ways.

As an example I recently ran into: the `npm outdated` command checks a project's NPM dependencies and prints a status report (outdated, deprecations, etc).

But the exit code is 1 if any dependencies are in need of an update! While it **successfully** generated a status report, it still returned `1`. Different return codes may indicate "true failures". I disagree with the design of this command, I'm [not the only one](https://github.com/npm/rfcs/issues/473) and [it's causing issues](https://github.com/nodejs/help/issues/4188).

Regardless, this command is far from alone. Plenty of other programs behave in a similar way. Also, I may want to consider certain exit codes to be errors/success _depending_ on the `stdout` output of the command, for example.

## Suggestion

I suggest adding the following optional argument to `Eio.Process.run`: `?is_success:(int -> bool)`

Since it runs after the process has exited, the user is able to inspect the output of `stdout` and `stderr` to make a decision.

> #### But why not just call `Eio.Process.spawn` instead?

`Eio.Process.run` is a nicer interface, I wouldn't want to lose on (or have to recreate its) error handling. Most importantly though, `Process.spawn` is a leaky abstraction while `Process.run` is not. I don't have to stop and dig into Eio internals to use `Process.run`, I can quickly invoke it without breaking mental focus.

Adding `is_success` minimizes the number of use cases that require `Process.spawn` and (in my opinion) is a nice addition to `Process.run` that doesn't significantly increase complexity for the users nor maintainers.

<sub>PR suggestions > Issue suggestions, am I right? 😄 </sub>

### DIFF
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -39,9 +39,9 @@ end
 let pid proc = proc#pid
 let await proc = proc#await
 
-let await_exn proc =
+let await_exn ?(is_success = Int.equal 0) proc =
   match proc#await with
-  | `Exited 0 -> ()
+  | `Exited code when is_success code -> ()
   | status -> raise (err (Child_error status))
 
 let signal proc = proc#signal
@@ -84,18 +84,18 @@ let spawn ~sw (t:#mgr) ?cwd ?stdin ?stdout ?stderr ?env ?executable args =
     ?stdout:(stdout :> Flow.sink option)
     ?stderr:(stderr :> Flow.sink option)
 
-let run (t:#mgr) ?cwd ?stdin ?stdout ?stderr ?env ?executable args =
+let run (t:#mgr) ?cwd ?stdin ?stdout ?stderr ?(is_success = Int.equal 0) ?env ?executable args =
   Switch.run @@ fun sw ->
   let child = spawn ~sw t ?cwd ?stdin ?stdout ?stderr ?env ?executable args in
   match await child with
-  | `Exited 0 -> ()
+  | `Exited code when is_success code -> ()
   | status ->
     let ex = err (Child_error status) in
     raise (Exn.add_context ex "running command: %a" pp_args args)
 
 let pipe ~sw (t:#mgr) = t#pipe ~sw
 
-let parse_out (t:#mgr) parse ?cwd ?stdin ?stderr ?env ?executable args =
+let parse_out (t:#mgr) parse ?cwd ?stdin ?stderr ?is_success ?env ?executable args =
   Switch.run @@ fun sw ->
   let r, w = pipe t ~sw in
   try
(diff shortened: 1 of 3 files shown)


lib_eio/process.ml
```diff
@@ -84,12 +84,13 @@ let spawn ~sw (t:#mgr) ?cwd ?stdin ?stdout ?stderr ?env ?executable args =
     ?stdout:(stdout :> Flow.sink option)
     ?stderr:(stderr :> Flow.sink option)
 
-let run (t:#mgr) ?cwd ?stdin ?stdout ?stderr ?env ?executable args =
+let run (t:#mgr) ?cwd ?stdin ?stdout ?stderr ?is_success ?env ?executable args =
```
talex5, line 87: Just a minor thing, but it might be simpler to provide a default function rather than handling the option later:
```suggestion
let run (t:#mgr) ?cwd ?stdin ?stdout ?stderr ?(is_success=Int.equal 0) ?env ?executable args =
```

tests/process.md
```diff
@@ -148,6 +148,14 @@ Eio.Io Process Child_error Exited (code 3),
   running command: bash -c "exit 3" "" foo
 ```
 
+Exit code success can be determined by is_success:
+
+```ocaml
+# run @@ fun mgr env ->
+  Process.run ~is_success:(fun code -> Int.(code = 3)) mgr ["bash"; "-c"; "exit 3"; ""; "foo"];;
```
talex5, line 155: Was this supposed to be here?
```suggestion
  Process.run ~is_success:(fun code -> Int.(code = 3)) mgr ["bash"; "-c"; "exit 3"];;
```

talex5 (commented):
Thanks, this looks useful! We should probably do the same for `parse_out`.

lib_eio/process.ml
SGrondin: Good idea

tests/process.md
SGrondin: Will fix, thanks

SGrondin:
Thanks for the review @talex5 
For `parse_out` do you think it would be better to also add the argument to `await_exn` (which is exposed in the .mli) or just to `parse_out` itself?

talex5:
> For parse_out do you think it would be better to also add the argument to await_exn (which is exposed in the .mli) or just to parse_out itself?

Adding to `await_exn` (as an optional argument) makes sense to me.

talex5 (approved):
> All done, feel free to edit the .mli docs/comments, it was difficult to phrase

I pushed some minor changes (and also added another test).

Ready to merge once CI passes - thanks!

PR #589, opened 2023-07-21
WIP: Getting MDX tests working on Windows

polytypic:


README.md
```diff
@@ -754,6 +754,7 @@ Eio.Io Net Connection_failure Refused Unix_error (Connection refused, "connect",
 If we ran this using another backend, the `Unix_error` part might change.
 To avoid this problem, you can use `Eio.Exn.Backend.show` to hide the backend-specific part of errors:
 
+<!-- $MDX skip -->
 ```ocaml
 # Eio.Exn.Backend.show := false;;
```
polytypic, line 759: For some reason this test never completes.  So, I disabled it.

lib_eio_windows/dune
```diff
@@ -1,12 +1,13 @@
...
  (name eio_windows)
  (public_name eio_windows)
- (library_flags :standard -ccopt -lbcrypt -ccopt -lntdll)
+ (library_flags :standard -cclib -lbcrypt -cclib -lntdll)
  (enabled_if (= %{os_type} "Win32"))
  (foreign_stubs
   (language c)
   (include_dirs ../lib_eio/unix/include)
   (names eio_windows_stubs eio_windows_cstruct_stubs))
+ (c_library_flags :standard -lbcrypt -lntdll)
```
polytypic, line 10: This is required to allow the stubs to be loaded dynamically by MDX.

lib_eio_windows/dune
```diff
@@ -1,12 +1,13 @@
 (library
  (name eio_windows)
  (public_name eio_windows)
- (library_flags :standard -ccopt -lbcrypt -ccopt -lntdll)
+ (library_flags :standard -cclib -lbcrypt -cclib -lntdll)
```
polytypic, line 4: Fixed to use `-cclib` as the `-l` options are for the linker rather than the compiler.

README.md
```diff
@@ -752,12 +752,17 @@ Eio.Io Net Connection_failure Refused Unix_error (Connection refused, "connect",
...
 - : unit = ()
+```
+
+to hide the backend-specific part of errors:
 
+<!-- $MDX skip -->
+```ocaml
 # Eio_main.run @@ fun env ->
```
polytypic, line 766: For some reason this test never completes. So, I disabled it.

README.md
```diff
@@ -842,7 +847,10 @@ let try_mkdir path =
 The checks also apply to following symlinks:
 
 ```ocaml
-# Unix.symlink "dir1" "link-to-dir1"; Unix.symlink "/tmp" "link-to-tmp";;
+# Unix.symlink "dir1" "link-to-dir1";
+  Unix.symlink
+    (if Sys.win32 then Sys.getenv "TEMP" else "/tmp")
```
polytypic, line 852: Welcome to Windows!

talex5: We could perhaps use `Filename.get_temp_dir_name ()` here. We don't really care where it points.

README.md
talex5: That seems worrying! Is it only an MDX thing, or does the example in `examples/net` also hang if the server side isn't running?

README.md
```diff
@@ -904,6 +912,7 @@ ONE TWO THREE
 If you want to capture the output of a process, you can provide a suitable `Eio.Flow.sink` as the `stdout` argument,
```
polytypic, line 912: I disabled the process tests.  There are two problems:

1. Process operations are not yet implemented on Windows.
2. Even if the process operations were implemented on Windows, the tests would not work on Windows, because they use commands that do not necessarily exist on Windows.

talex5: Yeah. We really need a way to selectively disable tests on different platforms (e.g. https://github.com/realworldocaml/mdx/issues/393).

polytypic:
FYI, I have a deadline for another project at the end of this month, so I will need to concentrate on that one for the rest of this week.  I will continue work on Windows Eio bits after that.

We could consider adding some TODO notes to things disabled in this PR currently, merge it &mdash; so we'd have more Windows tests running on CI &mdash; and then continue with the TODOs in separate PRs.  Or we can continue work in this PR and not merge it yet.  Either way is fine with me.

PR #592, opened 2023-07-26
Clarify configuration for lib_eio_linux and enable tests on other arches

dra27:
Following on #432. ocaml/ocaml#12405 proposes using `linux` everywhere for the "system" variable. The change doesn't semantically affect the build, but it seemed worth a clarifying comment not to _rely_ on the old system values in future.

That said, the tests are presently not being run for it on non-amd64, so the second commit extends those as well.


talex5:
Would certainly be good to get this fixed! I guess we should wait for https://github.com/ocaml/ocaml/pull/12405 to be merged before merging this one.

Issue #510, opened 2023-05-08
Add missing file operations


talex5:
The FS API is currently missing lots of functions that are in OCaml's `Unix`:

- [ ] `stat` and `lstat` (you can stat an open file though)
- [ ] `fsync`
- [ ] `seek`
- [ ] `truncate`
- [ ] `map_file`
- [ ] `link`, `symlink`, `readlink`
- [ ] `realpath`
- [ ] `chmod`
- [ ] `chown`
- [ ] `access`

adatario:
Similarily it might be nice to expose something that would allow the creation of temporary files without going trough `env#fs`.

Maybe by exposing a `temp_dir` path in the `Stdenv` that uses `Unix.get_temp_dir_name` ?

yawaramin:
Or using `Filename.get_temp_dir_name` to avoid Unix dependency.

PR #594, opened 2023-07-27
Add Path.mkdirs

patricoferris:
As per one of the tasks in #510. I'm almost certainly doing something wrong... Unfortunately I don't have a windows machine so I blindly added the windows test.

### DIFF
--- a/lib_eio/path.ml
+++ b/lib_eio/path.ml
@@ -35,6 +35,27 @@ let mkdir ~perm ((t:#Fs.dir), path) =
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "creating directory %a" pp (t, path)
 
+let mkdirs ?(exists_ok=false) ~perm ((t:#Fs.dir), path) =
+  let separate = String.split_on_char (Filename.dir_sep.[0]) in
+  let make_directory p =
+    try (
+      if exists_ok then
+        try t#mkdir ~perm p with Exn.Io ((Fs.E Fs.Already_exists _), _) -> ()
+      else
+        t#mkdir ~perm p
+    ) with Exn.Io _ as ex ->
+      let bt = Printexc.get_raw_backtrace () in
+      Exn.reraise_with_context ex bt "creating directory %a" pp (t, p)
+  in
+  let rec loop acc = function
+    | [] -> ()
+    | x :: xs ->
+      let dir = Filename.concat acc x in
+      make_directory dir;
+      loop dir xs
+  in
+  loop "" (separate path)
+
 let read_dir ((t:#Fs.dir), path) =
   try List.sort String.compare (t#read_dir path)
   with Exn.Io _ as ex ->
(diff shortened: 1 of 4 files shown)


patricoferris:
Hmm I think the Linux error (or the macOS/Windows incorrect success) is caused by catching https://github.com/ocaml-multicore/eio/blob/355f8dabf28b229dc18285971ebbce62e2c4f3f1/lib_eio_posix/fs.ml#L158 hence the diff in the mdx test :/

lib_eio/path.ml
```diff
@@ -35,6 +35,27 @@ let mkdir ~perm ((t:#Fs.dir), path) =
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "creating directory %a" pp (t, path)
 
+let mkdirs ?(exists_ok=false) ~perm ((t:#Fs.dir), path) =
+  let separate = String.split_on_char (Filename.dir_sep.[0]) in
```
talex5, line 39: I think we decided in #124 that the path separator is always `/` in Eio paths.

lib_eio/path.ml
```diff
@@ -35,6 +35,27 @@ let mkdir ~perm ((t:#Fs.dir), path) =
...
+      Exn.reraise_with_context ex bt "creating directory %a" pp (t, p)
+  in
+  let rec loop acc = function
+    | [] -> ()
+    | x :: xs ->
+      let dir = Filename.concat acc x in
+      make_directory dir;
+      loop dir xs
+  in
+  loop "" (separate path)
```
talex5, line 57: It's probably more efficient to start by checking the direct parent. If that exists (which is the common) case, there's no need to check further up (see e.g. [Python's makedirs](https://github.com/python/cpython/blob/87de2fbb1e9b94f25ae8c0c233fb4690a13e4b02/Lib/os.py#L200) or [Bos's create](https://github.com/dbuenzli/bos/blob/6320f302eb9487518491511292a11ee6f04bd6e9/src/bos_os_dir.ml#L31)).

lib_eio/path.mli
```diff
@@ -89,7 +89,11 @@ val with_open_out :
 (** {1 Directories} *)
 
 val mkdir : perm:File.Unix_perm.t -> _ t -> unit
-(** [mkdir ~perm t] creates a new directory [t] with permissions [perm]. *)
+(** [mkdir ?exist_ok ~perm t] creates a new directory [t] with permissions [perm]. *)
+
+val mkdirs : ?exists_ok:bool -> perm:File.Unix_perm.t -> _ t -> unit
+(** Recursively create directories. If [exist_ok] is [false] (the default) then we raise
```
talex5, line 95: `exists_ok` should only affect the final directory, not the intermediate ones (most of which will exist).

lib_eio/path.ml
patricoferris: Oh right, thanks for pointing that out :))

talex5:
Closing in favour of the newer version in #625 - thanks!

PR #599, opened 2023-08-11
Alternative stat API

patricoferris:
(this PR was originally adding `stat` support, but that got split out and merged separately)

This is mostly a draft until https://github.com/ocaml-multicore/ocaml-uring/pull/95 is merged and released. But this is a continuation of https://github.com/talex5/eio/tree/stat adding support for linux. I'll see what I can do Windows... 

This will also be useful for #594 as a fast "exists" check I think.

### DIFF
--- a/lib_eio/path.mli
+++ b/lib_eio/path.mli
@@ -130,11 +130,115 @@ val read_dir : _ t -> string list
 
 (** {2 Metadata} *)
 
-val stat : follow:bool -> _ t -> File.Stat.t
-(** [stat ~follow t] returns metadata about the file [t].
+val exists : _ t -> bool
+(** [exists t] returns [true] if [t] is a regular file or directory, and [false]
+    if [t] does not exist or is some other {!type:File.kind}.
+    Raises an exception if the path cannot be accessed due to permissions. *)
+
+val is_file : _ t -> bool
+(** [is_file] returns [true] if [t] is a regular file, and [false]
+    if [t] does not exist or is some other {!type:File.kind}.
+    Raises an exception if the path cannot be accessed due to permissions. *)
+
+val is_directory : _ t -> bool
+(** [is_directory] returns [true] if [t] is a directory, and false
+    if it does not exist or is some other {!type:File.kind}.
+    Raises an exception if the path cannot be accessed due to permissions. *)
+
+val kind : follow:bool -> _ t -> File.kind
+(** [kind ~follow t] returns the kind of file that [t] is.
 
     If [t] is a symlink, the information returned is about the target if [follow = true],
-    otherwise it is about the link itself. *)
+    otherwise it is about the link itself.
+
+    Equivalent to [stat ~follow t [File.Kind] Fun.id],
+    so use {!stat} to query multiple items from a file in one call. *)
+
+val size : follow:bool -> _ t -> int64
+(** [size ~follow t] returns the size of [t].
+
+    If [t] is a symlink, the information returned is about the target if [follow = true],
+    otherwise it is about the link itself.
+
+    Equivalent to [stat ~follow t [File.Size] Fun.id],
+    so use {!stat} to query multiple items from a file in one call. *)
+
+val perm : follow:bool -> _ t -> int
+(** [perm ~follow t] returns the file permissions of [t].
+
+    If [t] is a symlink, the information returned is about the target if [follow = true],
+    otherwise it is about the link itself.
+
+    Equivalent to [stat ~follow t [File.Perm] Fun.id],
+    so use {!stat} to query multiple items from a file in one call. *)
+
+val uid : follow:bool -> _ t -> int64
+(** [uid ~follow t] returns the user id associated with [t].
+
+    If [t] is a symlink, the information returned is about the target if [follow = true],
+    otherwise it is about the link itself.
+
+    Equivalent to [stat ~follow t [File.Uid] Fun.id],
+    so use {!stat} to query multiple items from a file in one call. *)
+
+val gid : follow:bool -> _ t -> int64
+(** [gid ~follow t] returns the group id associated with [t].
+
+    If [t] is a symlink, the information returned is about the target if [follow = true],
+    otherwise it is about the link itself.
+
+    Equivalent to [stat ~follow t [File.Gid] Fun.id],
+    so use {!stat} to query multiple items from a file in one call. *)
+
+val atime : follow:bool -> _ t -> float
+(** [atime ~follow t] returns the last access time of [t] as the seconds
+    since the start of the epoch.
+
+    If [t] is a symlink, the information returned is about the target if [follow = true],
+    otherwise it is about the link itself.
+
+    Equivalent to [stat ~follow t [File.Atime] Fun.id],
+    so use {!stat} to query multiple items from a file in one call. *)
+
+val mtime : follow:bool -> _ t -> float
+(** [mtime ~follow t] returns the last modified time of [t] as the seconds
+    since the start of the epoch.
+
+    If [t] is a symlink, the information returned is about the target if [follow = true],
+    otherwise it is about the link itself.
+
+    Equivalent to [stat ~follow t [File.Mtime] Fun.id],
+    so use {!stat} to query multiple items from a file in one call. *)
+
+val ctime : follow:bool -> _ t -> float
+(** [ctime t] returns the creation time of [t] as the seconds
+    since the start of the epoch.
+
+    If [t] is a symlink, the information returned is about the target if [follow = true],
+    otherwise it is about the link itself.
+
+    Equivalent to [stat ~follow t [File.Ctime] Fun.id],
+    so use {!stat} to query multiple items from a file in one call. *)
+
+val stat : follow:bool -> _ t -> ('a, 'b) File.stats -> 'a -> 'b
+(** [stat ~follow t f k] returns metadata about the file [t], querying multiple
+    fields [f] in one call and applying them to the continuation [k].
+
+    If [t] is a symlink, the information returned is about the target if [follow = true],
+    otherwise it is about the link itself.
+
+    For example, to follow a symlink and print the kind, size and last modified
+    time of its target:
+
+    {[ stat ~follow:true t File.[ Kind; Size; Mtime ]
+        (fun kind size mtime ->
+          traceln "kind: %a size: %Ld mtime: %f"
+            File.pp_kind size mtime)
+     ]}
+
+    If you only require access to one field, consider using simpler accessor
+    functions in this module such as {!val:size} or {!val:kind}.  *)
+
 
 (** {1 Other} *)
 
--- a/lib_eio_linux/low_level.ml
+++ b/lib_eio_linux/low_level.ml
@@ -303,36 +303,6 @@ let openat ~sw ?seekable ~access ~flags ~perm dir path =
   | Cwd -> openat2 ~sw ?seekable ~access ~flags ~perm ~resolve:Uring.Resolve.beneath path
   | Fs -> openat2 ~sw ?seekable ~access ~flags ~perm ~resolve:Uring.Resolve.empty path
 
-let fstat t =
-  (* todo: use uring  *)
-  try
-    let ust = Fd.use_exn "fstat" t Unix.LargeFile.fstat in
-    let st_kind : Eio.File.Stat.kind =
-      match ust.st_kind with
-      | Unix.S_REG  -> `Regular_file
-      | Unix.S_DIR  -> `Directory
-      | Unix.S_CHR  -> `Character_special
-      | Unix.S_BLK  -> `Block_device
-      | Unix.S_LNK  -> `Symbolic_link
-      | Unix.S_FIFO -> `Fifo
-      | Unix.S_SOCK -> `Socket
-    in
-    Eio.File.Stat.{
-      dev     = ust.st_dev   |> Int64.of_int;
-      ino     = ust.st_ino   |> Int64.of_int;
-      kind    = st_kind;
-      perm    = ust.st_perm;
-      nlink   = ust.st_nlink |> Int64.of_int;
-      uid     = ust.st_uid   |> Int64.of_int;
-      gid     = ust.st_gid   |> Int64.of_int;
-      rdev    = ust.st_rdev  |> Int64.of_int;
-      size    = ust.st_size  |> Optint.Int63.of_int64;
-      atime   = ust.st_atime;
-      mtime   = ust.st_mtime;
-      ctime   = ust.st_ctime;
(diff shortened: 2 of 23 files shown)


lib_eio_posix/eio_posix_stubs.c
```diff
@@ -27,8 +27,10 @@
 
 #ifdef ARCH_SIXTYFOUR
 #define Int63_val(v) Long_val(v)
+#define caml_copy_int63(v) Val_long(v)
 #else
 #define Int63_val(v) (Int64_val(v)) >> 1
+#define caml_copy_int63 caml_copy_int64
```
talex5, line 33: Int63 uses the *top* 63-bits (this was a mistake in my branch):
```suggestion
#define caml_copy_int63(v) caml_copy_int64(v << 1)
```

lib_eio_posix/eio_posix_stubs.c
```diff
@@ -161,6 +163,68 @@ CAMLprim value caml_eio_posix_mkdirat(value v_fd, value v_path, value v_perm) {
...
+
+  caml_unix_check_path(v_path, "fstatat");
+  path = caml_stat_strdup(String_val(v_path));
+  caml_enter_blocking_section();
+  ret = fstatat(Int_val(v_fd), path, &statbuf, Int_val(v_flags));
+  caml_leave_blocking_section();
+  caml_stat_free_preserving_errno(path);
+  if (ret == -1) uerror("fstatat", v_path);
+
+  v_ret = caml_alloc_small(12, 0);
```
talex5, line 206:
```suggestion
  v_ret = caml_alloc(12, 0);
```
This ensures the new record is populated with GC-safe values (I already fixed this in the uring PR).

lib_eio_linux/low_level.ml
```diff
@@ -366,6 +375,33 @@ let mkdir_beneath ~perm dir path =
   try eio_mkdirat parent leaf perm
   with Unix.Unix_error (code, name, arg) -> raise @@ Err.wrap_fs code name arg
 
+let float_of_time (s, ns) = Int64.to_float s +. (float ns /. 1e9)
```
talex5, line 378: OCaml itself makes sure the integer part isn't changed by rounding (https://github.com/ocaml/ocaml/issues/9490). We should probably do that too (here and in eio_posix).

lib_eio_posix/low_level.mli
```diff
@@ -36,7 +36,8 @@ val send_msg : fd -> ?fds:fd list -> ?dst:Unix.sockaddr -> Cstruct.t array -> in
 val getrandom : Cstruct.t -> unit
 
 val fstat : fd -> Unix.LargeFile.stats
-val lstat : string -> Unix.LargeFile.stats
+val lstat : string -> Unix.LargeFile.stats                      (* TODO: remove? *)
```
talex5, line 39: Should be able to remove this now.

talex5 (commented):
uring 0.7 is out now. I pushed an update for that to this PR.

avsm:
I'm just playing around with Patrick's help on a revised version of this interface that can minimise allocations for stats at the Eio level as well.

```
module Stat = struct
  type kind = [ `Unknown | `Regular_file | `Socket ]
  type 'a typ =
  | Ino : int64 typ
  | Kind : kind typ
  and ('a, 'ty) t =
    | [] : ('ty, 'ty) t
    | (::) : 'a typ * ('b, 'ty) t -> ('a -> 'b, 'ty) t

  let ino = Ino
  let kind = Kind

  let _x = [ino; kind; ino]

  type r = [`Ino | `Kind]

  let extract : type a. a typ -> a = function
    | Ino -> 1L
    | Kind -> `Unknown

  let rec stat : type a b. (a, b) t -> a -> b =
    fun v acc ->
      match v with
      | v :: tl -> 
        let res = extract v in
        let f = acc res in
        stat tl f
      | [] -> acc
end

type just_ino = { ino : int64 }
let pp_ino ppf t = Format.fprintf ppf "%Ld" t.ino

let () =
  let ino = Stat.(stat [ ino; kind ] (fun ino _kind -> { ino })) in
  Format.printf "%a" pp_ino ino
```

With this, the stat call takes in a continuation which receives just the specified fields, and the Linux backend can then translate that into a statx while the posix/other backends just do a full stat.  I'm just plumbing it all through to see what it looks like.

SGrondin:
@avsm I've just read the suggested patch. While I am personally familiar with both GADTs and defining custom list types, because both of those things "leak" into the `mli` I think the suggested `stat` API is slightly too obscure for the more novice OCaml developers. A more experienced developer would see the API and realize it's that way to avoid allocations.

Yes, with a good example or two it would be fine. But it would mean forcing most users to refer to the docs anytime they want to use `stat`. In comparison, the classic `stat` API is fully discoverable through `ocaml-lsp-server`/`merlin`.

Since the stats call is used in a wide range of use cases, what about offering `stat` (returns the classic `Stats.t` record), `exists` (returns a boolean), and `stats_apply` (advanced, GADT-based, allocation-conscious function)?

avsm:
I agree with the concerns about the obscurity of GADTs, but really do think that the default interface shouldn't be quite as inefficient as a complete retrieval of all the statistics, most of which are promptly thrown away.

Browsing through [sherlocode for Unix.stat](https://sherlocode.com/?q=Unix.stat) shows that a majority of the uses are of the form "(Unix.stat).field" to retrieve just one value.  We could just provide single-field accessors (for which `statx` is pretty good) on the same lines as `size` right now, but for all the fields.  Then the GADT interface would only be necessary when the developer wishes to retrieve multiple fields in one call.

SGrondin:
I think that's a great compromise.

patricoferris (commented):
With statx support and the others, do you think it would also be good to provide a `Path.exists` that uses the statx whilst we're here?

lib_eio/path.mli
```diff
@@ -128,6 +128,14 @@ val read_dir : _ t -> string list
 
     Note: The special Unix entries "." and ".." are not included in the results. *)
 
+(** {2 Metadata} *)
+
+val stat : follow:bool -> _ t -> ('a, 'b) File.Stat.t -> 'a -> 'b
+(** [stat ~follow t] returns metadata about the file [t].
+
+    If [t] is a symlink, the information returned is about the target if [follow = true],
+    otherwise it is about the link itself. *)
```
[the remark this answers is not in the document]
avsm: Added in 28169b593d00cfcbaf9093b35c84dd7c54f707c1

SGrondin:
Sorry for the slight bikeshedding, but do you think it would be an improvement to move the accessor functions into a `Stats` submodule? I think that would make it easier to "browse" those accessor functions to find the one you're looking for. I know I personally don't know all the stats fields by name. It would also declutter the `Path` module, making it easier for newcomers to understand its role versus Fs, File, Flow, Resource, etc.

> do you think it would also be good to provide a Path.exists that uses the statx whilst we're here?

I'd like to add my support for this suggestion. Otherwise it's something that a lot of users will cobble toghether, sometimes poorly by using `read_dir` or by catching an overly large range of exceptions in an attempt to detect a Not Found exception.

avsm:
I'm playing around with rearranging the accessor functions in the module so that the advanced ones are at the bottom of the interface rather than the top. That makes the simpler ones much more discoverable...

In the meanwhile though, the blocker to adding `Eio.Path.exists` is that `Unix.Unix_error(Unix.ENOENT,_,_)` cannot be caught in the Eio.Path module since Unix isn't available!

talex5:
`Unix.ENOENT` should never appear outside of `Eio_posix.Low_level`. I'm using https://github.com/talex5/eio/blob/bdac846f34f2ade585886270772a4e3d065ef498/lib_eio/path.ml#L163-L165:

```ocaml
let exists ~follow path =
  try ignore (stat ~follow path : File.Stat.t); true
  with Exn.Io (Fs.E Not_found _, _) -> false
```

avsm:
Take 3 pushed; I've:

- removed Eio.File.Stat entirely and just moved the type definitions into Eio.File. There's no conflict with the other definitions. It makes everything much more discoverable (to address @SGrondin's comment) since the user doesn't need to know of the existence of a `Stat` submodule.
- moved the complicated functions to the bottom of the mli file, so that the simpler accessor functions show up first. A simple change, but it makes a topdown reading of the odoc output much more beginner friendly.
- Added the Path.exists* functions. 

Still need:
- [ ] Windows
- [ ] test case updates to test the stat functions respect being confined (and do an EPERM exception there)
- [ ] add Eio.Pool to the statx buffer handling in Eio_linux

Another round of bikeshedding on this version welcome! :-)

patricoferris:
I haven't pushed it here just yet, but I have a WIP `stat` benchmark https://github.com/ocaml-multicore/eio/commit/b6c304600b24fb41a6765308feae6e5fa8b0357a (there's a rogue `Sys.command "rm -r"` in there so you might want to double-check it before you run!)

avsm:
Somewhat unexpectedly, hacking in an Eio.Pool to reuse statx buffers in the invocation of Eio_linux.Low_level.stat seems to consistently slow down @patricoferris' benchmark by 10% or so, and allocate more memory.

```
diff --git a/lib_eio_linux/low_level.ml b/lib_eio_linux/low_level.ml
index 07c1041..05f7138 100644
--- a/lib_eio_linux/low_level.ml
+++ b/lib_eio_linux/low_level.ml
@@ -352,10 +352,17 @@ let statx ?buf ?fd ?(flags=Uring.Statx.Flags.empty) path f k =
 in
 fn f k
 
+let rr = ref 0
+let stat_pool = Eio.Pool.create 1000
+  (fun () ->
+     incr rr;
+     Printf.eprintf "%d\n%!" !rr;
+     Uring.Statx.create ())
 let stat dir_fd ?flags path f k =
-  match dir_fd with
-  | FD fd -> statx ~fd ?flags path f k
-  | Cwd | Fs -> statx ?flags path f k
+    Eio.Pool.use stat_pool (fun buf ->
+    match dir_fd with
+    | FD fd -> statx ~buf ~fd ?flags path f k
+    | Cwd | Fs -> statx ~buf ?flags path f k)
 ```

(This version just printfs to make sure that the pool isn't freeing/reallocation, which it isn't. The right number of statx buffers are indeed allocated). Not had a chance to investigate further.

patricoferris:
Just a quick update to say I've consistently had issues with uring and big statting programs on a 5.15.0 kernel (it manifests as a `Not_found` exception but the file/directory definitely does exist). Don't see the issues on 6.2.0.

talex5:
I added a benchmark for fstat (#616) to see the effect of using the GADT vs a record. It seems to make no significant difference to performance, even when I remove the actual call to `fstat` and just use a fixed result in all cases. This is in the best case, accessing only one field. So this doesn't help performance on the OCaml side, but it does give the OS opportunities to optimise without penalising the OCaml side, which is good.

However, using uring for fstat is about 27x slower on my machine (Linux 6.1.52). Most of the time is spent in the kernel, and I think uring is pushing it to a worker thread. So we should probably leave that using the `fstat` system call for now.

patricoferris:
Interesting, sorry for dropping the ball a little on this PR. I'm not sure if it is related at all, but I noticed considerably slower code in https://github.com/patricoferris/aperitif when comparing Eio and the plain OCaml sequential code (confusingly under `miou/main_s`). I didn't really have to dig deeper and there's no point yet comparing with miou as it uses parallelism and the Eio version doesn't.

talex5:
I've spit out the low-level support into #617 and #618 so that can be reviewed separately.

avsm:
I'm ok with unblocking this PR with the current interface, but there are some important considerations we should keep in mind for the next rev that we've learnt from this.

* **uring support envelope:** This is highly dependent on the kernel version, and so we should establish some principles for the lower and upper bounds of when we activate uring for users. Some _functionality_ seems plain broken on older kernels (such as @patricoferris' report of 5.15 stat calls going wrong), and these should form the lower bound.  Our bar for _performance_ should probably be by testing on the very latest kernels available though, since uring does demand the bleeding edge for now. For example, I looked into @talex5's report of why the fstat syscall might be an order of magnitude faster than uring, and  discovered this hot-off-the-press [thread on why glibc's fstat is slow](https://lwn.net/SubscriberLink/944214/b673eccd39081c41/).  I think the answer actually lies in the fact that we're using the sandboxed syscalls by default, and not in worker threads. This will take time to figure out, so I'll push a WIP benchmark I have to the uring repository and we can test it out on multiple kernels and see exactly what's going on under the hood. 
* **memory usage:** It does still irk me that we're copying into a minor heap record so unnecessarily. I don't buy the argument that OCaml will eventually lay out records better in the future -- that sort of layout transformation is a considerable distance away.  However, I agree that we don't need to change the existing API until we have a clear understanding of the performance implications of the statx syscall before optimising anything else.
* **gadt interface:** I still like it -- it's got individual accessor functions and the 'advanced bit' isn't needed by the common case. However, again I am fine deferring this until we get a better handle on the performance. I would like any proposal for a better API to be accompanied by a clear perf/memory benchmark that illustrates the benefit, and there's too much to disentangle in this PR for that to happen right now!

talex5:
> Some functionality seems plain broken on older kernels (such as @patricoferris' report of 5.15 stat calls going wrong), and these should form the lower bound. 

Yes, we need to work out what the minimum should be. I'm on 6.1 and I haven't seen that error yet.

I'm working on a minimal patch to support stat with the current record system here: https://github.com/ocaml-multicore/eio/compare/main...talex5:eio:stat?expand=1 (will open a PR soon).

Would be good to get some benchmarks indeed. I did some quick tests recently and it looked like using the GADT with lots of fields was slower than using the record, while using it with a single field was no faster. Not sure why, though.

talex5:
I squashed and rebased this on top of #620, so it contains just the bits missing from that:
- Using a GADT to select which fields to use.
- `bench_stat` benchmark.
- Documentation for fields (I'll make a separate PR adding these to the record fields).
- `exists`, `is_file`, `is_directory` (need a separate PR for these).
- Using `statx` to replace `fstat` (too slow for now).
- `eio_stat` test (should probably be an example rather than a test).

talex5:
> Just a quick update to say I've consistently had issues with uring and big statting programs on a 5.15.0 kernel (it manifests as a Not_found exception but the file/directory definitely does exist). Don't see the issues on 6.2.0.

Probably related to this change in 5.18: [io-uring: Make statx API stable](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=1b6fe6e0dfecf8c82a64fb87148ad9333fa2f62e) (from @SGrondin's list at https://gist.github.com/SGrondin/be12fae9e4851fb864bc7d5de9297f1a).

talex5:
Closing for now, as most of this PR got merged and it's unclear it we want the other bits at all.

PR #602, opened 2023-08-15
Add Eio.Pool

talex5:
This is similar to `Lwt_pool`.

It provides a similar API to @darrenldl's draft in #566, but:
- Internally it uses a single `Cells` queue, which should be more efficient.
- It uses @patricoferris's suggestion of removing the switch.

I removed the clear function for now. According to https://sherlocode.com/?q=Lwt_pool.clear, the only user of `Lwt_pool.clear` is Lwt's own tests!

The docs note some complications regarding switches. We should probably try to make this simpler in future. e.g.

- It would be good to run `alloc` with a temporary switch: if it fails then everything is released, but if it succeeds then the resource gets adopted by the pool switch.
- Currently, switches can't be shared between domains, which makes using a pool across domains more difficult than it should be.

Instead of providing `validate`, `dispose` and `alloc` functions, we could just expose the slot and let the user deal with it. That would give maximum flexibility.

### DIFF
--- a/lib_eio/eio.ml
+++ b/lib_eio/eio.ml
@@ -9,6 +9,7 @@ module Mutex = Eio_mutex
 module Condition = Condition
 module Stream = Stream
 module Lazy = Lazy
+module Pool = Pool
 module Exn = Exn
 module Resource = Resource
 module Flow = Flow
--- a/lib_eio/eio.mli
+++ b/lib_eio/eio.mli
@@ -39,6 +39,9 @@ module Stream = Stream
 (** Delayed evaluation. *)
 module Lazy = Lazy
 
+(** A pool of resources. *)
+module Pool = Pool
+
 (** Cancelling fibers. *)
 module Cancel = Eio__core.Cancel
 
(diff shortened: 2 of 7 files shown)


avsm:
I am curious to see what the version that exposes the slot looks like, to avoid all the callbacks for the lifetime management...

talex5:
> I am curious to see what the version that exposes the slot looks like, to avoid all the callbacks for the lifetime management...

The simplest case would be just:
```ocaml
type 'a t
val create : int -> 'a t
val use : 'a t -> ('a ref -> 'b) -> 'b
```
When using the pool you get a slot (initially empty) and it's up to you to create a new resource and store it in the slot. That would mostly force people to create their own wrappers around pool.

Alternatively, the wrapper could be passed to `create`:
```ocaml
type 'a t

module type RESOURCE = sig
  type t
  val use_slot : t ref -> (t -> 'b) -> 'b
end

val create : (module RESOURCE) -> int -> 'a t

val use : 'a t -> ('a -> 'b) -> 'b
```

PR #603, opened 2023-08-16
Add Eio.Path.native

### LINKED ISSUE #513: Absolute path resolver?
I don't think it would be that uncommon for people to want to pass an `Eio.Fs.dir Eio.Path.t` into an `argv` for a process spawn. This happens  quite a lot in OBuilder (e.g. https://github.com/ocurrent/obuilder/blob/master/lib/btrfs_store.ml). At the moment I'm doing the `snd` trick which I know is wrong (I should probably follow the logic in the `resolve` method), it would be nice to have:

```ocaml
val resolve_absolute : Eio.Fs.dir Eio.Path.t -> string
```

talex5:
This is useful when interacting with non-Eio libraries, for spawning sub-processes, and for displaying paths to users.

Fixes #513.

### DIFF
--- a/lib_eio_posix/fs.ml
+++ b/lib_eio_posix/fs.ml
@@ -161,6 +161,21 @@ end = struct
     Eio.Resource.T (d, Handler.v)
 
   let pp f t = Fmt.string f (String.escaped t.label)
+
+  let native_internal t path =
+    if Filename.is_relative path then (
+      let p =
+        if t.dir_path = "." then path
+        else Filename.concat t.dir_path path
+      in
+      if p = "" then "."
+      else if p = "." then p
+      else if Filename.is_implicit p then "./" ^ p
+      else p
+    ) else path
+
+  let native t path =
+    Some (native_internal t path)
 end
 and Handler : sig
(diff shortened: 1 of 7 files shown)


lib_eio_posix/fs.ml
```diff
@@ -161,6 +161,21 @@ end = struct
     Eio.Resource.T (d, Handler.v)
 
   let pp f t = Fmt.string f (String.escaped t.label)
+
+  let native_int t path =
```
patricoferris, line 165: What's the reasoning for calling this `native_int` is it should for `internal` ? I couldn't help but think of `Stdlib.Nativeint` ^^"

patricoferris (approved):
Thanks, this looks good to me :)) I think it's the most reasonable approach and the docs give you all the caveats about why it might go quite wrong. I note other libraries e.g. Rust's std-cap go for a "compare metadata to rediscover the file from the parent approach".

lib_eio_posix/fs.ml
talex5: Good point - I've renamed it.

PR #608, opened 2023-08-17
Convert clocks to new resource system

talex5:
This is a continuation of #553.


lib_eio/unix/eio_unix.ml
```diff
@@ -40,8 +40,8 @@ module Stdenv = struct
     net : [`Unix | `Generic] Eio.Net.ty r;
     domain_mgr : Eio.Domain_manager.t;
     process_mgr : Process.mgr;
-    clock : Eio.Time.clock;
-    mono_clock : Eio.Time.Mono.t;
+    clock : float Eio.Time.clock_ty r;
```
avsm, line 43: why not also have a `Eio.Time.clock` that maps to `float Eio.Time.clock_ty`, as is done below for the mono clock?

avsm (approved):
Looks good, just a minor comment.

lib_eio/unix/eio_unix.ml
talex5: Mainly because I couldn't think of a good name for it. I tried `float_clock_ty`, but that's no shorter than just writing out `float clock_ty` anyway. We could use `clock_ty` just for floats, but then we need another name for the generic version.

avsm: Agreed, can't think of anything better. Leaving it like this for now seems fine.

PR #609, opened 2023-08-21
Add Eio.Lazy

talex5:
If one fiber tries to force a lazy value while another is already doing it, this will wait for the first one to finish rather than raising an exception (as `Stdlib.Lazy` does).

### DIFF
--- a/lib_eio/lazy.ml
+++ b/lib_eio/lazy.ml
@@ -1,0 +1,47 @@
+open Std
+
+type 'a state =
+  | Value of 'a
+  | Waiting of (unit Promise.u -> unit)
+  | Running of unit Promise.t   (* Wait until resolved and check again *)
+  | Failed of Exn.with_bt
+
+type 'a t = 'a state Atomic.t
+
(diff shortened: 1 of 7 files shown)


lib_eio/lazy.ml
```diff
@@ -0,0 +1,47 @@
...
+  | Waiting of (unit Promise.u -> unit)
+  | Running of unit Promise.t   (* Wait until resolved and check again *)
+  | Failed of Exn.with_bt
+
+type 'a t = 'a state Atomic.t
+
+let init = Waiting (fun _ -> assert false)
+
+let from_fun ~cancel fn =
+  let state = Atomic.make init in
```
SGrondin, line 14: Really minor suggestion: it might be slightly better to either pass `state` as an argument to `force` or to make `state` and `force` mutually recursive, as it saves a call to `Atomic.set` and (imo) makes the logic more explicit.

talex5: OCaml won't allow it:
```This kind of expression is not allowed as right-hand side of `let rec'```

PR #611, opened 2023-08-29
Add Eio.Flow.Pi.simple_copy

talex5:
This provides an easy way to implement the copy API when there are no optimisations you want to try.

### DIFF
--- a/lib_eio/flow.mli
+++ b/lib_eio/flow.mli
@@ -121,8 +121,13 @@ module Pi : sig
 
   module type SINK = sig
     type t
-    val copy : t -> src:_ source -> unit
+
     val single_write : t -> Cstruct.t list -> int
+
+    val copy : t -> src:_ source -> unit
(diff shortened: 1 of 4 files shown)


lib_eio/flow.mli
```diff
@@ -146,5 +151,8 @@ module Pi : sig
     | Source : ('t, (module SOURCE with type t = 't), [> source_ty]) Resource.pi
     | Sink : ('t, (module SINK with type t = 't), [> sink_ty]) Resource.pi
     | Shutdown : ('t, (module SHUTDOWN with type t = 't), [> shutdown_ty]) Resource.pi
+
+  val simple_copy : single_write:('t -> Cstruct.t list -> int) -> 't -> src:_ source -> unit
```
avsm, line 155: Would a better name for this function be "serial_copy"?  "simple" is subjective, but "serial" describes what it does.

PR #627, opened 2023-09-29
Add Path.rmtree

talex5:
We might want to add a `~force:true` argument at some point, but this will do to start with.

Needed for #601 and helpful for creating temporary directories (#510).

/cc @SGrondin


SGrondin (approved):
Looks great! And yes I'll be using it right away in #601

PR #629, opened 2023-10-07
eio_linux: mark as only available on Linux

talex5:
This avoids the CI trying to test it on e.g. FreeBSD.


talex5:
However, it still fails to find a solution on FreeBSD:
```
2023-10-07 13:33.47 [INFO] = freebsd-5.1_opam-2.1 =
      ocaml-base-compiler.5.1.0: Rejected by user-specified constraint ocaml-base-compiler (= 5.1)
```

/cc @tmcgilchrist

PR #631, opened 2023-10-09
Update minimum OCaml version to 5.1

talex5:
This is to allow using the new events system in future.


talex5:
The benchmark service uses OCaml 5.0 by default, so I had to add a `bench.Dockerfile` to fix that. For some reason, it doesn't show a comparison between the old and new benchmarks, but looking at them by eye it mostly got a bit faster.

PR #639, opened 2023-11-06
Eio.Executor_pool

SGrondin:
This is a continuation of #584. After discussing the design, we've agreed to first merge a simple and minimalistic version. I will be adding barriers, job weights, etc., in an upcoming version 2.

### DIFF
--- a/lib_eio/executor_pool.ml
+++ b/lib_eio/executor_pool.ml
@@ -1,0 +1,73 @@
+type job = Pack : {
+    fn : unit -> 'a;
+    w : ('a, exn) Result.t Promise.u;
+    weight : int;
+  } -> job
+
+type t = {
+  queue : job Sync.t;
+}
+
+let max_capacity = 1_000_000
+
+let max_capacity_f = float max_capacity
+
+(* This function is the core of executor_pool.ml.
+   Each worker runs in its own domain,
+   taking jobs from [queue] whenever it has spare capacity. *)
+let run_worker { queue } =
+  Switch.run @@ fun sw ->
+  let capacity = ref 0 in
+  let condition = Condition.create () in
+  (* The main worker loop. *)
+  let rec loop () =
+    while !capacity >= max_capacity do Condition.await_no_mutex condition done;
+    match Sync.take queue with
+    | Error `Closed -> `Stop_daemon
+    | Ok (Pack { fn; w; weight }) ->
+      capacity := !capacity + weight;
+      Fiber.fork ~sw (fun () ->
+          Promise.resolve w (try Ok (fn ()) with ex -> Error ex);
+          capacity := !capacity - weight;
+          Condition.broadcast condition
+        );
+      (* Give a chance to other domains to start waiting on [queue]
+         before the current thread blocks on [Sync.take] again. *)
+      Fiber.yield ();
+      (loop [@tailcall]) ()
+  in
+  loop ()
+
+let create ~sw ~domain_count domain_mgr =
+  let queue = Sync.create () in
+  let t = { queue } in
+  Switch.on_release sw (fun () -> Sync.close queue);
+  for _ = 1 to domain_count do
+    (* Workers run as daemons to not hold the user's switch from completing.
+       It's up to the user to hold the switch open (and thus, the executor pool)
+       by blocking on the jobs issued to the pool. *)
+    Fiber.fork_daemon ~sw (fun () ->
+        Domain_manager.run domain_mgr (fun () ->
+            run_worker t))
+  done;
+  t
+
+let enqueue { queue } ~weight fn =
+  if not (weight >= 0. && weight <= 1.) (* Handles NaN *)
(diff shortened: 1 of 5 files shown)


SGrondin:
The only open question remaining is whether is check the Switch's status when submitting jobs. I'm thinking it could be worth it to avoid silent deadlocks when the user does something funky with their lifetimes.

lib_eio/executor_pool.ml
```diff
@@ -0,0 +1,75 @@
...
+let create ~sw ~domain_count ~domain_concurrency domain_mgr =
+  let stream = Stream.create 0 in
+  let instance = { sw; stream; } in
+  for _ = 1 to domain_count do
+    start_domain ~sw ~domain_mgr ~limit:domain_concurrency stream
+  done;
+  instance
+
+let enqueue { sw; stream } f =
+  Switch.check sw;
```
SGrondin, line 58: @talex5 I believe `Switch.check` is not thread safe. Is there a safe way to accomplish this or should I just let the users who do strange things with their lifetimes risk deadlock?

lib_eio/executor_pool.ml
```diff
@@ -0,0 +1,76 @@
...
+  let stream = Stream.create 0 in
+  let instance = { stream; cancelled = Atomic.make false } in
+  Switch.on_release sw (fun () -> Atomic.set instance.cancelled true);
+  for _ = 1 to domain_count do
+    start_domain ~sw ~domain_mgr ~limit:domain_concurrency stream
+  done;
+  instance
+
+let enqueue { stream; cancelled } f =
+  if Atomic.get cancelled then raise (Failure "Executor_pool: Switch cancelled");
```
SGrondin, line 59: I know this is somewhat overkill, but without it users doing truly horrifying things with their timelines can end up in a (well deserved) deadlocked. 😆

talex5:
I experimented with some changes here: https://github.com/SGrondin/eio/compare/executorpool...talex5:eio:executorpool?expand=1

Needs a bit more thinking about, but this should ensure that jobs are always rejected if the pool shuts down, though it does add some cost to the fast-path.

SGrondin:
I like this a lot. Compared to the previous version it guarantees that even egregious misuse never leads to a deadlock at the cost of complexity and a small overhead. The previous version did a "best effort" to avoid deadlocks in such cases. I'll merge it into my branch and play with it a bit.

SGrondin:
I added more information to the mli, fixed a typo, and removed a few `Fiber.yield ()` from the tests (fewer are needed now that we don't use `Fiber.n_any`). The commit is [here](https://github.com/SGrondin/eio/commit/03bfde4fb5a589f3b9fb61c762b55d1daba0779f). I then recombined all those commits into one. I'm looking at `Stream.close` now.

SGrondin:
@talex5 I've made a small change in the API that I believe makes the Executor_pool more powerful without significantly increasing the complexity (both of the code and to the user). It's getting late in the design cycle to make this change, but I think you'll agree with it.

talex5:
What about fixing the domain capacity at 1 and having users say what fraction of the CPU they expect to use? The capacity seems arbitrary and might be hard to pick if you're sharing one pool with several parts of your program.

SGrondin:
Yes, I went back and forth on that one. I'll switch it over 👍

talex5 (approved):
I pushed a couple of minor changes (cope with `weight:nan` and use `invalid_arg`). Looks good to merge now!

lib_eio/executor_pool.mli
```diff
@@ -0,0 +1,58 @@
...
+  domain_count:int ->
+  _ Domain_manager.t ->
+  t
+(** [create ~sw ~domain_count dm] creates a new executor pool.
+
+    The executor pool will not block switch [sw] from completing;
+    when the switch finishes, all domain workers and running jobs are cancelled.
+
+    @param domain_count The number of domain workers to create.
+                        This number should not exceed the number of CPU cores on your system.
```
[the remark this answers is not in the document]
SGrondin: Done, I added a note.

talex5:
Thanks!

(I added a final minor fix to ensure the client's promise is resolved if the worker is cancelled just as it accepts a job.)

lib_eio/executor_pool.ml
```diff
@@ -0,0 +1,74 @@
...
+       by blocking on the jobs issued to the pool. *)
+    Fiber.fork_daemon ~sw (fun () ->
+        Domain_manager.run domain_mgr (fun () ->
+            run_worker t))
+  done;
+  t
+
+let enqueue { queue } ~weight fn =
+  if not (weight >= 0. && weight <= 1.) (* Handles NaN *)
+  then Fmt.invalid_arg "Executor_pool: weight not >= 0.0 && <= 1.0" weight
```
yawaramin, line 58: Should this be:

```ocaml
Fmt.invalid_arg "Executor_pool: weight not >= 0.0 && <= 1.0: %f" weight
```

?

PR #641, opened 2023-11-10
Allow closing synchronous streams

talex5:
This isn't currently exposed in the public interface, but that could be done easily enough.

Need to decide how to indicate end-of-stream. This version has `take` return `None`, but that also means that `take_nonblocking` returns `None` for both no-item and closed-stream.

/cc @SGrondin This might be useful for #639.

### DIFF
--- a/lib_eio/stream.ml
+++ b/lib_eio/stream.ml
@@ -120,12 +120,15 @@ let add t v =
   | Locking x -> Locking.add x v
 
 let take = function
-  | Sync x -> Sync.take x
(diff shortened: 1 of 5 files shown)


SGrondin:
This is great! It's something I need to port a major codebase to Eio and I can't be the only one. Good decision on ``[ `Would_block | `Closed]``.

Another useful and related feature is something similar to `Lwt_stream.closed` (`'a t -> unit Lwt.t`), it returns a promise that resolves once the stream is closed.

talex5:
Not sure about adding a closed promise. It's easy enough for users to do that themselves if they need it - it doesn't need any support from Eio.

PR #644, opened 2023-11-15
eio.mock: auto-advancing mock clock

talex5:
The clock advances automatically when the run queue is empty, as discussed in https://github.com/ocaml-multicore/eio/pull/584#discussion_r1313985513.

Requested by @SGrondin.

PR #645, opened 2023-11-15
Eio_mock.Backend.yield_until_stable

SGrondin:
This is a complement to the new auto-advancing mock clocks of #644 

I didn't replace the double-yield in `network.md` because #644 takes care of it.

This PR is intended to eliminate the need to manually yield more than once in tests, in the absence of clocks.


SGrondin:
As pointed out by @talex5, any use of `yield_until_stable` can be replaced by a `sleep 0` as of #644

PR #647, opened 2023-11-17
Faster and simpler Lf_queue

talex5:
The old code was based on a multi-consumer design, but specialised to a single consumer. As @polytypic points out, if there is only a single consumer then a more efficient design is possible.

Previously, adding an item required an atomic read and two CAS operations (in the uncontended case). It now requires an atomic read and one CAS. Taking an item previously required an atomic read in all cases, but now often requires no atomic operations. `push_head` no longer requires any atomic operations.

This also removes all the uses of `Obj.magic` from the module.


talex5:
Shows a nice improvement on some micro-benchmarks, e.g.

Screenshot 2023-11-19 at 10-22-28 OCaml Benchmarks

Issue #642, opened 2023-11-11
api questions and some migration feedback


mbarbin:
Hi!

I've recently gone through the exercise of migrating some code that used `Printf` & `Stdio.{In,Out}_channel` & `Fpath` to `Eio`, and I had a mix of questions, feedback, and data point of things that gave me pause. This is all very minor, things worked great, thank you for Eio! 

### Reading with Buf_read

This is just pointing out what I found confusing, which I believe is just some minor naming inconsistency: sometimes the reader checks that the next thing is as given, and consumes it (such as `Buf_read.string`), sometimes it is actually only reading but not checking `Buf_read.line`. And sometime a `take_` prefix is added `Buf_read.take_all` but sometimes not `Buf_read.line`.

I wish there was a common prefix for things that check things (e.g. check_ ?) and some common naming for things that take things (perhaps take_ , or no prefix if the naming of things that check do get their own prefix (e.g. `Buf_read.string` would be free if the current one becomes `Buf_read.check_string`). 

### Writing to a sink

I felt like missing a function to refactor code that would make use of format string, such as Printf.printf or Format.printf equivalent. So far this looks like `Eio.Flow.copy_string (Printf.sprintf "..." ...) sink`. It would be great to have a more direct alternate, that probably would take the sink as first argument and format second.

### Filename part of `Eio.Path`

I've transitioned some code that made use of `Fpath`, and in particular sections of code that would need to do some matching on the complete filename, using `Fpath.segments`. In the new `Eio` code, I've been using (path |> snd |> String.split ~on:'/'), which mostly do what I need. But I felt it's brittle, since it is relying on information I have looked in the implementation (that the second value of the tuple is manipulated with the `Filename` module). I don't mean to use `native` paths here, because I would like to be able to match on the path with an interface equivalent to `Fpath` that would be common regardless of the os in which this will be executed. I don't have a concrete proposal here. If you are interested perhaps I could give this more thoughts. What about for example:

```oaml
val Eio.Path.filename : _ Eio.Path.t -> string
val Eio.Path.segments : _ Eio.Path.t -> string list
```

### Type that encapsulate some capabilities during `create`

I've a type that offers functionality that interacts with the path and may spawn processes from time to time. I've taken so far the approach to require in `create` the part of the `env` that will be used by the other function of the interface, rather than adding each capability at the granularity of each function that operates on `t` that might need it.

```diff
- type t
+ type 'a t
```

```diff
val create
  : ...
[...]

PR #648, opened 2023-11-17
Add Path.{basename,dirname}

mbarbin:
Add `Path.basename` and `Path.dirname` as convenience wrappers around `Path.split`.

I had originally thought it would be useful for these functions not to return an option, but later changed my mind. I updated the PR to match @talex5's recommendation in #642:

> I think basename etc should just be e.g. snd (split t).

As it turns out, I now only uses `split` in the piece of code that I was migrating to eio that prompted my original feedback, thus I am not confident whether these wrappers are useful. Please feel free to close the PR. Thank you!

### DIFF
--- a/lib_eio/path.ml
+++ b/lib_eio/path.ml
@@ -53,6 +53,10 @@ let split (dir, p) =
       in
       Some ((dir, dirname), basename)
 
+let basename t = Option.map snd (split t)
+
+let dirname t = Option.map (fun ((_, dir), _) -> dir) (split t)
+
 let open_in ~sw t =
   let (Resource.T (dir, ops), path) = t in
   let module X = (val (Resource.get ops Fs.Pi.Dir)) in
--- a/lib_eio/path.ml
+++ b/lib_eio/path.ml
@@ -55,7 +55,7 @@ let split (dir, p) =
 
 let basename t = Option.map snd (split t)
 
-let dirname t = Option.map (fun ((_, dir), _) -> dir) (split t)
+let dirname t = Option.map fst (split t)
 
 let open_in ~sw t =
   let (Resource.T (dir, ops), path) = t in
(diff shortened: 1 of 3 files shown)


mbarbin:
Hi @talex5 and team,

I want to express my gratitude for your time and the thoughtful feedback you've
provided on issue #642. It's been a valuable learning experience for me.

After further consideration and usage of the `split` function in my code, I've
come to realize that the `basename` and `dirname` convenience wrappers may not
be the best match for the code I was migrating to `Eio`. The `split` function
provides sufficient functionality for my current needs.

I am also having second thoughts on implementing `dirname` as a simple wrapper
using `snd`, as it can potentially return an empty string, which is an invalid
path. This can cause functions like `Fpath.of_string` to fail. It's worth noting
too that this behavior diverges from the Unix commands `dirname`, which might be
surprising:

```sh
$ dirname ' '
.
$ dirname '/'
/
$ dirname '.'
.
```

I believe this could potentially lead to confusion and unexpected bugs. As a
data point, I wasn't careful enough and the very first attempt I made to using
`dirname` resulted in a runtime error:
`Fpath.of_string` raising `Invalid_argument \"\": invalid path`.

I want to clarify that my feedback in #642 was based on my personal experience
and the specific use case I was working on. I'm still exploring the best way to
articulate what I felt was missing regarding path manipulation, and I hope to
provide more constructive feedback in the future, perhaps opening a more
specific issue, TBD.

Given these reasons, I've decided to close this PR. However, I don't want this
to discourage anyone else who might find wrappers of this kind useful from
creating another PR. I believe that different use cases might warrant different
solutions, and this might be beneficial to someone else with other motivating
examples.

Thank you for your time and for the great work you're doing on the Eio library.
I look forward to discussing more in the future.

talex5:
Thanks - that's really useful feedback. It does sound like we need to wait until we have a use for these functions before deciding how they should behave.

PR #650, opened 2023-11-26
Executor_pool docs

SGrondin:
Changes to the README:
- Replaced a few `"5.0"` with `"5.1"` since Eio now requires OCaml 5.1
- Moved the `Domainslib` and `kcas` sections to their own files in `doc/`
- Rewrote the `Multicore Support` section to include, in order:
  - an overview of the tools available
  - a comparison between `Executor_pool` and `Domain_manager.run`. Assumption: expert users will know when they need to directly create threads, no need to explain the rationale to them.
  - docs and example for `Executor_pool`
  - docs and example for `Domain_manager.run`
  - a basic "Multicore Performance Guide" based on our conversation with @samoht
- ~I've moved the following sections to be after `Synchronization tools`:~
  - ~`Multicore Support` (the user shoul learn about Promise, Mutexes, Streams, etc. first)~
  - ~`Running processes` (same, and I expect this feature to see less frequent use than most modules that were placed ahead of it in the docs)~
  - ~`Tracing` (this is a somewhat advanced feature, mostly geared towards debugging. I find it easier to not constantly mix "features" with "testing+debugging" in the docs)~
  - ~`Testing with Mocks` (same as Tracing)~
- ~Moved the `Mock Clocks` section from `Time` to `Testing with Mocks`~
- Removed the `Worker Pool` example from the `Streams` section
  - I kept as much information as possible. Feel free to re-add anything here.

Please let me know if any of the above causes issues. Feel free to make changes directly on this branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # Eio -- Effects-Based Parallel IO for OCaml
 
-Eio provides an effects-based direct-style IO stack for OCaml 5.0.
+Eio provides an effects-based direct-style IO stack for OCaml 5.
 For example, you can use Eio to read and write files, make network connections,
 or perform CPU-intensive calculations, running multiple operations at the same time.
 It aims to be easy to use, secure, well documented, and fast.
@@ -17,7 +17,7 @@ Eio replaces existing concurrency libraries such as Lwt
 * [Motivation](#motivation)
 * [Current Status](#current-status)
 * [Structure of the Code](#structure-of-the-code)
-* [Getting OCaml 5.0](#getting-ocaml-50)
+* [Getting OCaml 5.1](#getting-ocaml-51)
 * [Getting Eio](#getting-eio)
 * [Running Eio](#running-eio)
 * [Testing with Mocks](#testing-with-mocks)
@@ -36,23 +36,19 @@ Eio replaces existing concurrency libraries such as Lwt
 * [Running processes](#running-processes)
 * [Time](#time)
 * [Multicore Support](#multicore-support)
+    * [Executor Pool](#executor-pool)
+    * [Domain Manager](#domain-manager)
 * [Synchronisation Tools](#synchronisation-tools)
     * [Promises](#promises)
-    * [Example: Concurrent Cache](#example-concurrent-cache)
+      * [Example: Concurrent Cache](#example-concurrent-cache)
     * [Streams](#streams)
-    * [Example: Worker Pool](#example-worker-pool)
     * [Mutexes and Semaphores](#mutexes-and-semaphores)
     * [Conditions](#conditions)
-    * [Example: Signal handlers](#example-signal-handlers)
+      * [Example: Signal handlers](#example-signal-handlers)
 * [Design Note: Determinism](#design-note-determinism)
 * [Provider Interfaces](#provider-interfaces)
 * [Example Applications](#example-applications)
 * [Integrations](#integrations)
-    * [Async](#async)
-    * [Lwt](#lwt)
-    * [Unix and System Threads](#unix-and-system-threads)
-    * [Domainslib](#domainslib)
-    * [kcas](#kcas)
 * [Best Practices](#best-practices)
     * [Switches](#switches-1)
     * [Casting](#casting)
@@ -67,7 +63,7 @@ The `Unix` library provided with OCaml uses blocking IO operations, and is not w
 For many years, the solution to this has been libraries such as Lwt and Async, which provide a monadic interface.
 These libraries allow writing code as if there were multiple threads of execution, each with their own stack, but the stacks are simulated using the heap.
 
-OCaml 5.0 adds support for "effects", removing the need for monadic code here.
+OCaml 5 adds support for "effects", removing the need for monadic code here.
 Using effects brings several advantages:
 
 1. It's faster, because no heap allocations are needed to simulate a stack.
@@ -87,7 +83,7 @@ and we hope that Eio will be that API.
 
 See [Eio 1.0 progress tracking](https://github.com/ocaml-multicore/eio/issues/388) for the current status.
 Please try porting your programs to use Eio and submit PRs or open issues when you find problems.
-Remember that you can always fall back to using Lwt libraries to provide missing features if necessary.
+Remember that you can always [fall back to using Lwt libraries](#lwt) to provide missing features if necessary.
 
 See [Awesome Multicore OCaml][] for links to work migrating other projects to Eio.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 # Eio -- Effects-Based Parallel IO for OCaml
 
-Eio provides an effects-based direct-style IO stack for OCaml 5.0.
+Eio provides an effects-based direct-style IO stack for OCaml 5.1.
```
Sudha247, line 5: Should we make this OCaml 5.1+?

README.md
```diff
@@ -1559,28 +1256,357 @@ let main ~config_changed =
...
+  - The [Synchronisation Tools](#synchronisation-tools) section documents various threadsafe tools to communicate across domains, such as `Eio.Promise` and `Eio.Stream`.
+  - The [Saturn](https://github.com/ocaml-multicore/saturn) library also offers a collection of efficient threadsafe data structures.
 
-In particular, if you test your code by providing (deterministic) mocks then the tests will be deterministic.
-An easy way to write tests is by having the mocks call `traceln` and then comparing the trace output with the expected output.
-See Eio's own tests for examples, e.g., [tests/switch.md](tests/switch.md).
+### Executor Pool
 
-## Provider Interfaces
+An executor pool distributes jobs among a pool of domains workers (threads).
```
Sudha247, line 1283: I'm not sure we need to mention "thread" here. It can be counter-intuitive. I believe thread here is referring to a thread of execution. In general a thread in OCaml means systhread by default.

README.md
```diff
@@ -1559,28 +1256,357 @@ let main ~config_changed =
...
+- `Eio.Executor_pool`
+- spawning a one-off domain using `Eio.Domain_manager.run`
 
-Performing IO with external objects (e.g., `stdout`, files, or network sockets) will introduce non-determinism,
-as will using multiple domains.
+The Executor Pool is the recommended way of leveraging OCaml 5's multicore capabilities.
+It's built on top of `Eio.Domain_manager` and offers several advantages:
+- Executor Pool domains (threads) are initialized in advance to minimize overhead. They're also reused, unlike direct usage of `Eio.Domain_manager.run`.
+- Each Executor Pool domain worker can execute multiple jobs concurrently. Whenever a job blocks due to e.g. I/O, the domain's scheduler switches execution to another fiber and/or job.
+- The Executor Pool supports heterogenous jobs efficiently. For example, it will seamlessly run a large number of I/O-bound jobs on one domain while reserving an entire other domain (thread) for a single CPU-bound job. You control this behavior through the `~weight` argument required to submit a job.
```
Sudha247, line 1272: Maybe it's worth mentioning `weight` allows running computations in a domain until it reaches capacity. I understood this after looking at the mli file.

I'm also curious whether we can automate this somehow. There's a risk of users providing inaccurate weights. But that's not a conversation for this PR.

README.md
```diff
@@ -1559,28 +1256,357 @@ let main ~config_changed =
...
+Note: the exact `traceln` output of this example is non-deterministic, because the OS is free to schedule domains as it likes.
+
+`~weight` is the anticipated proportion of a CPU core used by the job.
+Here we use `~weight:1.0` because the job is entirely CPU-bound: it never waits for I/O or other syscalls.
+`~weight` must be `>= 0.0` and `<= 1.0`;
+Example: given an IO-bound job that averages 2% of one CPU core, pass `~weight:0.02`.
+
+Each domain worker starts new jobs until the total `~weight` of its running jobs reaches `1.0`.
+
+### Domain Manager
```
Sudha247, line 1364: Perhaps move the Domain Manager part on top and then introduce the Executor pool as an improvement?

README.md
```diff
@@ -1559,28 +1256,357 @@ let main ~config_changed =
...
+- If your program never hits 100% CPU usage, it's unlikely that parallelizing it will improve performance.
+- Try to avoid reading or writing to memory that's modified by other domains after the start of your job.
+  Ideally, your jobs shouldn't need to interact with other domains' "working data".
+  Aim to make your jobs as independent as possible.
+  - If unavoidable, the [Saturn](https://github.com/ocaml-multicore/saturn) library offers a collection of efficient threadsafe data structures.
+  - The [Multicore Guide](./doc/multicore.md) also explains how to safely and efficiently share memory between domains
+- It's often easier to design code to be multithreading friendly from the start
+  (by making longer, independent jobs) than by refactoring existing code.
+- There's a cost associated with creating a domain, so try to use the same domains for longer periods of time. The Executor Pool takes care of this automatically.
+- Having a large number of domains active at the same time imposes additional overhead on both the OS scheduler and the OCaml runtime, even if those domains are idle.
```
Sudha247, line 1416: I think some of the issues with idle domains and performance were fixed in 5.1 https://github.com/ocaml/ocaml/pull/11903. It's ok to leave the note as is. If there are more issues, it might be worth reporting upstream.

Sudha247 (commented):
Thanks @SGrondin! This is great. I've left some comments below.

README.md
balat: Or just OCaml 5

README.md
SGrondin: I thought a lot about this yet I'm still unsure about it, so I appreciate your perspective.

The reasoning was that _I think_ that starting within the next 6 months, Eio will become the single most common entrypoint for OCaml 4 developers to learn about OCaml 5's multicore feature. Within the same timeframe, I also believe that Eio will be **the** place where C#, Java, Go, etc., developers learn how to perform I/O and use multithreading in OCaml. Those developers will be familiar with the concept of threads and I wanted to emphasize that whenever they see the word "domain", they can rest assured that it maps 1:1 to a thread.

Thoughts? What do you think?

README.md
SGrondin: I'd prefer to keep the docs on the Executor Pool self-sufficient so users don't have to learn about a different module to understand it. I've made small tweaks to help with the confusion you've encountered.

README.md
SGrondin: Good point, I've made a few changes to help.

I think it's okay if the average user ends up looking at the mli through their code editor or on the https://ocaml-multicore.github.io/eio/eio/Eio/ site once they start using a module if it helps keep the README more focused on just: the purpose of a module, how it fits within the library, and advice about using it effectively.

> I'm also curious whether we can automate this somehow. There's a risk of users providing inaccurate weights. But that's not a conversation for this PR.

I've been thinking the same thing but I haven't been able to figure out a decent abstraction that brings enough value to justify losing the simplicity (in the API) of a bare float. But yeah, that's a separate discussion for another time and place. Please let me know if you come up with any ideas or even fragments of ideas!

README.md
SGrondin: Ah, good to know!

SGrondin:
Thank you @Sudha247 and @balat for the comments! I've pushed a new commit with a few small changes based on your feedback.

README.md
Sudha247: Fair enough. Maybe add a one-line explanation of what a domain is (and how it relates to the notion of thread in other languages)? We can also link to the manual page on parallelism https://v2.ocaml.org/manual/parallelism.html. I'm still not sure about correlating thread and domain because one domain can host many systhreads.

SGrondin: Good idea, done.

Sudha247 (approved):
LGTM. I'll wait until @talex5 has had a chance to review before merging.

talex5:
It's hard to see what changes with sections being moved and edited at the same time. It seems like this is combining three things:
1. Some obvious fixes, e.g. requiring OCaml 5.1
2. Moving things around
3. Documenting the executor pool

It would be helpful to separate these, as 2 needs more thinking about. For example, the section introducing `traceln` is moved later, but all of the examples use `traceln`. The part about mocks could perhaps go a bit later, but the reason for having it at the front is to show an immediate benefit of passing the environment to `main`, as people might wonder why we're doing that.

The tracing section comes right after the fibers section so you can see visually what it looks like to run fibers concurrently. Ideally, we'd also illustrate the other examples, and readers can run the tracing tool to understand any code they write. As the tools aren't ready yet, it currently just shows how to dump the logs, which I agree isn't useful here, but this is just a place-holder while that's being written.

talex5:
Most of this was merged in #650, so closing.

PR #655, opened 2023-12-04
Add Buf_write.printf (custom formatter)

SGrondin:
After a discussion with @talex5 I rewrote #652 using a custom formatter and a mutable boolean instead of a symbolic formatter. It's shorter and more performant but it adds just a liiiitle bit more mutable state. I'm reminded of [this famous xkcd](https://xkcd.com/292/).

### DIFF
--- a/lib_eio/buf_write.mli
+++ b/lib_eio/buf_write.mli
@@ -46,7 +46,7 @@
     or application-specific output APIs.
 
     A Buf_write serializer manages an internal buffer and a queue of output
-    buffers. The output bufferes may be a sub range of the serializer's
+    buffers. The output buffers may be a sub range of the serializer's
     internal buffer or one that is user-provided. Buffered writes such as
     {!string}, {!char}, {!cstruct}, etc., copy the source bytes into the
     serializer's internal buffer. Unbuffered writes are done with
@@ -122,6 +122,18 @@ val cstruct : t -> Cstruct.t -> unit
 (** [cstruct t cs] copies [cs] into the serializer's internal buffer.
     It is safe to modify [cs] after this call returns.
     For large cstructs, it may be more efficient to use {!schedule_cstruct}. *)
+
+val printf : t -> ('a, Format.formatter, unit) format -> 'a
+(** [printf t fmt ...] formats the arguments according to the format string [fmt].
+
+    It supports all formatting and pretty-printing features of the Format module.
+    The formatter's internal buffer is flushed to [t] after the call, without flushing [t] itself.
+    Explicit flushes (e.g. using [@.] or [%!]) perform a full (blocking) flush of [t]. *)
+
+val make_formatter : t -> Format.formatter
+(** [make_formatter t] creates a new formatter that writes to [t].
+
+    Flushing the formatter also flushes [t] itself. *)
 
 val write_gen
   :  t
--- a/lib_eio/buf_write.ml
+++ b/lib_eio/buf_write.ml
@@ -163,6 +163,7 @@ type t =
   ; mutable bytes_written  : int        (* Total written bytes. Wraps. *)
   ; mutable state          : state
   ; mutable wake_writer    : unit -> unit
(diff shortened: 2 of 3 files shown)


talex5:
Thanks - that's a lot simpler!

I'm not sure the formatter needs to be part of the writer at all. I pushed a commit which makes it entirely separate, so that each call to `printf` gets a formatter in the default state. I'm not sure whether that's an API improvement or not - it's not clear whether people will want to persist configuration changes between prints - but it's nice not having extra fields in `Buf_write.t`.

I wonder if we should only provide `printf`. The user can always do `printf "%t" f` to get access to the formatter if they want it, and it's easy enough for them to make their own custom formatter (with whatever flushing behaviour they prefer) for complex cases.

SGrondin:
I took a few days to think about this.

1. I think the change from `get_formatter` to `make_formatter` is a substantial improvement.
2. I'm torn about recreating the formatter on every use of `printf`. You've mentioned that OCaml format strings are quite slow... but recreating the fairly large formatter record on every call feels like punishing the user just to save an 8-byte pointer in the writer record. I think it would be surprising behavior, too. I understand the user can make their own formatter, but then they also need to flush it themselves, etc, and while none of that added complexity is _hard_, I don't think any of it is _necessary_ to put on the user. It's also potentially twice the number of values for the user to store or pass around.

Please let me know what you think of the commit I just pushed. I kept all of your changes, but I readded the formatter to the Buf_write record, this time only using a single field instead of 2, so 8 bytes instead of 16, plus it's initialized with a None so there's no extra allocations unless you use `printf`.

Thanks as always for your time and feedback, the resulting APIs we've developed together have so far been better than either one of our individual starting points.

lib_eio/buf_write.mli
```diff
@@ -123,6 +123,19 @@ val cstruct : t -> Cstruct.t -> unit
     It is safe to modify [cs] after this call returns.
     For large cstructs, it may be more efficient to use {!schedule_cstruct}. *)
 
+val printf : t -> ('a, Format.formatter, unit) format -> 'a
+(** [printf t fmt x y z] formats the arguments according to the format string [fmt].
+    It supports all formatting and pretty-printing features of the Format module.
+    Accordingly, explicit flushes using [@.] or [%!] must perform a full (blocking) flush
+    so consider using [Fiber.fork] in such cases. *)
```
talex5, line 130: Telling people to use Fiber.fork as a work-around seems odd, and might lead to errors if two fibers use the same formatter at once. Possibly the flushing behaviour should be controlled by an argument here.

Also, the word `must` seems out of place. We're not asking the user to do anything, just noting that we have chosen to do a full flush in this case.

talex5 (commented):
With your commit, it fails on the second use with:
```
Fatal error: exception Sys_error("Buf_write.printf: invalid concurrent access")
```
It just needs to reset `is_formatting` when using the cached version. Now that it's stateful, it would be good for the unit-test to test using it twice.

talex5:
Maybe we should call this `Buf_write.format`, in case we want to add a `printf` that does `Printf.sprintf` later.

lib_eio/buf_write.mli
SGrondin: Fixed. Agreed. I've got reservations about an argument for the flushing behavior, we can talk more about it at the next dev meeting.

SGrondin:
Fixed and rebased.

Yeah, that was an embarrassing oversight on my part. 😅 

> Maybe we should call this Buf_write.format, in case we want to add a printf that does Printf.sprintf later.

I don't think I understand what `Buf_write.sprintf` would do. If `sprintf` means writing into a string, I don't see how it would be different from `Format.sprintf` itself?

talex5 (approved):
Thanks!

> I don't think I understand what Buf_write.sprintf would do. If sprintf means writing into a string, I don't see how it would be different from Format.sprintf itself?

It's just a bit faster if you don't need the extra formatting. Probably doesn't matter.

PR #656, opened 2023-12-10
Improve tracing

talex5:
This adds tracing of cancellation contexts and OS operations, and
documents and cleans up the API a bit.

When reading events, instead of the consumer providing one function per
event, we now use a single function taking a variant. This makes the API
easier to use and lets the caller decide whether they want to handle all
events (checked by the compiler) or just some of them.

Note: eio_posix yields before each operation to avoid starvation.
However, if it needed to block then it did another yield afterwards when
retrying, which isn't necessary and clutters up the traces. This commit
also removes that.

### DIFF
--- a/examples/trace/main.ml
+++ b/examples/trace/main.ml
@@ -9,13 +9,6 @@
 
 open Eio.Std
 
-(* [handle pp] handles an event by writing the timestamp, ring ID and user data with [traceln].
-   [pp] is used to format the user data. *)
-let handle pp : _ Eio_runtime_events.handler =
-  fun ring ts v ->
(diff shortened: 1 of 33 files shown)


talex5:
Using a prototype visualiser, this is what `examples/net` looks like with these events:

trace

c-cube:
are the green zones representing times where a fiber is actually running?

c-cube:
That's fairly cool, I don't think I've seen that in any "classic" trace visualizer (perfetto, tracy, etc.). It means you emit a lot of events though, don't you?

talex5:
We only switch fibers when a fiber suspends, and we need to log why it's suspending anyway, so it's not a lot more events.

PR #657, opened 2023-12-16
Add Eio.Pool.use ~never_block

SGrondin:
This PR adds the optional parameter `?never_block:bool` to `Eio.Pool.use`. The default is `false`.

Quoting `pool.mli`:
> If `true` and the pool has reached maximum capacity, then a fresh resource is created to ensure that this `use` call does not wait for a resource to become available. This resource is immediately disposed after [f x] returns.

In my opinion, this feature makes `Eio.Pool` a valid choice for a wider range of use cases.
| Method | Availability | Overhead |
|-----|-----|-----|
| Creating a resource every time | ✅ Never blocks | 🔴 No reuse |
| Eio.Pool.use | 🔴 Can block | ✅ Maximal reuse |
| Eio.Pool.use ~never_block:true | ✅ Never blocks | 🟡 Some reuse |

### DIFF
--- a/lib_eio/pool.ml
+++ b/lib_eio/pool.ml
@@ -19,7 +19,7 @@ type 'a slot = 'a option ref
 module Cell = struct
   (* The possible behaviours are:
 
-     1. Suspender : In_transition -> Request            Suspender waits for a resource  
+     1. Suspender : In_transition -> Request            Suspender waits for a resource
         1.1. Resumer : Request -> Finished              Resumer then providers a resource
         1.2. Suspender : Request -> Finished            Suspender cancels
      2. Resumer : In_transition -> Resource             Resumer provides a spare resource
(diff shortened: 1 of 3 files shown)


lib_eio/pool.ml
```diff
@@ -114,25 +113,44 @@ let run_with t f slot =
...
   | In_transition ->
-      (* It would have been better if more resources were available.
-         If we still have capacity, add a new slot now. *)
-      maybe_add_slot t;
+    let current = Atomic.get t.slots in
+    match current < t.max_slots with
+    | false when never_block ->
+      (* We are at capacity, but cannot block.
+         Create a new resource to run f but don't add it to the pool. *)
+      Atomic.set cell Finished;
```
talex5, line 149: This isn't quite right. The cell might contain a `Resource` by now. If so, we should use that.

(probably needs an `Atomic.exchange cell Finished` here)

talex5 (commented):
Looks useful - thanks!

(I rebased and also updated the test to check that dispose works for the temporary resource)

lib_eio/pool.ml
SGrondin: Thanks, I'll fix it this weekend if you don't get to it first

SGrondin: Pushed. I have a remaining question for you that I'm keeping for the devs meeting, so I think merging this PR should wait until after the meeting.

talex5 (approved):
Thanks!

PR #658, opened 2023-12-16
Keep pool of systhreads for blocking operations

SGrondin:
It works (on all backends!) and stress testing hints at it being free of race conditions. Addresses #607 

This is still a work in progress, any and all suggestions and criticisms are welcome.

It has to keep a separate threadpool per domain as per [Interaction with systhreads](https://v2.ocaml.org/manual/parallelism.html#s:par_systhread_interaction) or else things quickly deadlock:
> The systhreads created on a particular domain remain pinned to that domain. Only one systhread at a time is allowed to run OCaml code on a particular domain.

General thoughts:
- I picked `run_in_systhread` as the place to implement this feature to allow all backends to benefit from it since they all call `Eio_unix.run_in_systhread`.
  - Would it be better to implement it in the scheduler as this [WIP branch does it](https://github.com/ocaml-multicore/eio/compare/main...talex5:eio:wip-thread-pool?expand=1)?
- I selected a scaling strategy that's totally different from Lwt's.
  - Lwt creates a new systhread whenever all systhreads are busy, except when there's already 1000 busy systhreads, in which case Lwt performs the job synchronously/blocking!
  - In comparison, this PR's threadpool will always make a new thread if all systhreads are busy. It then keeps up to ~25~20 threads to wait for new jobs (per domain).
- The fact that these threadpools are per-domain obviously runs the risk of increasing OS scheduler overhead due to the sheer number of threads we might end up creating.
  - I'm hoping that users will leverage the Executor_pool for the majority of their parallelism needs because it maximizes reuse of domains and therefore systhreads.
- ~I'm thinking it might be interesting to terminate threads that have been idle for a long time. It's worth trying and benchmarking, at least.~
  - This is now implemented
- ~Should Eio users be able to call `run_in_systhread`? Before OCaml 5 I frequently used `Lwt_preemptive.detach` to call C code that releases the runtime lock. Now users can just run that code on a separate domain. Would there be any benefit executing such code on a pooled systhread instead as opposed to e.g. `Executor_pool.submit_exn ~weight:0.0`?~
  - Users already have access to it: `Eio_unix.run_in_systhread`

### DIFF
--- a/examples/hello/dune
+++ b/examples/hello/dune
@@ -1,4 +1,3 @@
 (executable
   (name main)
-  (libraries eio_main)
-  (modes byte))
+  (libraries eio_main))
--- a/examples/hello/dune
+++ b/examples/hello/dune
@@ -1,3 +1,4 @@
 (executable
   (name main)
-  (libraries eio_main) (modes byte))
+  (libraries eio_main)
+  (modes byte))
--- a/examples/hello/dune
+++ b/examples/hello/dune
@@ -1,3 +1,3 @@
 (executable
   (name main)
-  (libraries eio_main))
+  (libraries eio_main) (modes byte))
--- a/lib_eio/unix/thread_pool.c
+++ b/lib_eio/unix/thread_pool.c
@@ -1,0 +1,113 @@
+/* This file is mostly taken from OCaml PR#12867.
+ * It exposes pthread_cond_timedwait(3) for timed_semaphore.ml.
+ * This file (and timed_semaphore.ml) can both be deleted once
+ * this feature is available in OCaml itself.
+*/
+
+#include <pthread.h>
+#include <errno.h>
+#include <string.h>
+#include <math.h>
+
+#include <caml/memory.h>
+#include <caml/alloc.h>
+#include <caml/fail.h>
+
+#define CAML_INTERNALS
+#include <caml/runtime_events.h> // for CAML_EV_BEGIN and CAML_EV_END
+#include <caml/sys.h> // for caml_strerror
+#if defined(_WIN32)
+#include <caml/winsupport.h> // for CAML_ULONGLONG_FILETIME
+#endif
+#undef CAML_INTERNALS
+
+
+#if defined(_WIN32)
+/* There are 11644473600 seconds between 1 January 1601 (the NT Epoch) and 1
+ * January 1970 (the Unix Epoch). FILETIME is measured in 100ns ticks.
+ */
+#define CAML_NT_EPOCH_100ns_TICKS 116444736000000000ULL
+#else
+#include <sys/time.h>
+#endif
+
+typedef pthread_cond_t* sync_condvar;
+#define Condition_val(v) (* (sync_condvar *) Data_custom_val(v))
+typedef pthread_mutex_t * sync_mutex;
+#define Mutex_val(v) (* ((sync_mutex *) Data_custom_val(v)))
+
+static void sync_check_error(int retcode, char * msg)
+{
+  char * err;
+  char buf[1024];
+  int errlen, msglen;
+  value str;
+
+  if (retcode == 0) return;
+  if (retcode == ENOMEM) caml_raise_out_of_memory();
+  err = caml_strerror(retcode, buf, sizeof(buf));
+  msglen = strlen(msg);
+  errlen = strlen(err);
+  str = caml_alloc_string(msglen + 2 + errlen);
+  memcpy (&Byte(str, 0), msg, msglen);
+  memcpy (&Byte(str, msglen), ": ", 2);
+  memcpy (&Byte(str, msglen + 2), err, errlen);
+  caml_raise_sys_error(str);
+}
+
+#if defined(_WIN32)
+static inline void sync_populate_timespec(double timeout_sec, struct timespec * ts)
+{
+  double integr, frac, until;
+  CAML_ULONGLONG_FILETIME utime;
+
+  GetSystemTimeAsFileTime(&utime.ft);
+  until = (utime.ul - CAML_NT_EPOCH_100ns_TICKS) * 1e-7 + timeout_sec;
+  frac = modf(until, &integr);
+  ts->tv_sec = (time_t)integr;
+  ts->tv_nsec = ceil(1e9 * frac);
+}
+#else
+static inline void sync_populate_timespec(double timeout_sec, struct timespec * ts)
+{
+  double integr, frac;
+  frac = modf(timeout_sec, &integr);
+
+  struct timeval tv;
+  gettimeofday(&tv, 0);
+  ts->tv_sec = tv.tv_sec + (time_t)integr;
+  ts->tv_nsec =
+    (uint64_t)tv.tv_usec * (uint64_t)1000 +
+    (uint64_t)ceil(1e9 * frac);
+
+  if (ts->tv_nsec >= 1e9) {
+    ts->tv_sec++;
+    ts->tv_nsec -= 1e9;
+  }
+}
+#endif
+
+value eio_unix_condition_timedwait(value v_cond, value v_mut,
+                                           value v_timeout_sec)
+{
+  CAMLparam3(v_cond, v_mut, v_timeout_sec);
+  sync_condvar cond = Condition_val(v_cond);
+  sync_mutex mut = Mutex_val(v_mut);
+  double timeout_sec = Double_val(v_timeout_sec);
+  int retcode;
+  struct timespec ts;
+
+  sync_populate_timespec(timeout_sec, &ts);
+
+  CAML_EV_BEGIN(EV_DOMAIN_CONDITION_WAIT);
+  caml_enter_blocking_section();
+  retcode = pthread_cond_timedwait(cond, mut, &ts);
+  caml_leave_blocking_section();
+  if (retcode == ETIMEDOUT) {
+    CAMLreturn(Val_false);
+  }
+  sync_check_error(retcode, "Condition.timed_wait");
+  CAML_EV_END(EV_DOMAIN_CONDITION_WAIT);
+
+  CAMLreturn(Val_true);
+}
--- a/lib_eio/unix/thread_pool.c
+++ b/lib_eio/unix/thread_pool.c
@@ -1,8 +1,7 @@
 /* This file is mostly taken from OCaml PR#12867.
  * It exposes pthread_cond_timedwait(3) for timed_semaphore.ml.
  * This file (and timed_semaphore.ml) can both be deleted once
- * this feature is available in OCaml itself.
-*/
+ * this feature is available in OCaml itself. */
 
 #include <pthread.h>
 #include <errno.h>
--- a/lib_eio/unix/thread_pool.ml
+++ b/lib_eio/unix/thread_pool.ml
@@ -3,7 +3,7 @@
    kept alive to wait for more work to arrive.
    This number was chosen somewhat arbitrarily but benchmarking
    shows it to be a good compromise. *)
-let max_standby_systhreads_per_domain = 25
+let max_standby_systhreads_per_domain = 20
 
 type job =
 | New
@@ -22,6 +22,8 @@ module Mailbox = struct
 
   let create () = { available = Semaphore.Binary.make false; cell = New }
 
+  let dummy = create ()
+
   let put mbox x =
     (* The Semaphore contains an atomic frontier,
        therefore [cell] does not need to be an atomic *)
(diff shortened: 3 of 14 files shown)


lib_eio/unix/thread_pool.ml
```diff
@@ -0,0 +1,155 @@
...
+    pool.num_available <- pool.num_available - 1;
+    true
+
+(* https://v2.ocaml.org/manual/parallelism.html#s:par_systhread_interaction
+   "Only one systhread at a time is allowed to run OCaml code on a particular domain."
+   So we keep a separate threadpool per domain. *)
+let key =
+  Domain.DLS.new_key @@ fun () ->
+    let pool = create () in
+    Domain.at_exit (fun () -> terminate pool);
```
polytypic, line 144: This is unfortunately problematic, because [`Domain.DLS` is currently not thread-safe](https://github.com/ocaml/ocaml/issues/12677) and, if I'd redesign it, I'd make it lock-free and specify that the init function (this function) passed to `Domain.DLS.new_key` may be called multiple times for a domain (but all concurrent accesses will eventually return only one of the results returned by the init function calls).  This means that anything that has side-effects that should be performed only once cannot be done inside the init function.

So, more specifically, it is possible that `Domain.at_exit` is called more than once and should not be done in the init function.  In [a Picos example](https://github.com/ocaml-multicore/picos/blob/7df26128a555edd9c69fba6199dc6eafd9e7b109/test/lib/elements/async_unix.ml#L131), I used a systhread safe locking approach (using `poll error` to perform initialization just once.

Currently `Domain.DLS` is completely unsafe with respect to systhreads (accesses of unrelated keys can mess things up), but I would hope that in the future it will be safe to allocate a new mutable object in the init function (currently it is not &mdash; nothing in `Domain.DLS` is currently thread-safe) and other systhread concurrency issues will be fixed.

lib_eio/unix/thread_pool.ml
```diff
@@ -0,0 +1,155 @@
...
+    then Mailbox.put (Array.unsafe_get pool.threads i) Exit
+  done
+
+(* Can't [assert false] in [@poll error] code because it would allocate *)
+exception Assertion_failure
+
+(* This function is called from within a [@poll error] function,
+   so we cannot allocate, call OCaml functions, or use for/while loops.
+   See https://discuss.ocaml.org/t/using-poll-error-attribute-to-implement-systhread-safe-data-structures/12804
+   The only option left to us is to unroll the loop by hand.
```
polytypic, line 66: You could also use lock-free programming techniques instead of unrolling.  The critical part is when you update the state.  At that point you need to atomically check that the state is consistent with the update (`compare`) before the update (`set`) is done.

Essentially something like this:

```ocaml
(** Atomically determines whether to keep a thread. *)
let[@poll never] atomically_should_keep pool =
  pool.num_available < Array.length pool.available &&
  not pool.terminating

(** Non-atomically tries to find a slot for a thread to keep. *)
let try_find_available_slot pool =
  let rec loop i =
    if i < Array.length pool.available then
      if Array.unsafe_get pool.available i then
        Some i
      else
        loop (i + 1)
    else
      None
  in
  loop 0

(** Tries to atomically keep the thread at specified slot. *)
let[@poll error] try_atomically_keep pool i mbox =
  not pool.terminating &&
  Array.unsafe_get pool.available i && begin
    Array.unsafe_set pool.available i true;
    Array.unsafe_set pool.threads i mabox;
    pool.num_available <- pool.num_available + 1;
    true
  end

let rec keep_thread_or_exit pool mbox =
  if atomically_should_keep pool then
    match try_find_available_slot pool with
    | Some i ->
      if not (try_atomically_keep pool i mbox) then begin
        keep_thread_or_exit pool mbox
      end
    | None ->
      keep_thread_or_exit pool mbox
  else
    raise Thread.Exit
```

The above is only to give an idea.  I have neither tested nor compiled the above.

lib_eio/unix/thread_pool.ml
SGrondin: Thank you for this information. I'm definitely surprised by this, as the mli for `Domain.DLS` even shows an example that combines `Domain.DLS.new_key` with `Domain.at_exit`, but as we've discussed elsewhere both those features have deep issues that I need to work around.

lib_eio/unix/thread_pool.ml
SGrondin: That's pretty much what I had in the first version (with atomics)! I like the idea of using it with `[@poll error]` code, that's neat.

SGrondin:
I had to revert back to doing reads and writes inside a single `[@poll error]` section because I would occasionally see extreme contention (100ms+) with retry-loop CAS code.
This latest commit automatically scales down a pool that doesn't enough usage.
I had to implement [`Condition.timed_wait` (using `pthread_cond_timedwait`)](https://github.com/ocaml/ocaml/issues/4104#issuecomment-473674943) in order to get semaphores with timeouts.
If timed semaphores stay in this PR until it's merged then I'll open a PR to add all that to OCaml itself. As Xavier pointed out it's a desired feature, they just need someone to take the time to do it.

talex5:
> I would occasionally see extreme contention (100ms+) with retry-loop CAS code.

What are you using to test this? We only use sys-threads to run system calls, which take a relatively long time, so I wouldn't expect to see much contention here.

SGrondin:
@talex5 `stress/stress_systhreads.ml`, added in this PR.

It's meant to be a worst case scenario: `eio_posix` so no io_uring, 5 domains on a 4 core CPU, 100 fibers per domain, every fiber simultaneously writing to the disk with file contention as well.
All those syscalls starting nearly simultaneously was causing extreme contention in the CAS loops so I rewrote those sections back to the original style and compared the 2 versions.
The difference was visible at human scale! I saw it stall for close to half a second once.
In comparison, the current code never fails to write and runs perfectly smooth.
The only downside is the visual ugliness of the unrolled loop; it's more predictable, more performant, less overhead, etc.

As a side note, this PR led to my first `ocaml` repo PR! https://github.com/ocaml/ocaml/pull/12867 Please take a look if you're familiar with the posix, time and timers area of the codebase.

lib_eio/unix/thread_pool.c
```diff
@@ -0,0 +1,113 @@
...
+
+  CAML_EV_BEGIN(EV_DOMAIN_CONDITION_WAIT);
+  caml_enter_blocking_section();
+  retcode = pthread_cond_timedwait(cond, mut, &ts);
+  caml_leave_blocking_section();
+  if (retcode == ETIMEDOUT) {
+    CAMLreturn(Val_false);
+  }
+  sync_check_error(retcode, "Condition.timed_wait");
+  CAML_EV_END(EV_DOMAIN_CONDITION_WAIT);
```
SGrondin, line 110: It seems wrong to have this line _after_ both `CAMLreturn(Val_false)` and `sync_check_error()`, but that's how the standard library does it.

examples/hello/dune
```diff
@@ -1,3 +1,4 @@
 (executable
   (name main)
-  (libraries eio_main))
+  (libraries eio_main)
+  (modes byte))
```
talex5, line 4: I guess this was for testing something, but we shouldn't complicate the hello-world example.

lib_eio/unix/thread_pool.ml
```diff
@@ -0,0 +1,212 @@
...
+    n_available = 0;
+    terminating = false;
+  }
+
+(* This function is necessary in order to immediately terminate the systhreads
+   on stand-by, without having to wait for them to shut down automatically.
+   Without this termination mechanism, domains that have executed a call on
+   a systhread in the last [max_wait_seconds] seconds would have to wait for
+   their timeout to occur, introducing long unwanted pauses. *)
+let terminate pool =
```
talex5, line 105: Does this need to be `[@poll error]` too? Threads can finish while it's running the loop.

lib_eio/unix/timed_semaphore.ml
```diff
@@ -0,0 +1,53 @@
+(* This file is mostly taken from OCaml PR#12867.
+ * It exposes pthread_cond_timedwait(3) for timed_semaphore.ml.
+ * This file (and thread_pool.c) can both be deleted once
+ * this feature is available in OCaml itself. *)
```
talex5, line 4: OCaml is LGPL whereas Eio is ISC, and the OCaml devs have previously not been willing to relicense code for us.

stress/dune
```diff
@@ -1,3 +1,4 @@
 (executables
-  (names stress_semaphore stress_proc)
-  (libraries eio_main))
+  (names stress_semaphore stress_proc stress_systhreads)
+  (libraries eio_main)
+  (modes byte))
```
talex5, line 4: Running stress tests in byte mode doesn't seem very realistic.

talex5 (commented):
> All those syscalls starting nearly simultaneously was causing extreme contention in the CAS loops so I rewrote just those critical sections back to the original style and compared the 2 versions.
The difference was visible at human scale! I even saw the CAS loop version stall for close to half a second once.

I'm going to do some benchmarking to try to get to the bottom of this. It doesn't make sense to me that contention could be noticeable here. In the worst case, OCaml decides to switch threads just as we allocate a cons cell for the free list, and we have to do it again. This only takes a few cycles and can't happen more often than OCaml switches threads (every 20ms or so).

Also, immediately after doing this we add the fiber to the run queue, which is also a CAS operation and should therefore have the same problem.

I'm also unclear on the rules for assuming thread switches don't happen. I think someone suggested that poll error doesn't work in bytecode, but not sure.

> Would it be better to implement it in the scheduler

The advantage there is that the scheduler can handle the timeouts (since it already handles time) and we don't need this condition wait stuff. It shouldn't be any slower, as we have to perform an effect to suspend the calling fiber anyway. I'll have a go at doing that this week and see if it works...

polytypic:
> I'm going to do some benchmarking to try to get to the bottom of this.

👍 I'd also love to understand this phenomena better.

My _assumption_ is that when you use more domains than there are hardware threads then the OS will have to switch between domains.  What I assume then happens is that the mechanism that OCaml uses internally to switch between systhreads doesn't notice this and basically switches between systhreads prematurely before a systhread has really had a chance to run much at all, because the domain wasn't running (some other domain was running instead).  This could indeed cause very long delays.  However, this is just my speculation &mdash; I have not yet verified whether this is actually what the runtime would effectively do.  If my speculation is roughly correct, then, first of all, one really should not spawn more domains than there are hardware threads.  Second, it might be useful to consider improvements to the the internal ticking mechanism that OCaml uses to switch between systhreads on a domain to take into account the possibility that a domain might not actually get CPU time during the tick interval.

SGrondin:
> What I assume then happens is that the mechanism that OCaml uses internally to switch between systhreads doesn't notice this and basically switches between systhreads prematurely before a systhread has really had a chance to run much at all, because the domain wasn't running (some other domain was running instead).

This is my theory as well.

talex5:
I started writing a benchmark and quickly discovered a strange delay joining domains with systhreads that was skewing the results: https://github.com/ocaml/ocaml/issues/12948

domain-join

SGrondin:
@talex5 I just pushed a commit that replaces both `available: bool array;` and `n_available: int` with a bitfield as per Arthur's suggestion in the dev meeting.
It allowed me to remove that horrible unrolled loop. Bitfields introduce some bit twiddling complexity, but I think it's worth it overall.

lib_eio/unix/thread_pool.ml
SGrondin: (discussed in the dev meeting) Threads check for `terminating`.

lib_eio/unix/timed_semaphore.ml
SGrondin: (discussed in the dev meeting) The waiting will be done using the scheduler instead, the C code will go away.

polytypic:
Hmm... Note that within a `[@poll error]` function you can do quite a bit aside from allocating.  So, if you just move the allocations outside of the `[@poll error]` function you can, for example, manipulate a dynamic list of nodes.  IOW, you can, for example, implement a wait-free systhread safe bounded stack/cache as a linked list:

```ocaml
module Wait_free_systhread_safe_bounded_stack : sig
  type 'a t

  val create : capacity:int -> 'a t
  val try_push : 'a t -> 'a -> bool
  val pop_opt : 'a t -> 'a option
end = struct
  type ('a, _) node =
    | Null : ('a, [> `Null ]) node
    | Node : { mutable next : 'a link; value : 'a } -> ('a, [> `Node ]) node

  and 'a link = Link : ('a, [< `Null | `Node ]) node -> 'a link [@@unboxed]

  type 'a t = { capacity : int; mutable count : int; mutable nodes : 'a link }

  let create ~capacity = { capacity; count = 0; nodes = Link Null }

  let[@poll error] try_push t (Node node_r as node : (_, [< `Node ]) node) =
    t.count < t.capacity
    && begin
         t.count <- t.count + 1;
         node_r.next <- t.nodes;
         t.nodes <- Link node;
         true
       end

  let try_push t value = try_push t (Node { next = Link Null; value })

  let[@poll error] pop_opt t =
    match t.nodes with
    | Link Null -> Null
    | Link (Node node_r as node) ->
        t.count <- t.count - 1;
        t.nodes <- node_r.next;
        node

  let pop_opt t =
    match pop_opt t with Null -> None | Node node_r -> Some node_r.value
end
```

lib_eio/unix/thread_pool.ml
```diff
@@ -0,0 +1,216 @@
...
+  pool.terminating <- true;
+  if not (Bitfield.is_empty pool.available) then
+    for i = 0 to max_standby_systhreads_per_domain - 1 do
+      if Bitfield.get pool.available i
+      then Mailbox.put pool.threads.(i) Exit
```
polytypic, line 149: Hmm... It is usually prudent to join with any domains, threads, or fibers.  Granted there shouldn't be much that could go wrong here.

lib_eio/unix/thread_pool.ml
```diff
@@ -0,0 +1,216 @@
...
+  else (pool.initialized <- true; true)
+
+let run_on_systhread ~enqueue fn =
+  let pool = Domain.DLS.get key in
+  if needs_init pool then Domain.at_exit (fun () -> terminate pool);
+  let mbox =
+    let res = ref Mailbox.dummy in
+    if try_get_thread pool res
+    then !res
+    else make_thread pool
```
polytypic, line 214: Hmm... Another way to organize the pool would be to separate the management of the threads from the queue of jobs to perform.  Basically, jobs would be put into a queue and worker threads would take jobs from the queue to perform.  This way you could avoid spawning a new thread for each new job after the pool capacity has been exhausted.

SGrondin: Thank you for your review! I always appreciate you sharing your expertise in this area.

> This way you could avoid spawning a new thread for each new job after the pool capacity has been exhausted.

At the moment that's done on purpose. I absolutely did not want to risk user code deadlocking _only_ when the threadpool is at maximum capacity. This tends to cause bugs in production that are hard to discover during development. Do you have a recommendation?

> Basically, jobs would be put into a queue and worker threads would take jobs from the queue to perform.

I wanted to make sure that jobs would _always_ go to Thread 0 if Thread 0 is available, otherwise _always_ go to Thread 1 if Thread 1 is available, etc. This way we can guarantee that only the minimum number of threads are running at any given time because it allows the other threads to timeout and shutdown. In other words, the scheduling unfairness is on purpose. If there's a better way to achieve this effect, I'm all ears!

> Instead of having a separate mailbox (and semaphore) for every worker thread

In yesterday's dev meeting we discussed how @talex5 will be implementing the wait/timeout logic in the scheduler itself so we should be able to get rid of all the semaphores and all the C code I ported from my own [ocaml/ocaml#12867](https://github.com/ocaml/ocaml/pull/12867) PR. I do like your suggestion a lot though and may switch to it if beneficial after the scheduler changes.

polytypic: >  I absolutely did not want to risk user code deadlocking only when the threadpool is at maximum capacity.

That is a good point.

I think there are two completely different possible designs/semantics:
1. One that fundamentally has a hard limit on the number of systhreads to use to run things concurrently.
2. One that fundamentally has no such limit.

The current implementation is a bit of both.  It uses a fixed size array to manage the systhreads.  But, then, in case more concurrency is requested, it runs even more systhreads.

I think this adds to the complexity.  I suspect that changing the design to clearly implement one or the other could simplify the internal implementation.  Perhaps not (only) in terms of lines of code, but also in terms of conceptual clarity.  Indeed, it wasn't entirely clear to me that spawning more threads was really intentional.

So, I would recommend trying to rethink the internals with the idea that there is no limit on the number of systhreads.

polytypic: > This way we can guarantee that only the minimum number of threads are running at any given time because it allows the other threads to timeout and shutdown. In other words, the scheduling unfairness is on purpose. If there's a better way to achieve this effect, I'm all ears!

I see.  So, what you really want is to use the thread that was the most recent thread to become idle.  That is because then the threads that became idle earlier can timeout.   IOW, for picking the next idle thread to pass a new job to you want LIFO and for picking the next idle thread to die/kill you want FIFO.  Basically a deque, except that the idea here is that threads remove themselves from the deque.

I suspect a dynamic doubly-linked list could be used for this and it would likely be a bit simpler (fewer lines of code and also a simpler concept to understand) than the array with a bitfield.

But this is also assuming that threads timeout on their own.  An alternative approach could be to manage the threads externally.  Idle threads would add themselves to a data structure (along with a timestamp, for example).  Then some external mechanism would remove old threads.  With Eio this external mechanism could be something like a per domain fiber that periodically explicitly kills old idle threads when there are some idle threads.  This approach could also remove the need for timedwait (which I think really would be a great addition to the Stdlib, but it may take quite some time before we get that).

talex5:
I rebased on the benchmark commit and collected some measurements in #679:

systhread-pool

The commits are (from left to right):

1. 964ed27 is before the thread-pool
2. 5b39011 (with the branch icon) is "Keep pool of systhreads for blocking operations"
3. 0807692 is "Threadpool: Use safe-points instead of atomics"
4. 006afcb is "Minor thread pool cleanups": the tip of this PR
5. f7a6c85 is identical, to check if 4 was a fluke

So from this it looks to me like:
- Thread-pooling helps a lot (2 to 3 times faster)
- Getting rid of atomics makes no detectable difference (given there is some noise anyway)

This seems very reasonable to me, but if you can modify the benchmark to show the latency problem you saw, or capture a trace with `eio-trace record`, that would be useful. Otherwise, I'll probably go back to the simpler version.

SGrondin:
Closing in favor of #681

talex5:
BTW, I found out I why this PR wouldn't build on my machine: https://github.com/ocaml/dune/issues/9979

PR #659, opened 2023-12-18
Mention Path module in File and Fs documentation

talex5:
Someone trying to open a file may start looking in these modules. Explain what each module is for and point people at the Path module.

Based on feedback from @clecat.


clecat:
Thanks a lot for those modifications

Issue #649, opened 2023-11-20
Full tracing support


talex5:
#635 added initial support for tracing using OCaml's runtime events system. However, more features are needed to make it useful:

- [x] Add events about cancellation contexts (#656).
- [x] Add more labels to make the traces clearer (see #661).
- [x] Trace syscalls.
- [x] Trace systheads.
- [x] Write a visualisation tool that can take advantage of the extra structure in Eio (https://github.com/ocaml-multicore/eio-trace).

PR #661, opened 2023-12-21
Tracing: add labels to switches

talex5:
This makes the traces easier to read. Here's the network example with the new labels:

switch

I also modified `Net.run_server` to use one fewer switch, to clean up the trace.

Part of #649.


SGrondin:
Would it be less ambiguous to use the label `Fiber.List.all` instead of just `all`, for example? If the rendered image looks too cluttered maybe it's something visualization tools could choose to truncate to `all` themselves?

talex5:
It's just `Fiber.all` in this case. My guess is that a collection of fibers tagged "all" will be clear enough, but we'll have to get some experience with larger traces to be sure.

PR #662, opened 2023-12-22
Fix non-idempotent tests

SGrondin:
I found these 2 non-idempotent tests when saving a file while `dune runtest -w` is running interrupted a test in the middle and left `runtest` unable to succeed until `tmp-foo.txt` was removed by hand.


talex5 (approved):
Thanks!

PR #663, opened 2023-12-31
Optimise Flow.copy with Buf_read.as_flow

talex5:
By default, `Flow.copy` creates a 4KB buffer and copies the data through that. However, if the source of the copy is a buffered reader then it is much more efficient to use its buffer directly.

This updates the flow you get from `Buf_read.as_flow` to offer this optimisation, and also updates the eio_posix backend's flow to use this optimisation when available (eio_linux already supported this).

Detected in a benchmark by @leostera.

Possibly we should increase the default buffer size too.

This PR is made of two commits. The first just adds a benchmark to show the effect of the change. On my Linux machine, I get:

```
buf_read: 1091.79 MB/s   # Before
buf_read: 3306.98 MB/s   # After
```

### DIFF
--- a/bench/bench_copy.ml
+++ b/bench/bench_copy.ml
@@ -1,3 +1,5 @@
+(* A client opens a connection to an echo service and sends a load of data via it. *)
+
 open Eio.Std
 
 let chunk_size = 1 lsl 16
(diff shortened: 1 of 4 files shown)


leostera:
Excited to see this! ✨

If I find any other sore spots I'll try and be a little more proactive opening issues or tagging y'all 🙏🏼 it's been a busy week over here.

And please feel more than welcome to bench against Riot, it may help me find bottlenecks like this too!

Happy new year team, keep kicking ass 👏🏼

SGrondin (approved):
I haven't fully explored this area of the code yet but this looks good to me!

PR #666, opened 2024-01-05
Fix quoting of quotes in process error messages

talex5:
When displaying an argument with quotes, they should be escaped to avoid ambiguity (`Fmt.quote` doesn't do this; it just puts quotes around the outside of the string).


SGrondin (approved):
Interesting, that's news to me as well

PR #675, opened 2024-01-26
Improve traces

talex5:
- Trace `Eio_unix.run_in_systhread`, adding labels for all operations currently using this. Operations performed in systhreads weren't previously shown at all.
- Instead of waiting for fibers to finish in a protected block, use a non-cancellable wait. This simplifies the traces in some cases.
- Trace `Unix.close`, `Uring.submit` and `traceln`.

The network example now looks like this:

net


SGrondin (approved):
Looks great, this is going to make #658 interesting to try out.
Once this is merged I'll rebase #658

talex5:
That would be great - I'm keen to look at the thread pooling next!

PR #678, opened 2024-01-31
Benchmark Eio_unix.run_in_systhread

talex5:
This is currently very slow because we create a new systhread for each operation. This is to allow measuring the speed-up from #658. Each thread does a call to `getuid`, which is one of the fastest system calls.

On my machine, a blocking call to `Unix.getuid` takes around 60ns on average. Using a systhread increases that to about 30,000ns. #658 makes using a systhread about three times faster.

### DIFF
--- a/bench/bench_systhread.ml
+++ b/bench/bench_systhread.ml
@@ -1,0 +1,76 @@
+(* Measure the overhead of [Eio_unix.run_in_systhread]. *)
+
+open Eio.Std
+
+let n_iters = 1000
+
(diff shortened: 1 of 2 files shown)


SGrondin:
> #658 makes using a systhread about three times faster.

I'm glad to hear that! Hopefully it will be even better once the waiting is done in the scheduler, which I think will allow it to stop using semaphores entirely.

SGrondin (approved):

PR #680, opened 2024-02-02
Eio backend for JavaScript environments

vouillon:
This provides:
- `eio_js_backend`, a simple backend;
- `js_of_ocaml-eio`, a counterpart to `js_of_ocaml-lwt`;
- `eio_brr`, Eio support for Brr (Eio variants of Brr functions that returns directly instead of returning a future).

This has been heavily inspired by #405. Compared to this PR, there is a clear separation between the scheduler and the JavaScript bindings. Also, it is possible to use Eio from JavaScript callbacks (e.g. from event handlers) as was attempted in #534.

The scheduler is quite different from other Eio backends. It provides a `start` function that starts executing asynchronously a function in a context where Eio effects can be performed instead of a blocking `run` function. Indeed, one cannot block waiting for events in JavaScript but one needs to return to the JavaScript event loop. The `start` function can be called in JavaScript callbacks to use Eio.

`eio_jsoo` is not complete yet.


talex5:
Thanks - this looks great (I only read the `eio_js_backend.ml` file; it looks sensible to me).

For others who want to try the example:

1. You need to install `opam install js_of_ocaml-ppx` manually (this is why CI is failing).
2. You need to `dune build @default` to get the resource files. `dune build` alone isn't enough.
3. Opening the index.html in a browser doesn't work. The Firefox console doesn't show anything by default, but if you switch to "Multiprocess (slower)" then it shows "Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at file:///home/user/work/eio/_build/default/lib_js_of_ocaml_eio/examples/boulderdash/maps.txt. (Reason: CORS request not http)". To fix that, I ran `docker run --rm -it -p8080:80 -v (pwd)/_build/default/lib_js_of_ocaml_eio/examples/boulderdash/:/usr/share/nginx/html:ro nginx` and then opened http://localhost:8080/.
4. Then I was able to play the example game, and it worked.

PR #681, opened 2024-02-03
Keep pool of systhreads for blocking operations (simplified)

talex5:
This is a simplified version of @SGrondin's #658:

- Add `Run_in_systhread` effect. This makes the thread-pool part of the
  scheduler's state, avoiding the need for DLS.

- Remove `max_standby_systhreads_per_domain`. Suggested by @polytypic. All threads are returned to the free-list.

- Simplify termination. Use one atomic instead of two, by using a custom list type.

- Use the scheduler's timer to drop idle threads. This avoids the need to add timeouts to semaphores, which would add a dependency on https://github.com/ocaml/ocaml/pull/12867. I had to modify `Zzz` slightly to allow non-fiber timeouts.

The cleanup logic is that when a fiber running a task resumes, it adds a 20ms timer to the scheduler (if we don't already have one). When it fires, it tells all idle threads to exit. More complex schemes are possible, but having to recreate a few threads every 20ms shouldn't matter.

### DIFF
--- a/lib_eio/unix/dune
+++ b/lib_eio/unix/dune
@@ -5,7 +5,7 @@
   (language c)
   (include_dirs include)
   (names fork_action stubs))
- (libraries eio unix threads mtime.clock.os))
+ (libraries eio eio.utils unix threads mtime.clock.os))
 
 (rule
  (enabled_if %{bin-available:lintcstubs_arity_cmt})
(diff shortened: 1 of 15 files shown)


SGrondin (approved):
I don't see any issues with it and I think it's a solid first pass. I have not tested it yet however, unlike #658 against which I developed for weeks, so I'd hold off on releasing it right away. There are also some optimizations I'd like to make and benchmark once it's merged.

talex5:
Thanks - I'm very interested in the results of the testing! And especially to find out if you still see the latency problem with this version.

PR #684, opened 2024-02-06
Make Switch.on_release thread safe

### LINKED ISSUE #669: Switches should be domain-safe
Currently, a switch can only be used from the domain in which it was created (and it rejects attempts to use it from another domain). It would be useful to remove this restriction. It's particularly important when implementing any kind of cache or pool that is shared between domains, since a resource allocated in one domain should be usable by the others.

Perhaps the `fibers` field could be made atomic, allowing any domain to prevent a switch from finishing while it's adding a resource to it. e.g. to allocate a resource:

1. Atomically increment `fibers` from a positive value.
2. Create the resource and attach it to the switch.
3. Decrement `fibers`.

The `on_release` list would need to become thread-safe too.

If changing all operations to do this is too difficult or hurts performance, it might be possible instead to add just a thread-safe transfer operation. Then other domains could open resources in their own domain and then transfer them to the cache's switch afterwards. A transfer operation would be useful even with a single domain.

Requested by @art-w and @clecat.

talex5:
This is needed to allow resource pools to be shared between domains. Fixes #669.

It works by putting a mutex around the `on_release` list. I tried a lock-free version too, but it was quite messy.

There is one change to the existing behaviour: it used to be possible to attach new resources to a switch in a release handler, but now that is rejected. I don't think the old behaviour was useful, but it could be brought back if needed (we'd just need to take the mutex again at the end).

I was thinking of deprecating `remove_hook` in favour of `try_remove_hook`, but all existing uses were safe so I decided not to do that for now.

Note: There are some places where we do `Switch.check` before creating a resource. This isn't thread-safe, and might pass even after the switch has ended if called from a different domain. That shouldn't be a problem (the switch could always have ended after the check anyway), but might annoy TSan.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ dscheck:
 stress:
 	dune exec -- ./stress/stress_proc.exe
 	dune exec -- ./stress/stress_semaphore.exe
+	dune exec -- ./stress/stress_release.exe
 
 docker:
 	docker build -t eio .
--- a/lib_eio/core/eio__core.mli
+++ b/lib_eio/core/eio__core.mli
@@ -87,24 +87,41 @@ module Switch : sig
       Release handlers are run in LIFO order, in series.
 
       Note that [fn] is called within a {!Cancel.protect}, since aborting clean-up actions is usually a bad idea
-      and the switch may have been cancelled by the time it runs. *)
+      and the switch may have been cancelled by the time it runs.
+      You cannot attach new resources to a switch once the cancel hooks start to run.
+
+      This function is thread-safe (but not signal-safe).
+      If the switch finishes before [fn] can be registered,
+      it raises [Invalid_argument] and runs [fn] immediately instead. *)
 
   type hook
   (** A handle for removing a clean-up callback. *)
 
   val null_hook : hook
-  (** A dummy hook. Removing it does nothing. *)
+  (** A dummy hook. [try_remove_hook null_hook = false]. *)
 
   val on_release_cancellable : t -> (unit -> unit) -> hook
   (** Like [on_release], but the handler can be removed later.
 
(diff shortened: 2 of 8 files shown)


polytypic:
> It works by putting a mutex around the on_release list. I tried a lock-free version too, but it was quite messy.

I took a very brief look and, IIUC, then what you basically need is a lock-free set.  `on_release_cancellable` adds a `hook` to the set and you can later decide to `remove_hook` it.  Also, when the switch finishes, all the hooks in the set are run.

I developed a lock-free starvation avoiding data structure to deal with a similar problem in Picos.  See [here](https://github.com/ocaml-multicore/picos/blob/7df26128a555edd9c69fba6199dc6eafd9e7b109/lib/picos/bootstrap.ml#L37-L94).  Both `try_attach` and `detach` are lock-free and work in amortized constant time.  The elements in the set are mutable (think `_ option Atomic.t`) and `detach` clears the element so that the payload is not leaked.  Nodes are removed from the set only lazily.  This is achieved by maintaining a `balance`.  Once the balance crosses a threshold, both attach and detach operations will perform a `gc` to drop nodes whose elements have been detached.  This ensures that the set will not use more than a constant factor more space than strictly necessary.  Because both attach and detach perform the gc at the same point the GC is effectively wait-free (i.e. if you e.g. attach and that requires gc, then the only reason to require a retry is that another operation performed the gc for you).

talex5:
Thanks - that does sound like what we want. I'm not sure making it lock-free is necessary yet - adding the mutex didn't seem to have any noticeable effect on the benchmarks and there isn't likely to be much contention for the Irmin use (opening a file takes much longer than taking the mutex and appending to the list). But if it becomes a problem I'll use that code.

PR #694, opened 2024-02-15
eio_posix: use directory FDs instead of realpath

talex5:
realpath was an old hack from the libuv days and isn't safe against races with symlinks being updated.

This should be faster too in some cases.

### DIFF
--- a/lib_eio/unix/eio_unix.mli
+++ b/lib_eio/unix/eio_unix.mli
@@ -103,6 +103,7 @@ module Private : sig
   module Thread_pool = Thread_pool
 
   val read_link : Fd.t option -> string -> string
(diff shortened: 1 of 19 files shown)


SGrondin (commented):
I've read through it. I don't have enough knowledge of this area of the code to formally approve it, but this area is well tested and I have a codebase that performs a lot of file operations so I'll test this PR on it.

talex5:
The only noticeable change in the benchmarks is for the Path.stat rmtree time:

Before: Removed in 20.94 s
After: Removed in 6.65 s

talex5:
I've merged this so we can get some more testing. Please file issues with any problems you notice!

PR #704, opened 2024-02-25
Fix to pass with both old and new Kcas

polytypic:
The `Loc.t` type has changed in Kcas, which makes a MDX block fail with upcoming Kcas.  This changes the block such that it will pass with both old and future Kcas.


talex5 (approved):
Thanks!

PR #706, opened 2024-02-29
Require MDX < 2.4.0

talex5:
The MDX devs have decided to change the behaviour of includes. What was previously a request to copy some text from one file to another now, in mdx 2.4.0, will also execute it!!

See https://github.com/realworldocaml/mdx/pull/446


talex5:
Hmm, that failed on macos for some reason:
```
Fatal error: exception Mdx_test.Test_block_failure(_, "./examples/signals/main.ml: No such file or directory")
```
Constrain to use the old version instead for now.

PR #710, opened 2024-03-10
Bound MDX version on all packages

talex5:
#706 only added the bound in one place, but when releasing it's needed on all packages.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -34,7 +34,7 @@
  (depends
   (alcotest (and (>= 1.7.0) :with-test))
   (eio (= :version))
-  (mdx (and (>= 2.2.0) :with-test))
+  (mdx (and (>= 2.2.0) (< 2.4.0) :with-test))
   (logs (and (>= 0.7.0) :with-test))
   (fmt (>= 0.8.9))
   (cmdliner (and (>= 1.1.0) :with-test))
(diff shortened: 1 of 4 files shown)


dune-project
```diff
@@ -34,7 +34,7 @@
  (depends
   (alcotest (and (>= 1.7.0) :with-test))
   (eio (= :version))
-  (mdx (and (>= 2.2.0) :with-test))
+  (mdx (and (>= 2.2.0) (< 2.4.0) :with-test))
```
yawaramin, line 37: Any reason why <2.4.0 specifically? Is there a breaking change in 2.4.0?

anmonteiro: 2.4 introduced OCaml 5.2 compatibility.

talex5: 2.4.0 started executing code while copying it. Looks like 2.4.1 has reverted this change (https://github.com/realworldocaml/mdx/releases/tag/2.4.1).

yawaramin: Ah, got it. I guess in the future >2.4.0 <3.0.0 would also work, if they agree to stick to a semantic versioning logic.

talex5: I've updated it to 2.4.1 in #712.

PR #713, opened 2024-03-20
(draft) Eio.Net.connect: optionally bind before connect

art-w:
I'm looking to port some existing code which does a [bind before connect](https://idea.popcount.org/2014-04-03-bind-before-connect/) and couldn't find a way to do it with Eio. I wasn't familiar with this trick, but it's apparently useful when you have multiple public IPs and need to control which one will be seen by the server.

Let me know if there's a better way to do this!.. I'm marking this PR as a draft because I'm starting to think this might actually require a new `bind_connect` function, to also expose the `REUSEADDR/PORT` options as they only make sense when `?bind` is set.

### DIFF
--- a/lib_eio/mock/net.ml
+++ b/lib_eio/mock/net.ml
@@ -36,7 +36,7 @@ module Impl = struct
     Switch.on_release sw (fun () -> Eio.Resource.close socket);
     socket
 
-  let connect t ~sw addr =
+  let connect t ~sw ~options:_ addr =
     traceln "%s: connect to %a" t.label Eio.Net.Sockaddr.pp addr;
     let socket = Handler.run t.on_connect in
     Switch.on_release sw (fun () -> Eio.Flow.close socket);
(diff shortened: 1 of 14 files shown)


SGrondin:
I'd interested to hear what made you select this approach over exposing a `bind` function to the user?

talex5:
This seems reasonable to me.

Possibly we should have a dedicated `options` argument (extensible variant) as there will likely be loads more special features that could be added (see e.g. #575).

e.g. something like
```ocaml
Net.connect net dst
   ~options:[Source_addr src]
```

Need to decide what to do if a backend doesn't support an option. e.g. mirage-tcpip doesn't look like it supports this (https://github.com/mirage/mirage-tcpip/blob/75382e296ff6779338c3b7b069c885d5d8e10eeb/src/core/tcp.mli#L76).

talex5:
Updated and merged as #875 - thanks!

PR #716, opened 2024-03-22
Introduce HOFs that wrap functions to run with a locked mutex

### LINKED ISSUE #336: Mutex API ergonomics
Can we draw inspiration from Rust's [Mutex](https://doc.rust-lang.org/std/sync/struct.Mutex.html) to make `Eio.Mutex` a bit more ergonomic? E.g. usage adapted from https://github.com/ocaml-multicore/eio#the-rest-mutex-semaphore-and-condition:

```ocaml
open Eio

let of_path = Mutex.create

let save m data =
  Mutex.use_rw m ~protect:true @@ fun path ->
  Path.save path data ~create:(`Or_truncate 0o644)

let load m = Mutex.use_ro m Path.load
```

We would need signatures like this:

```ocaml
module Mutex : sig
  type _ t

  val create : 'a -> 'a t
  (** Create in an unlocked state holding the data. *)

  val use_rw : protect:bool -> 'a t -> ('a -> 'b) -> 'b
  (** Lock for read-write access. *)

  val use_ro : 'a t -> ('a -> 'b) -> 'b
  (** Lock for read access. *)
end
```

Having said that, I realize we don't have borrow/move checking in OCaml so we wouldn't get the same level of safety. But it should make it a bit more ergonomic as the key idea is that a mutex is always paired with its data.

yawaramin:
Introduce wrappers that can turn any function `f : a -> b` into a new function that runs with a wrapped mutex. `a` can even be a mutable value that can then be modified only while the mutex is held.

Fix #336

### DIFF
--- a/lib_eio/eio_mutex.ml
+++ b/lib_eio/eio_mutex.ml
@@ -119,3 +119,18 @@ let use_rw ~protect t fn =
     (* {mutex t R * locked t} *)
     poison t ex;
     raise ex
+
+let wrap_ro t f x =
+  lock t;
+  match f x with
+  | y -> unlock t; y
+  | exception ex -> unlock t; raise ex
+
+let wrap_rw ~protect t f x =
+  lock t;
+  match if protect then Cancel.protect f else f x with
+  | y -> unlock t; y
+  | exception ex ->
+    poison t ex;
+    raise ex
+
--- a/lib_eio/eio_mutex.ml
+++ b/lib_eio/eio_mutex.ml
@@ -128,7 +128,8 @@ let wrap_ro t f x =
 
 let wrap_rw ~protect t f x =
   lock t;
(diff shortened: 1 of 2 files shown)


talex5:
This seems not quite the same as your #336 proposal - in fact, it seems incompatible with that idea (of storing the protected resource inside the mutex). Here it's passed as an argument instead.

This "wrap" is effectively the same as "use", except using the alternative API style of taking an argument for the function.

e.g. it allows writing `wrap_ro m foo x` instead of `use_ro m (fun () -> foo x)`.

The problem usually with this form is that it's really easy to mess it up when you have more than one argument:

```ocaml
wrap_ro m foo x y
```

Tthis takes `m` while creating the partially-applied `foo x`, but then runs `foo x y` without the lock, which was probably not what was wanted.

PR #719, opened 2024-03-24
Benchmarks: record uname, Eio backend, and number of cores

talex5:


talex5:
Looks like the benchmark machine is running an old version of Docker:

`Fatal error: exception Unix.Unix_error(Unix.ENOSYS, "pidfd_open", "")`

(looks like support was added in 2020: https://github.com/moby/moby/pull/41665, version 20.10.0)

PR #720, opened 2024-03-24
eio_linux: require Linux >= 5.15

talex5:
This allows removing a work-around that required checking whether every flow was a tty (and which was doing this on every read).

On my machine, this doesn't seem to have much effect on the benchmarks, but it does clean up the strace output.

### DIFF
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -154,16 +154,9 @@ module Flow = struct
 
   let close = Eio_unix.Fd.close
 
-  let is_tty t = Fd.use_exn "isatty" t Unix.isatty
-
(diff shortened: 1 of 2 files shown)


SGrondin:
The linked PR mentions it being fixed in Linux 5.14 so I'm curious why >=5.15?

talex5:
Two reasons: 5.15 is what the CI machines are running (otherwise we might prefer 6.1 as the minimum), and I'm not sure 5.14 added any new uring operations, so detecting it would be harder.

avsm:
I'm pretty uncomfortable with anything <6.6 for uring operations, so this one's fine by me :-)

SGrondin (approved):

PR #722, opened 2024-04-04
Add _BSD_SOURCE flag to fix eio_posix on OpenBSD.

prgbln:
eio_posix currently fails to build on OpenBSD 7.4 due to `O_NOFOLLOW` not being exposed:

```
#=== ERROR while compiling eio_posix.1.0 ======================================#
# context     2.0.10 | openbsd/x86_64 | ocaml-base-compiler.5.1.1 | https://opam.ocaml.org#a92d9dfb
# path        ~/.opam/5.1.1/.opam-switch/build/eio_posix.1.0
# command     ~/.opam/5.1.1/bin/dune build -p eio_posix -j 1 @install
# exit-code   1
# env-file    ~/.opam/log/eio_posix-34687-68bbf5.env
# output-file ~/.opam/log/eio_posix-34687-68bbf5.out
### output ###
# -> stderr:
# [...]
#  | /tmp/build_550fc6_dune/ocaml-configuratorf7473a/c-test-2/test.c:105:12: error: use of undeclared identifier 'O_NOFOLLOW'
#  | /tmp/build_550fc6_dune/ocaml-configuratorf7473a/c-test-2/test.c:105:12: error: use of undeclared identifier 'O_NOFOLLOW'
#  | /tmp/build_550fc6_dune/ocaml-configuratorf7473a/c-test-2/test.c:105:12: error: use of undeclared identifier 'O_NOFOLLOW'
#  | /tmp/build_550fc6_dune/ocaml-configuratorf7473a/c-test-2/test.c:105:12: error: use of undeclared identifier 'O_NOFOLLOW'
#  | /tmp/build_550fc6_dune/ocaml-configuratorf7473a/c-test-2/test.c:105:12: error: use of undeclared identifier 'O_NOFOLLOW'
#  | /tmp/build_550fc6_dune/ocaml-configuratorf7473a/c-test-2/test.c:105:12: error: use of undeclared identifier 'O_NOFOLLOW'
#  | /tmp/build_550fc6_dune/ocaml-configuratorf7473a/c-test-2/test.c:105:12: error: use of undeclared identifier 'O_NOFOLLOW'
#  | /tmp/build_550fc6_dune/ocaml-configuratorf7473a/c-test-2/test.c:105:12: error: use of undeclared identifier 'O_NOFOLLOW'
#  | fatal error: too many errors emitted, stopping now [-ferror-limit=]
#  | 20 errors generated.
# Error: failed to compile program
```
[...]

PR #723, opened 2024-04-04
eio_posix: fix filesystem tests on OpenBSD

jebrosen:
Looks like someone beat me to it on OpenBSD compilation (#722), but I did see some failures in the test suite and found a solution. This fix seems reasonable to me, but I definitely might have missed something - these two functions have differed on this check since they were first introduced in #694.

---

In the 'resolve' function there is a check for items being symlinks which was missing ELOOP as an indicator. This patch adds ELOOP, to match the similar check in 'open_beneath_fallback'.

This fixes filesystem tests on OpenBSD.


talex5 (approved):
Thanks!

PR #742, opened 2024-06-19
eio_windows: improve openat error handling

talex5:
@kentookura reported on #eio seeing this when doing `Eio.Path.load`:

> Fatal error: exception Unix.Unix_error(Unix.ENOENT, "openat", "")

This patch adds the failing path argument to the `Unix_error` and wraps it as an Eio exception.

Also, the `fs` example is now run by the Windows CI (though I don't expect it to show this problem as it doesn't use `open_dir`).


kentookura:
I should say that the error occurred without calling `open_dir`.

PR #747, opened 2024-08-07
Eio.Pool.use: move never_block argument

talex5:
Optional arguments should go first so that it's obvious to the compiler when they're not being used. Example of a program where this makes a difference:

```ocaml
let run use = use ignore

let () =
  let p = Eio.Pool.create 1 ignore in
  run (Eio.Pool.use p)
```


avsm (approved):
This looks sensible to me!

PR #748, opened 2024-08-18
Add batched copy for regular and block device files

koonwen:
The default behavior of `Flow.copy` in the uring backend doesn't leverage batching syscalls to be sent in one shot. In principle, files that are bounded in size can use this batched submission pattern to reduce context switching overhead. The batched copy implementation is taken directly from `lib_eio_linux/tests/eurcp.ml`. 

Old behaviour:
trace_old

```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ------------------
 33.44    0.000418          13        30           io_uring_enter
```

New behaviour:
trace_new

```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ------------------
  0.00    0.000000           0         7           io_uring_enter
```

Tests:
```json

#OLD
+default_fs: 288.79 MB/s
{
  "config": {
    "uname": "Linux debian-thinkpad 6.1.0-23-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.99-1 (2024-07-15) x86_64 GNU/Linux",
    "backend": "linux",
    "recommended_domain_count": 4
  },
  "results": [
    {
      "name": "Flow.copy",
      "metrics": [
        {
          "name": "default_fs",
          "value": 302819851.7110036,
          "units": "bytes/s",
          "description": "default_fs Flow.copy"
        }
      ]
    }
  ]
}

#NEW
+default_fs: 462.35 MB/s
{
  "config": {
    "uname": "Linux debian-thinkpad 6.1.0-23-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.99-1 (2024-07-15) x86_64 GNU/Linux",
    "backend": "linux",
    "recommended_domain_count": 4
  },
  "results": [
    {
      "name": "Flow.copy",
      "metrics": [
        {
          "name": "default_fs",
          "value": 484808364.3907238,
          "units": "bytes/s",
          "description": "default_fs Flow.copy"
        }
      ]
    }
  ]
}
```

### DIFF
--- a/bench/bench_copy.ml
+++ b/bench/bench_copy.ml
@@ -6,6 +6,22 @@ let chunk_size = 1 lsl 16
 let n_chunks = 10000
 let n_bytes = n_chunks * chunk_size
 
+let rec await pid =
+  match Unix.waitpid [] pid with
+  | (_pid, status) -> status
+  | exception Unix.Unix_error (Unix.EINTR, _, _) -> await pid
+
+let run prog args =
+  await (Unix.create_process prog (Array.of_list (prog :: args)) Unix.stdin Unix.stdout Unix.stderr)
+
+let check_status = function
+  | Unix.WEXITED 0 -> ()
+  | _ -> assert false
+
+let generate_file name =
+  run "fallocate" [ "-l"; string_of_int n_bytes; name ]
+  |> check_status
+
 let run_client sock =
   Fiber.both
     (fun () ->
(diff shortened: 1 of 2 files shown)


talex5 (commented):
Thanks for investigating this!

Doing this in `Flow.copy` might be a problem, because normally it copies bytes in order but this doesn't. For example, if I'm doing `tail -f dst` then I might miss some parts of the file. But having an `Eio.Path.copy` that worked this way would be reasonable.

Also, it shouldn't rely on fixed buffers as they're not always available (this is why the benchmarks are failing). I'm not convinced they're any faster either (Anil said they only make a difference for networking, and rarely even then, I think).

Issue #762, opened 2024-10-04
Improve handling of Windows paths


talex5:
In https://github.com/ocaml-multicore/eio/issues/124, @samwgoldman said:

> For Flow, we have the need to share paths between platforms. These are relative paths, using the `/` path separator. We only support Windows systems that transparently support the `/` path separator, so we can just pass these paths into Windows APIs without transforming them.

However, Eio also needs to support absolute paths on Windows and it seems this doesn't work (see https://discuss.ocaml.org/t/how-to-specify-a-full-windows-path-in-eio/14880). It would be good to investigate why.

@dra27 commented:
> I must confess that introducing yet another notation for a Windows path doesn’t look at all sensible to me! `/C:/path` is simply invalid, and if that’s ever printed (error messages, exceptions, etc.) it kinda just looks like Eio doesn’t know what it’s doing.

As mentioned in the original issue:

> The [Fpath](https://erratique.ch/software/fpath) library encodes lots of knowledge about Windows paths. It probably can't be used directly because its paths behave differently depending on the host platform (preventing e.g. a program running on Linux from creating paths that are to be used on Windows). The path rules should probably be per-filesystem instead.

I think we should give up on a uniform representation of paths and instead move some operations (e.g. `split`) to `Fs.Pi.DIR`, so that each file-system can provide its own implementation.

RichardChukwu:
Thanks for bringing up this issue, @talex5 

I agree that we need to improve how Eio handles Windows paths. Right now, using relative paths with the / separator works for some cases, but we have issues with absolute paths on Windows that need fixing.

**Things to Consider:**
1. Path Format: I understand @dra27's point about not wanting to add another way to represent Windows paths. We want to keep things clear, and a single format can help avoid confusion, especially in error messages.

2. Filesystem-Specific Solutions: It might be a good idea to let each filesystem handle its own path operations. By moving functions like splitting paths to Fs.Pi.DIR, we can keep the API clean and make it easier to support different systems.

3. Using the Fpath Library: We should look into the Fpath library, but we need to make sure it fits our needs and doesn’t create more issues with path handling across different platforms.

**Next Steps:**
Investigation: I’ll start looking into how paths are currently managed in Eio and identify what’s causing problems on Windows.

Community Input: I’d love to hear any ideas from others on how we can solve this.

I’m looking forward to working together to improve Windows path support in Eio!

samwgoldman:
[...]

Issue #761, opened 2024-10-04
Get main test-suite running on Windows


talex5:
Most of Eio's tests are in the `tests` directory, which isn't tested on Windows (due to the `(enabled_if (<> %{os_type} "Win32"))` in `tests/dune`).

A good first step would be to find out what happens if you remove that. Probably a few of the tests will fail (and should be disabled on Windows for now). Or possibly there are problems using MDX on Windows. Either way, it would be good to document the current situation.

Implementing https://github.com/realworldocaml/mdx/issues/393 would be really helpful for Eio.

patricoferris:
Indeed! I took a little look at the feasibility of this and MDX seems to be in good shape with the caveat that the MDX files use LF and not CRLF (I believe we can force Git to do something about that). Windows + VSCode users can also tell VSCode to use LF in the bottom right corner. CRLF files _hang_ in MDX... 

Unfortunately though, it looks like we actually need compiler changes to make this work. We run into https://github.com/ocaml/ocaml/issues/12999 because MDX is using a top-level environment IIUC. I spoke to @dra27 about this offline and a short-term solution is to make `caml_unix_error` a `CAMLextern` (although this may lead to other issues...) or to use `-fvisibility` which also means making compiler changes. If I find the time I'll look at both, but until then unfortunately I think we have to wait or continue with making tests non-bytecode for Windows :// 

We could also perhaps _remove_ uses of unix errors in our Windows stubs and use our own exception as a workaround perhaps... but that feels slightly annoying given this is just so we can test the library in MDX ! But maybe worth it... I don't understand the issue fully enough though as loading `eio_windows` in `ocaml` or `utop` works fine...

patricoferris:
To be clear though, working on the sections MDX work would be great and then we could disable all the tests _inside_ the MDX files so that come the fateful day that we're just turning off tests because of implementation issues in `eio_windows` (and not issues in ocaml/ocaml or elsewhere) we can do that easily by uncommenting a line or however it will support "turning off" parts of the markdown file for Windows!

RichardChukwu:
I'd like this to be assigned to me please @patricoferris @talex5 

Thank you

patricoferris:
Hey @RichardChukwu,

I think it would be best to focus on a single issue and it's PR (namely #763) at a time. Unfortunately, that does mean waiting for a review but I think that is only fair for other applicants. We'll do our best to review it in a timely manner. Thanks :))

RichardChukwu:
[...]

PR #763, opened 2024-10-04
Add stress fd test

### LINKED ISSUE #450: Add more stress tests and benchmarks
The only stress test we have is `stress_semaphore.ml` and the benchmarks mostly test the synchronisation primitives but not the OS integration (except for `bench_fd.ml`, which tests reading from `/dev/zero`). We need more.

In particular, we have found several resource leaks and races when e.g. spawning and joining domains in a tight loop (e.g. #421). The https://github.com/ocaml-multicore/retro-httpaf-bench/tree/master/cohttp-eio benchmark is useful; more like that (or simpler) would be good.

Testing for FD leaks would also be worth doing. e.g. checking that we have the same number of open FDs at the end of running the tests.

It would be very interesting to look at the benchmarks with e.g. https://magic-trace.org/ to see if there are obvious inefficiencies left.

The benchmarks could go in this repository (if simple) or we could create a new project just for that (if they have lots of dependencies).

Stress testing https://github.com/ocaml-multicore/lwt_eio would be useful too.

RichardChukwu:
### Description:
This pull request adds a new stress test (stress_fd.ml) to check for file descriptor (FD) leaks in Eio. The test ensures that the number of open FDs at the beginning of the test matches the count at the end, helping to identify resource leaks under high load.

### Test Details:
Test Objective: Validate that Eio properly handles file descriptors during concurrent domain and fiber operations without leaks.
Approach:

The test creates multiple domains and fibers, which perform various IO operations (e.g., reading/writing to a file).
At the end of the test, the number of open file descriptors is checked to ensure no leaks occurred.

Rationale: This addresses a common source of performance degradation and resource exhaustion when managing IO operations in parallel.

### Benchmark Improvements:
This test contributes to the ongoing effort to improve benchmarking and stress testing of Eio, particularly in areas related to OS integration, as discussed in issue #450.

Related Issues:
related to #450: Add more stress tests and benchmarks


RichardChukwu:
Hello @patricoferris I'd appreciate your feedback on this, thank you

patricoferris (changes requested):
Before reviewing the code, let's get the stress test building :)) 

The `dune project` file isn't doing anything. I think reading [the dune quickstart](https://dune.readthedocs.io/en/stable/quick-start.html) will help see the differences between dune-project and dune files.

Additionally, have a go at running some of the stress tests yourself, e.g. 

```
dune exec -- stress/stress_semaphore.exe
```

Then have a look to see how these are being built by dune, so you can add your test there too.

talex5:
This looks AI-generated to me. The code doesn't match the description. Also, https://github.com/ocaml-multicore/eio/issues/762#issuecomment-2394783534 looks AI-generated to me, though I'm not very familiar with this stuff.

@patricoferris do we need to add some rule somewhere about this? I see https://kernelnewbies.org/Outreachyfirstpatch says "Contributions that include content generated by AI tools that may rely on sources that are not compatible with GPLv2 thus cannot be accepted into the Linux kernel. This includes ChatGPT." Maybe we need something similar for Eio?

RichardChukwu:
Does it mean I can't work on this until issue #761 is resolved please? @patricoferris 

@talex5 I looked at the other stress tests and wrote the code in the same manner, please I'd like to know if that applies?

patricoferris:
> Does it mean I can't work on this until issue https://github.com/ocaml-multicore/eio/issues/761 is resolved please?

Yes, that's right. Let's keep our focus on a single issue and PR.

> Contributions that include content generated by AI tools that may rely on sources that are not compatible with GPLv2 thus cannot be accepted into the Linux kernel. This includes ChatGPT.

I think we should have a rule, though I'm not sure what it is. I'm not a lawyer, and from the _very_ light searching I did, it would seem that the licensing issue exists in a pretty grey area. However, that's not to say we can't have our own rule. I'll draft something for the CONTRIBUTING.md maybe? 

@RichardChukwu -- we really appreciate you taking the time to engage and contribute to Eio. To be clear, I don't think what we're saying here is that contributions _can't_ use AI. But it should be used as a tool to _help_ write a PR, not as a tool to write the whole PR, your comments and PR description for you. There are a few reasons for this. 

1. Most importantly, it obfuscates for us how you think. Without being able to see how you think, it makes it much harder for us to help you learn and subsequently to land the PR.
2. It is more work for us. This is related to (1), but reviewing code generated by AI at this point is inefficient and unproductive. We could have piped the issue into ChatGPT (or similar) ourselves, but the point of Outreachy is to help _you_ contribute to open-source.

So I think we can maybe try this PR again, with your thoughts, descriptions, code and maybe even bugs :)) ? In which case, I'm going to close this PR. Looking forward to seeing a new one when you have the time.

RichardChukwu:
Duly noted @patricoferris thank you so much for this feedback, I understand you perfectly and will take corrections accordingly.

Let me figure out another issue to work on then. How about assigning #761 to me first, since that's needed for this isuse to be worked on and is more related to Windows?

PR #765, opened 2024-10-06
Add advice about using AI for code generation

patricoferris:
cc: @talex5 and @avsm 

This is based on #763. Happy for it to say something different, but thought I would get the ball rolling here.


avsm:
LGTM also; we can revise this in the future but good to be conservative now from a licensing perspective.

PR #766, opened 2024-10-11
Implement symlink logic for Windows in test_symlink

create2000:
This PR updates tests for improved compatibility across Windows and Unix-based systems. Specifically, it addresses the differences in symlink permissions and behaviors between these systems, and modifies the test_symlink function to handle these differences gracefully.

Key Changes

1. Introduced symlink_safe, that attempts to create symlinks with ~to_dir flags for directories. This function catches permission errors and skips symlink creation if necessary, particularly on systems that restrict symlink permissions (e.g., Windows).
2. Updated the test_symlink logic to check for the operating system and conditionally create symlinks. 

Testing: Tests were conducted on both Windows and Unix-like environments (WSL) to validate the compatibility improvements. Special attention was given to:

1. Permission Errors: Verified that permission-related errors are gracefully handled without disrupting the test flow.
2. Directory Existence and Structure: Confirmed that symlink paths and directories are correctly created or handled based on the system's symlink behavior.

### DIFF
--- a/lib_eio_windows/test/dune
+++ b/lib_eio_windows/test/dune
@@ -1,5 +1,7 @@
 (test
-  (name test)
+  (name test)  ; Main test suite
+  (libraries alcotest kcas eio.mock eio_windows)
   (package eio_windows)
-  (build_if (= %{os_type} "Win32"))
-  (libraries alcotest kcas eio.mock eio_windows))
+  (build_if (= %{os_type} "Win32")))
+
+
--- a/lib_eio_windows/test/test_fs.ml
+++ b/lib_eio_windows/test/test_fs.ml
@@ -23,11 +23,6 @@ let try_mkdir path =
   | () -> traceln "mkdir %a -> ok" Path.pp path
   | exception ex -> raise ex
 
-let try_mkdirs ?exists_ok path =
-  match Path.mkdirs ?exists_ok path ~perm:0o700 with
-  | () -> traceln "mkdirs %a -> ok" Path.pp path
-  | exception ex -> traceln "@[<h>%a@]" Eio.Exn.pp ex
-
 let try_rename p1 p2 =
   match Path.rename p1 p2 with
   | () -> traceln "rename %a to %a -> ok" Path.pp p1 Path.pp p2
@@ -80,7 +75,7 @@ let test_exclusive env () =
   Eio.traceln "fiest";
   Path.save ~create:(`Exclusive 0o666) path "first-write";
   Eio.traceln "next";
-  try
+  try 
     Path.save ~create:(`Exclusive 0o666) path "first-write";
     Eio.traceln "nope";
     failwith "Should have failed"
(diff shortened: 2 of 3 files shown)


lib_eio_windows/test/dune
```diff
@@ -1,5 +1,7 @@
 (test
-  (name test)
+  (name test)  ; Main test suite
+  (libraries alcotest kcas eio.mock eio_windows)
```
patricoferris, line 3: These changes are not necessary (I don't think the comment is too important and shifting the order only adds extra things to review).

lib_eio_windows/test/test_fs.ml
```diff
@@ -130,69 +125,68 @@ let test_mkdir env () =
   Unix.rmdir "subdir\\nested";
   Unix.rmdir "subdir"
 
-let test_mkdirs env () =
```
patricoferris, line 133: Looks like we might have forgotten to add this test to the test-suite. Instead of removing it can we add it at the bottom?

lib_eio_windows/test/test_fs.ml
```diff
@@ -130,69 +125,68 @@ let test_mkdir env () =
...
+    let symlink_safe ~to_dir src dst = 
+      try
+        Unix.symlink ~to_dir src dst
+      with
+      Unix.Unix_error (Unix.EPERM, _, _) ->
+        Printf.printf "User lacks permission to create symlinks. Skipping symlink creation for %s -> %s.\n" src dst
+        | e -> raise e
+      in
+
+  if Sys.os_type = "Win32" then
```
patricoferris, line 152: What I had in mind was to simply _skip_ the whole test if symlinking wasn't supported. Something along the lines of 

```ocaml
if not Unix.has_symlink then () else begin
 (* The rest of the test *)
end
```

patricoferris (changes requested):
Hey @create2000,

Thanks for the PR. This wasn't quite what I had in mind, I've outlined that a bit more in the comment on the `test_symlink` test itself.

Also, in general, when contributing try to keep each PR focused on the single issue it is trying to fix. If there are other bits and pieces you want to make changes to you can either in a separate PR or at the very least in a separate commit. There's a lot of code movement here that is unrelated to bug being fixed.

lib_eio_windows/test/test_fs.ml
create2000: I did something of this nature but thought otherwise. I'd rework on it. Thank you

lib_eio_windows/test/test_fs.ml
create2000: Alright. I'll fix at the bottom and see how it goes

lib_eio_windows/test/dune
create2000: Sorry about the addition of extra things to review. Would remove it.

create2000:
So I am thinking, initially i used the ```Sys.os_type ``` to check if the current system is windows and if it is, it should skip the whole test. I am trying to know which would be a better approach between that and using this: 

```
if not Unix.has_symlink then () else begin
```

patricoferris:
I think `has_symlink` is what we want. We still want to test on Windows when this is available :))

create2000:
I have updated the code

patricoferris:
Thanks @create2000 -- this looks better, but the git history now is a little confusing given you've made and undone changes. If you could reset back to the start and add just the changes for the symlink test that would be very much appreciated :))

create2000:
@patricoferris  I deleted this branch and have sent a new PR to ensure that my git history is not confusing. Hope this doesn't add more work. Thank you

Issue #750, opened 2024-08-26
Relax the return types of Eio.Process.pipe


rizo:
## Motivation
Currently the `Eio.Process.pipe` function has the following type:

```ocaml
val pipe : sw:Switch.t -> _ mgr ->
  [Flow.source_ty | Resource.close_ty] Std.r *
  [Flow.sink_ty | Resource.close_ty] Std.r
```

This means that the created sources and sinks are required to be "closable". This restriction currently prevents compatibility with types that are monomorphically just sinks or just sources (_i.e._, without the `` `Close `` tag).

My suggestion is to change the type of `pipe` to:

```ocaml
val pipe : sw:Switch.t -> _ mgr ->
  [< Flow.source_ty | Resource.close_ty] Std.r *
  [< Flow.sink_ty | Resource.close_ty] Std.r
```

This would allow the resulting sources and sinks to be implicitly used as any of the allowed subtypes.

## Examples

To elaborate on why I think this is useful, consider the two examples below.

### This works: flexible source consumers

```ocaml
let src, _snk = Eio.Process.pipe ~sw proc_mgr in
Eio.Flow.read_all src
```

If we expand the `read_all` type, it accepts `` [> `R] `` flows, which of course type-checks because any additional variants, like `` `Close `` are allowed. Thus, we can see that `read_all` is a "good" consumer and does not limit our types.

### This does not work: fixed flow types

Some APIs use monomorphic versions of source and sink types. For example, the [Cohttp_eio.Body.t](https://github.com/mirage/ocaml-cohttp/blob/642bc8250826535c88487edabf0a7253a0b77230/cohttp-eio/src/body.ml#L1) type is defined as:

```ocaml
type t = Eio.Flow.source_ty Eio.Resource.t
```

Which means that the following example does not work:

```ocaml
let src, _snk = Eio.Process.pipe ~sw proc_mgr in
Cohttp_eio.Server.respond ~status:`OK ~body:src
```

Failing with:
```
25 |   Cohttp_eio.Server.respond ~status:`OK ~body:src
                                                   ^^^
Error: This expression has type [ `Close | `Flow | `R ] Eio.Resource.t
       but an expression was expected of type
         Cohttp_eio__.Body.t = Eio.Flow.source_ty Eio.Resource.t
       Type [ `Close | `Flow | `R ] is not compatible with type
         Eio.Flow.source_ty = [ `Flow | `R ]
       The second variant type does not allow tag(s) `Close
```

[...]

PR #767, opened 2024-10-14
Relax return types of Eio.Process.pipe

Arogbonlo:
This updates the `pipe` function in the `Make_mgr` module to use polymorphic variants for its return types. The change allows the returned source and sink to be compatible with APIs expecting monomorphic types.
**Change:**
- Modified `pipe` function signature in `Make_mgr`
- Updated from `[...]` to `[>...]` for both source and sink types

This addresses issue #750, improving flexibility when working with various APIs that use stricter type definitions for sources and sinks.
- Added new test case demonstrating compatibility with constrained types

### DIFF
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -64,7 +64,7 @@ module Pi = struct
     type t
 
     val pipe :
-      t ->
+      'a t ->
       sw:Switch.t ->
       ([> Flow.source_ty | Resource.close_ty] r * [> Flow.sink_ty | Resource.close_ty] r)
 
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -67,7 +67,7 @@ module Pi = struct
       t ->
       sw:Switch.t ->
       ([> Flow.source_ty | Resource.close_ty] r * [> Flow.sink_ty | Resource.close_ty] r)
-      
+
     val spawn :
       t ->
       sw:Switch.t ->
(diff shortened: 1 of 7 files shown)


patricoferris:
Hey @Arogbonlo,

Thanks for the PR. There are a few issues here that need addressed:

 - The original issue was to make the return type of `pipe` allow for _all possible subtypes_ (`<`) not to have _at least_ the variants it describes (`>`).
 - This is a type change and doesn't really require a test.

I think perhaps you should start again updating the types in `lib_eio/process.{ml.mli}` and following this through to `eio_unix`. I think you are also going to need to update the `Pipe` effect in `lib_eio/unix/private.ml` too. No other parts of the types should be changed (e.g. you changed `_ mgr` to  `'a t`)

Arogbonlo:
I'll effect them immediately. Thank you.
@patricoferris

patricoferris:
Hey @Arogbonlo how's it going? Do you need any help?

Arogbonlo:
> Hey @Arogbonlo how's it going? Do you need any help?

Thank you so much for the check!

I'll push my progress in a few hours, so you can check it and let me know where i have to make changes.

Arogbonlo:
Hey @patricoferris 

I just made a commit, please I'll appreciate your feedback on it.

Thank you.

patricoferris (changes requested):
The changes are not building, see the [CI pipeline](https://github.com/ocaml-multicore/eio/pull/767/checks?check_run_id=31644899324).

Did you try running this locally? Also, it is good practice to use good commit messages instead of ".". If you need to redo something, you can always `git reset` and `git push -f` to your own branch.

PR #768, opened 2024-10-14
fix: test_symlink

create2000:
Worked on disabling the windows symlink.

### DIFF
--- a/lib_eio_windows/test/test_fs.ml
+++ b/lib_eio_windows/test/test_fs.ml
@@ -114,6 +114,16 @@ let test_append env () =
   Path.save ~create:`Never ~append:true test_file "2nd-write";
   Alcotest.(check string) "append" "1st-write-original2nd-write" (Path.load test_file)
 
+  let test_mkdir env () =
+    let cwd = Eio.Stdenv.cwd env in
+    try_mkdir (cwd / "subdir");
+    try_mkdir (cwd / "subdir\\nested");
+    let test_file = cwd / "subdir\\nested\\test-file" in
+    Path.save ~create:(`Exclusive 0o600) test_file "data";
+    Alcotest.(check string) "mkdir" "data" (Path.load test_file);
+    Unix.unlink "subdir\\nested\\test-file";
+    Unix.rmdir "subdir\\nested";
+    Unix.rmdir "subdir"
 
 let test_symlink env () =
   (* 
@@ -218,17 +228,6 @@ let try_failing_unlink env () =
   in
   ()
 
-  let test_mkdir env () =
-    let cwd = Eio.Stdenv.cwd env in
-    try_mkdir (cwd / "subdir");
-    try_mkdir (cwd / "subdir\\nested");
-    let test_file = cwd / "subdir\\nested\\test-file" in
-    Path.save ~create:(`Exclusive 0o600) test_file "data";
-    Alcotest.(check string) "mkdir" "data" (Path.load test_file);
-    Unix.unlink "subdir\\nested\\test-file";
-    Unix.rmdir "subdir\\nested";
-    Unix.rmdir "subdir"
-
 let test_remove_dir env () =
   let cwd = Eio.Stdenv.cwd env in
   try_mkdir (cwd / "d1");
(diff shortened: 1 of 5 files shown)


lib_eio_windows/test/test_fs.ml
```diff
@@ -119,80 +114,61 @@ let test_append env () =
...
-  try_mkdir (cwd / "subdir");
-  try_mkdir (cwd / "subdir\\nested");
-  let test_file = cwd / "subdir\\nested\\test-file" in
-  Path.save ~create:(`Exclusive 0o600) test_file "data";
-  Alcotest.(check string) "mkdir" "data" (Path.load test_file);
-  Unix.unlink "subdir\\nested\\test-file";
-  Unix.rmdir "subdir\\nested";
-  Unix.rmdir "subdir"
-
-let test_mkdirs env () =
```
patricoferris, line 133: As mentioned in the previous PR, could we add this test to the bottom of the file alongside the other tests instead of removing it. Thanks!

lib_eio_windows/test/test_fs.ml
```diff
@@ -119,80 +114,61 @@ let test_append env () =
   Path.save ~create:`Never ~append:true test_file "2nd-write";
   Alcotest.(check string) "append" "1st-write-original2nd-write" (Path.load test_file)
 
-let test_mkdir env () =
```
patricoferris, line 122: Why did you move `test_mkdir` from here to further down the file? It is an extra diff that doesn't seem to be necessary, but perhaps I am mistaken?

lib_eio_windows/test/test_fs.ml
```diff
@@ -119,80 +114,61 @@ let test_append env () =
...
-  Unix.symlink ~to_dir:true "foo" "sandbox\\dangle";
-  try_mkdir (cwd / "tmp");
-  Eio.Path.with_open_dir (cwd / "sandbox") @@ fun sandbox ->
-  try_mkdir (sandbox / "subdir");
-  try_mkdir (sandbox / "to-subdir\\nested");
-  let () =
-    try
+
+  if not (Unix.has_symlink ()) then
+    Printf.printf "Skipping test_symlink on systems that don't support symlinks.\n"
```
patricoferris, line 135: This looks good to me, thank you!

patricoferris (changes requested):
The core fix is good, but we need to not also make other changes whilst doing it :))

lib_eio_windows/test/test_fs.ml
create2000: No, i think it must have been an oversight from me.

lib_eio_windows/test/test_fs.ml
create2000: Alright. Thanks.

create2000:
@patricoferris -- I have done the fixes. Is this okay?

patricoferris:
Definitely getting there @create2000.

I would recommend resetting this history and only (a) making the fix to `test_symlink` and [adding the test_mkdirs into the tests of this file](https://github.com/ocaml-multicore/eio/blob/216ccff845d68ada8c7d3bd31581cbf37e8ee791/lib_eio_windows/test/test_fs.ml#L268-L279).

create2000:
Thank you @patricoferris -- I have added the ```test_mkdirs``` to the tests. About making the fix to ```test_symlink``` I understand it to mean resetting my git history to reflect `fix: test_symlink`. If this is it, I have done that.

patricoferris:
Thanks @create2000 :))

> mean resetting my git history to reflect fix: test_symlink

My point here is that right now, the [diff](https://github.com/ocaml-multicore/eio/pull/768/files) is a little confusing. We have four distinct changes happening:
 1. The fix for the symlink test
 2. Adding `test_mkdirs` to the test suite
 3. White-spacing/formating changes
 4. The actual `test_mkdirs` function being moved from the top of the file to the bottom
 
 Ideally, this would only be changes (1) and (2). I'm okay with (3) slipping into this PR, especially since the codebase has no automatic formatting. However, (4) seems to be a change that is a resulting of first deleting that function and then rewriting it somewhere else. It would be good if the history wasn't like that.
 
 So concretely, I think we want to redo this PR with just (1), (2) and (3). Does that make sense?

create2000:
Thank you @patricoferris . I understand now what you mean. I will close this PR and [here](https://github.com/ocaml-multicore/eio/pull/771) is the new PR created to keep everything clean.  I hope the new PR meets the corrections. 

Also, In recording my contribution on the Outreachy contribution page, I am asked to provide a URL of my contributions if it is merged or accepted. As this is still ongoing, can I still submit a URL in the hope that it will get merged before the contribution deadline? This is to keep with the advice of Outreachy to record our contributions early.

PR #769, opened 2024-10-17
Make fork_action.h a public_header

patricoferris:
For other libraries that depend on Eio and wish to provide new fork actions, it would be useful to install the `fork_action.h` header.


talex5 (approved):
Thanks!

PR #770, opened 2024-10-18
Relax return types of Eio.Process.Pipe

Arogbonlo:
This PR relaxes the return types of the pipe function in Eio.Process to make the code more flexible. The main goal was to allow the pipe function to handle a broader range of types for input and output flows, while still maintaining compatibility with the rest of the codebase.

**Changes:**

- **Relaxed Type Definitions**: I’ve updated the return types for pipe so they can accept any subtype of `Flow.source_ty` and `Flow.sink_ty`. This will give us more flexibility when working with processes that require different types of flows.

- **Polymorphic Variants**: I used polymorphic variants (`[< .. ]`) to allow for more flexible types, ensuring that the code still works with the broader variants expected in other parts of the system.

**Testing:**
I ran dune runtest and confirmed that all tests are passing with these changes. The new types work fine, and the tests ensure everything is still functioning as expected.

> **Relaxing the types for pipe allows us to work with a wider variety of process setups and flows, without being too restrictive about the types we return. It also makes the code easier to extend in the future.**

This addresses issue https://github.com/ocaml-multicore/eio/issues/750, improving flexibility when working with various APIs that use stricter type definitions for sources and sinks.

### DIFF
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -64,7 +64,7 @@ module Pi = struct
     type t
 
     val pipe :
-      t ->
+      'a t ->
       sw:Switch.t ->
       ([> Flow.source_ty | Resource.close_ty] r * [> Flow.sink_ty | Resource.close_ty] r)
 
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -147,7 +147,6 @@ let pipe (type tag) ~sw ((Resource.T (v, ops)) : [> tag mgr_ty] r) =
   let r = (r : [Flow.source_ty | Resource.close_ty] r :> [< Flow.source_ty | Resource.close_ty] r) in
   let w = (w : [Flow.sink_ty   | Resource.close_ty] r :> [< Flow.sink_ty   | Resource.close_ty] r) in
   r, w
-  
 let parse_out (type tag) (t : [> tag mgr_ty] r) parse ?cwd ?stdin ?stderr ?is_success ?env ?executable args =
   Switch.run ~name:"Process.parse_out" @@ fun sw ->
   let r, w = pipe t ~sw in
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -61,12 +61,12 @@ module Pi = struct
 
   module type MGR = sig
     type tag
-    type t
+    type t 
 
     val pipe :
       t ->
       sw:Switch.t ->
-      [< `Flow_source | `Resource_close] r * [< `Flow_sink | `Resource_close] r
+      [< Flow.source_ty | Resource.close_ty] r * [< Flow.sink_ty | Resource.close_ty] r
 
     val spawn :
       t ->
@@ -141,7 +141,8 @@ let run t ?cwd ?stdin ?stdout ?stderr ?(is_success = Int.equal 0) ?env ?executab
     let ex = err (Child_error status) in
     raise (Exn.add_context ex "running command: %a" pp_args args)
 
-let pipe (type tag) ~sw ((Resource.T (v, ops)) : [> tag mgr_ty] r) =
+let pipe (type tag) ~sw ((Resource.T (v, ops)) : [> tag mgr_ty] r) :
+  ([< Flow.source_ty | Resource.close_ty ] r * [< Flow.sink_ty | Resource.close_ty ] r) =
   let module X = (val (Resource.get ops Pi.Mgr)) in
   X.pipe v ~sw
 
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -66,7 +66,7 @@ module Pi = struct
     val pipe :
       t ->
       sw:Switch.t ->
-      [< Flow.source_ty | Resource.close_ty] r * [< Flow.sink_ty | Resource.close_ty] r
+      [Flow.source_ty | Resource.close_ty] r * [Flow.sink_ty | Resource.close_ty] r
 
     val spawn :
       t ->
@@ -147,7 +147,7 @@ let pipe (type tag) ~sw ((Resource.T (v, ops)) : [> tag mgr_ty] r) =
   let r = (r : [Flow.source_ty | Resource.close_ty] r :> [< Flow.source_ty | Resource.close_ty] r) in
   let w = (w : [Flow.sink_ty   | Resource.close_ty] r :> [< Flow.sink_ty   | Resource.close_ty] r) in
   r, w
-
+  
 let parse_out (type tag) (t : [> tag mgr_ty] r) parse ?cwd ?stdin ?stderr ?is_success ?env ?executable args =
   Switch.run ~name:"Process.parse_out" @@ fun sw ->
   let r, w = pipe t ~sw in
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -61,7 +61,7 @@ module Pi = struct
 
   module type MGR = sig
     type tag
-    type t 
+    type t
 
     val pipe :
       t ->
@@ -141,10 +141,12 @@ let run t ?cwd ?stdin ?stdout ?stderr ?(is_success = Int.equal 0) ?env ?executab
     let ex = err (Child_error status) in
     raise (Exn.add_context ex "running command: %a" pp_args args)
 
-let pipe (type tag) ~sw ((Resource.T (v, ops)) : [> tag mgr_ty] r) :
-  ([< Flow.source_ty | Resource.close_ty ] r * [< Flow.sink_ty | Resource.close_ty ] r) =
+let pipe (type tag) ~sw ((Resource.T (v, ops)) : [> tag mgr_ty] r) =
   let module X = (val (Resource.get ops Pi.Mgr)) in
-  X.pipe v ~sw
+  let r, w = X.pipe v ~sw in
+  let r = (r : [Flow.source_ty | Resource.close_ty] r :> [< Flow.source_ty | Resource.close_ty] r) in
+  let w = (w : [Flow.sink_ty   | Resource.close_ty] r :> [< Flow.sink_ty   | Resource.close_ty] r) in
+  r, w
 
 let parse_out (type tag) (t : [> tag mgr_ty] r) parse ?cwd ?stdin ?stderr ?is_success ?env ?executable args =
   Switch.run ~name:"Process.parse_out" @@ fun sw ->
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -67,7 +67,7 @@ module Pi = struct
       t ->
       sw:Switch.t ->
       ([> Flow.source_ty | Resource.close_ty] r * [> Flow.sink_ty | Resource.close_ty] r)
-      
+
     val spawn :
       t ->
       sw:Switch.t ->
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -64,9 +64,9 @@ module Pi = struct
     type t
 
     val pipe :
-      'a t ->
+      t ->
       sw:Switch.t ->
-      ([> Flow.source_ty | Resource.close_ty] r * [> Flow.sink_ty | Resource.close_ty] r)
+      ([< Flow.source_ty | Resource.close_ty] r) * ([< Flow.sink_ty | Resource.close_ty] r)
 
     val spawn :
       t ->
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -63,11 +63,11 @@ module Pi = struct
     type tag
     type t
 
-  
+    val pipe :
       t ->
       sw:Switch.t ->
-      [< Flow.source_ty | Resource.close_ty] r * [< Flow.sink_ty | Resource.close_ty] r
-
+      ([> Flow.source_ty | Resource.close_ty] r * [> Flow.sink_ty | Resource.close_ty] r)
+      
     val spawn :
       t ->
       sw:Switch.t ->
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -63,10 +63,10 @@ module Pi = struct
     type tag
     type t
 
-    val pipe :
+  
       t ->
       sw:Switch.t ->
-      [Flow.source_ty | Resource.close_ty] r * [Flow.sink_ty | Resource.close_ty] r
+      [< Flow.source_ty | Resource.close_ty] r * [< Flow.sink_ty | Resource.close_ty] r
 
     val spawn :
       t ->
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -66,7 +66,7 @@ module Pi = struct
     val pipe :
       t ->
       sw:Switch.t ->
-      ([< Flow.source_ty | Resource.close_ty] r) * ([< Flow.sink_ty | Resource.close_ty] r)
+      [< `Flow_source | `Resource_close] r * [< `Flow_sink | `Resource_close] r
 
     val spawn :
       t ->
--- a/lib_eio/process.mli
+++ b/lib_eio/process.mli
@@ -140,7 +140,11 @@ val parse_out :
 
 (** {2 Pipes} *)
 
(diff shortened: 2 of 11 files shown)


Arogbonlo:
Hey @patricoferris. Please look at this and let me know if there are lapses. 

Thank you.

lib_eio_linux/eio_linux.ml
```diff
@@ -537,8 +537,8 @@ let run_event_loop (type a) ?fallback config (main : _ -> a) arg : a =
       | Eio_unix.Private.Pipe sw -> Some (fun k ->
           match
             let r, w = Low_level.pipe ~sw in
-            let r = (Flow.of_fd r :> _ Eio_unix.source) in
-            let w = (Flow.of_fd w :> _ Eio_unix.sink) in
+            let r = (Flow.of_fd r :> [< Eio_unix.source_ty] Eio.Std.r) in
```
patricoferris, line 540: I actually think we can remove the explicit conversions here no?

```suggestion
            let r = Flow.of_fd r in
```

lib_eio_linux/eio_linux.ml
```diff
@@ -537,8 +537,8 @@ let run_event_loop (type a) ?fallback config (main : _ -> a) arg : a =
...
+            let w = (Flow.of_fd w :> [< Eio_unix.sink_ty] Eio.Std.r) in
```
patricoferris, line 541:
```suggestion
            let w = Flow.of_fd w in
```

lib_eio/unix/process.ml
```diff
@@ -118,8 +118,8 @@ end) = struct
   type tag = [ `Generic | `Unix ]
 
   let pipe _ ~sw =
-    (Private.pipe sw :> ([Eio.Resource.close_ty | Eio.Flow.source_ty] r *
-    [Eio.Resource.close_ty | Eio.Flow.sink_ty] r))
+    (Private.pipe sw :> ([< Eio.Resource.close_ty | Eio.Flow.source_ty] r *
```
patricoferris, line 121: I don't think we need these coercions here either.

patricoferris (commented):
This is looking great, I think there some coercions that are now unnecessary though.

lib_eio_linux/eio_linux.ml
Arogbonlo: Thank you for this review!
For the explicit conversions, I'd suggest we use it since we're dealing with cases where polymorphism is needed. 
And also, I just generalized the values of `r `and `w`.

I think the base benefit of removing it is just that it makes the code cleaner.

lib_eio/unix/process.ml
Arogbonlo: yes yes.
this is going to be better right?
```
let pipe _ ~sw =
    let source, sink = Private.pipe sw in
    (source, sink)
```

lib_eio_linux/eio_linux.ml
Arogbonlo: Though I could replace it with these

```
let r = Flow.of_fd r in
let w = Flow.of_fd w in
```

Arogbonlo:
Hey @patricoferris 
I've made modifications. I'd appreciate your looking into it.

Thank you.

lib_eio/process.ml
```diff
@@ -61,12 +61,12 @@ module Pi = struct
 
   module type MGR = sig
     type tag
-    type t
+    type t
```
talex5, line 64: @Arogbonlo please look at your PR yourself before submitting it for review. This change is obviously not supposed to be here.

lib_eio/process.ml
```diff
@@ -141,7 +141,8 @@ let run t ?cwd ?stdin ?stdout ?stderr ?(is_success = Int.equal 0) ?env ?executab
     let ex = err (Child_error status) in
     raise (Exn.add_context ex "running command: %a" pp_args args)
 
-let pipe (type tag) ~sw ((Resource.T (v, ops)) : [> tag mgr_ty] r) =
+let pipe (type tag) ~sw ((Resource.T (v, ops)) : [> tag mgr_ty] r) :
+  ([< Flow.source_ty | Resource.close_ty ] r * [< Flow.sink_ty | Resource.close_ty ] r) =
```
talex5, line 145: What is this change for? Did you mean to include it?

I suspect what you really want here is:
```ocaml
let pipe (type tag) ~sw ((Resource.T (v, ops)) : [> tag mgr_ty] r) =
  let module X = (val (Resource.get ops Pi.Mgr)) in
  let r, w = X.pipe v ~sw in
  let r = (r : [Flow.source_ty | Resource.close_ty] r :> [< Flow.source_ty | Resource.close_ty] r) in
  let w = (w : [Flow.sink_ty   | Resource.close_ty] r :> [< Flow.sink_ty   | Resource.close_ty] r) in
  r, w
```

Then most of the other changes can go away.

(it's annoying OCaml requires this to be written out in full; you'd think it could do the cast automatically)

lib_eio/process.mli
```diff
@@ -171,7 +174,7 @@ module Pi : sig
     val pipe :
       t ->
       sw:Switch.t ->
-      [Flow.source_ty | Resource.close_ty] r * [Flow.sink_ty | Resource.close_ty] r
+      [< Flow.source_ty | Resource.close_ty] r * [< Flow.sink_ty | Resource.close_ty] r
```
talex5, line 177: Or, you could remove this change and do the cast in `pipe`. That avoids the need for every implementation of MGR to do the cast, which is an API change.

lib_eio/unix/eio_unix.mli
```diff
@@ -97,7 +97,7 @@ module Private : sig
     | Await_readable : Unix.file_descr -> unit Effect.t      (** See {!await_readable} *)
     | Await_writable : Unix.file_descr -> unit Effect.t      (** See {!await_writable} *)
     | Get_monotonic_clock : Eio.Time.Mono.ty r Effect.t
-    | Pipe : Eio.Switch.t -> (source_ty r * sink_ty r) Effect.t    (** See {!pipe} *)
+    | Pipe : Eio.Switch.t -> ([< source_ty] r * [< sink_ty] r) Effect.t
```
talex5, line 100: Why did you remove the comment? Did you intend to do that?

lib_eio/unix/private.ml
```diff
@@ -7,7 +7,7 @@ type _ Effect.t +=
   | Await_readable : Unix.file_descr -> unit Effect.t
   | Await_writable : Unix.file_descr -> unit Effect.t
   | Get_monotonic_clock : Eio.Time.Mono.ty r Effect.t
-  | Pipe : Switch.t -> (source_ty r * sink_ty r) Effect.t
+  | Pipe : Switch.t -> ([< source_ty] r * [< sink_ty] r) Effect.t
```
talex5, line 10: You shouldn't need this. Just cast the result.

talex5 (changes requested):
Note: you don't need to close the PR every time you make a change. Just force-push the new version to your branch (`git push -f`).

lib_eio/process.mli
Arogbonlo: wow. this is a brilliant one! 
Thank you so much for this.

lib_eio/unix/private.ml
Arogbonlo: If I'm using 
```
let pipe (type tag) ~sw ((Resource.T (v, ops)) : [> tag mgr_ty] r) =
  let module X = (val (Resource.get ops Pi.Mgr)) in
  let r, w = X.pipe v ~sw in
  let r = (r : [Flow.source_ty | Resource.close_ty] r :> [< Flow.source_ty | Resource.close_ty] r) in
  let w = (w : [Flow.sink_ty   | Resource.close_ty] r :> [< Flow.sink_ty   | Resource.close_ty] r) in
  r, w
``` , then i dont think I'll have to cast the result in a pipe. Right?

**Comment by Arogbonlo** [comment]
hello @talex5 @patricoferris.
Just made the corrections you highlighted. I'd appreciate if you take a look at it and let me know what you think.

thank you.

**Comment by Arogbonlo** [comment]
Hello @talex5 @patricoferris ,

I hope you're having a good day.
Just a gentle reminder to have a check on this. Thank you for understanding.

File: lib_eio/unix/private.ml [hunk 9]
```diff
@@ -7,12 +7,14 @@ type _ Effect.t +=
...
```
patricoferris, line 10: Seems this is still not resolved @Arogbonlo ?

lib_eio/unix/eio_unix.mli
```diff
@@ -97,7 +97,7 @@ module Private : sig
     | Await_readable : Unix.file_descr -> unit Effect.t      (** See {!await_readable} *)
     | Await_writable : Unix.file_descr -> unit Effect.t      (** See {!await_writable} *)
     | Get_monotonic_clock : Eio.Time.Mono.ty r Effect.t
-    | Pipe : Eio.Switch.t -> (source_ty r * sink_ty r) Effect.t    (** See {!pipe} *)
+    | Pipe : Eio.Switch.t -> ([< source_ty] r * [< sink_ty] r) Effect.t   (** See {!pipe} *)
```
patricoferris, line 100: This can also change back.

lib_eio/process.mli
patricoferris, line 177: As @talex5 points out this can also be fixed @Arogbonlo.

patricoferris (changes requested):
There are still outstanding changes from @talex5's review

Arogbonlo:
> There are still outstanding changes from @talex5's review

I had options of either casting in pipes or writing the cast in full as below: 
```
let pipe (type tag) ~sw ((Resource.T (v, ops)) : [> tag mgr_ty] r) =
  let module X = (val (Resource.get ops Pi.Mgr)) in
  let r, w = X.pipe v ~sw in
  let r = (r : [Flow.source_ty | Resource.close_ty] r :> [< Flow.source_ty | Resource.close_ty] r) in
  let w = (w : [Flow.sink_ty   | Resource.close_ty] r :> [< Flow.sink_ty   | Resource.close_ty] r) in
  r, w
```
of which I chose to write in full. Or do you think otherwise about this my choice?

patricoferris:
Yes the cast is now in `val pipe` of the Process interface. This means all the other places can go back to how they were I think. The problem is we would like to reduce the API changes being made (some of my original comments were wrong w.r.t to this). 

Take a look at this commit 655bfed5d3047fa0dfcdbdf652384bb69c3061ae. These changes revert some of the changes already made, so it might be good to reset back to the start and just add the casts in the pipe leaving most of the code unchanged and force pushing here. Does that make sense ?

Arogbonlo:
> Yes the cast is now in `val pipe` of the Process interface. This means all the other places can go back to how they were I think. The problem is we would like to reduce the API changes being made (some of my original comments were wrong w.r.t to this).
> 
> Take a look at this commit [655bfed](https://github.com/ocaml-multicore/eio/commit/655bfed5d3047fa0dfcdbdf652384bb69c3061ae). These changes revert some of the changes already made, so it might be good to reset back to the start and just add the casts in the pipe leaving most of the code unchanged and force pushing here. Does that make sense ?

oh yeah, I get you. I'll make those corrections right away.
Thank you.

Arogbonlo:
hello @talex5 @patricoferris.

I just made the outstanding corrections you highlighted. I'd appreciate it if you take a look at it and let me know what you think...

Thank you.

Arogbonlo:
Hello @talex5 @patricoferris ,

Just a gentle reminder to have a check on this. Thank you.

lib_eio/unix/private.ml
```diff
@@ -7,13 +7,13 @@ type _ Effect.t +=
   | Await_readable : Unix.file_descr -> unit Effect.t
   | Await_writable : Unix.file_descr -> unit Effect.t
   | Get_monotonic_clock : Eio.Time.Mono.ty r Effect.t
-  | Pipe : Switch.t -> (source_ty r * sink_ty r) Effect.t
+  | Pipe : Switch.t -> ( source_ty r * sink_ty r) Effect.t
 
 let await_readable fd = Effect.perform (Await_readable fd)
 let await_writable fd = Effect.perform (Await_writable fd)
 
+(* Using Obj.magic to cast the result of Effect.perform (Pipe sw) *)
```
patricoferris, line 15: This comment needs to go.

lib_eio/unix/eio_unix.mli
```diff
@@ -97,7 +97,7 @@ module Private : sig
     | Await_readable : Unix.file_descr -> unit Effect.t      (** See {!await_readable} *)
     | Await_writable : Unix.file_descr -> unit Effect.t      (** See {!await_writable} *)
     | Get_monotonic_clock : Eio.Time.Mono.ty r Effect.t
-    | Pipe : Eio.Switch.t -> (source_ty r * sink_ty r) Effect.t    (** See {!pipe} *)
+    | Pipe : Eio.Switch.t -> (source_ty r * sink_ty r) Effect.t   (** See {!pipe} *)
```
patricoferris, line 100: This whitespace change shouldn't be here. It might be easier to revert to the start and add your changes back rather than trying to undo changes.

lib_eio/unix/private.ml
```diff
@@ -7,13 +7,13 @@ type _ Effect.t +=
   | Await_readable : Unix.file_descr -> unit Effect.t
   | Await_writable : Unix.file_descr -> unit Effect.t
   | Get_monotonic_clock : Eio.Time.Mono.ty r Effect.t
-  | Pipe : Switch.t -> (source_ty r * sink_ty r) Effect.t
+  | Pipe : Switch.t -> ( source_ty r * sink_ty r) Effect.t
```
patricoferris, line 10: Similar whitespace comment.

lib_eio/process.mli
```diff
@@ -140,7 +140,10 @@ val parse_out :
 
 (** {2 Pipes} *)
 
-val pipe : sw:Switch.t -> _ mgr -> [Flow.source_ty | Resource.close_ty] r * [Flow.sink_ty | Resource.close_ty] r
+val pipe : 
+  sw:Switch.t -> 
+  _ mgr -> 
+  [Flow.source_ty | Resource.close_ty] r * [Flow.sink_ty | Resource.close_ty] r
```
patricoferris, line 146: Now we seem to have lost the change you were trying to fix 😅 ?

The type of this outer (i.e. not the function `pipe` instead the `Pi` interface) should be the relaxed return types.

patricoferris (changes requested):

lib_eio/process.mli
Arogbonlo: true true. How did I miss that.. I've just rectified that.

Arogbonlo:
hey @talex5 @patricoferris.

I just made the changes. You can check them out.
Thank you.

lib_eio/unix/private.ml
```diff
@@ -13,7 +13,6 @@ let await_readable fd = Effect.perform (Await_readable fd)
 let await_writable fd = Effect.perform (Await_writable fd)
 
 let pipe sw = Effect.perform (Pipe sw)
-
```
talex5, line 16: @Arogbonlo Sorry to keep on about this, but you MUST look at your PRs before asking people to review them. Make sure all the changes are supposed to be there.

Arogbonlo: Thank you for this.

But please I'd need your assistance here. Where am I missing it on this change?

Arogbonlo: @talex5  In the previous commit, there was a comment I put there. So the only thing I did here was to delete the comment. 

Except i don't get what you mean. Please I'd like to be clear on it. Thank you!

patricoferris: @Arogbonlo -- the change you made here is a deleted newline. Again, this is why reverting your changes (rather than adding a comment and then deleting a comment) would make more sense. We can always squash all of the commits to tidy up the history, but it would be good for the contributor to take care of this.

These kinds of changes (newlines, whitespace etc.) are unnecessary (and add to the review process). As a contributor it is part of your job to also review your own code for these kinds of changes and remove them. Other reviewers want to focus on the _meaningful_ code changes and help you get the logic correct.

PR #771, opened 2024-10-18
fix: test_symlink

create2000:
worked on the test_symlink function


patricoferris (approved):
Looks great and works for me, could you just indent the `if not` code to match the rest? I was going to do it but I can't push to your PR.

create2000:
Thank you @patricoferris -- I have fixed the indentation.

talex5 (approved):
Looks good.

The reason dune isn't warning about unused functions here (there are others too) is that the tests are spread across multiple files. Possibly we could do with some nearly-empty .mli files here (but not needed as part of this PR).

PR #777, opened 2024-10-29
WIP: Skip network tests

create2000:
This is a work-in-progress for skipping tests when network is unavailable

PR #778, opened 2024-10-29
WIP: added skip

create2000:
This is a work-in-progress that adds skip to the test file.


create2000:
The questions asked [here](https://github.com/ocaml-multicore/eio/pull/777#issuecomment-2443669406) are still much needed here too. Thank you :))

talex5 (changes requested):
You need to define the `skip` function too (to raise a suitable exception for MDX).

BTW, you don't need to open a new PR for every revision. Just update your existing branch (`git push -f` to force-update if you're undoing old commits).

lib_eio_linux/tests/test.ml
```diff
@@ -32,9 +32,10 @@ let test_poll_add () =
   assert (sent = 1);
   let result = Promise.await_exn thread in
   Alcotest.(check string) "Received data" "!" result
+  )
 
 let test_poll_add_busy () =
-  Eio_linux.run ~queue_depth:2 @@ fun _stdenv ->
+  Eio_linux.run ~fallback:skip (fun _stdenv ->
```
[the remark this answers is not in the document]
create2000: Alright. I have made some changes.

lib_eio_linux/tests/test.ml
```diff
@@ -19,7 +23,7 @@ let read_one_byte ~sw r =
     )
 
 let test_poll_add () =
-  Eio_linux.run @@ fun _stdenv ->
+  Eio_linux.run ~fallback:skip (fun _stdenv ->
```
patricoferris, line 26: Still no need to make the change from `@@` to `(`

patricoferris (commented):
For skipping tests in the Alcotest framework there is https://ocaml.org/p/alcotest/1.8.0/doc/Alcotest/index.html#val-skip

lib_eio_linux/tests/test.ml
create2000: Thank you @patricoferris -- I  have gone through the documentation, To get it correct, I used `Alcotest.skip`. Something like this 

```
let skip_io_uring () = 
    Alcotest.skip "Skipping test: io_uring not available in Docker";;
``` 
However, when i use this, i encounter this error: `Alcotest.skip "Skipping test: io_uring not available in Docker";;
Error: This expression has type string but an expression was expected of type unit`

Is there something i am doing wrong? I would appreciate help.

patricoferris: @create2000 the type of `skip` is 

```ocaml
val skip : unit -> 'a 
```

It doesn't expect a string as the first argument but a value of type `unit` (i.e `()`).  You could perhaps log something with `Eio.traceln` and then `Alcotest.skip ()`. Not that this is only for the Alcotest-based Linux tests.

create2000: @patricoferris Thank you for the clarification. I have modified it and logged a message using the `Eio.tracein` which was displayed in the generated file after running the file. 

This is the new modification: 

```
let skip_io_uring () =
  Eio.traceln "Skipping test: io_uring not available in Docker";
  Alcotest.skip ()
```

```
let test_poll_add () =
  Eio_linux.run ~fallback:(skip_io_uring()) 
```

talex5 (commented):
Note that in #777 you were trying to make this work for MDX (.md) files, but this PR is changing an Alcotest (.ml) file. Here you don't need a new exception, as Alcotest already supports skipping. However, you should use the error message that `Eio_linux.run` gives you, rather than just assuming that the problem is Docker.

create2000:
Hello @talex5 -- thank you for the insight. How do I get the error message from **Eio_linux.run**? 

Also, since Alcotest already provides a skip,

talex5:
The argument to fallback is ``[`Msg of string]``.

create2000:
Hello @talex5 -- regarding adding the skip feature to MDX, I discovered there are two .md files: spawn.md and fd_sharing.md. Am I to create a new .md file or add the skip feature to one of the already existing .md files?

create2000:
@talex5 -- I have been trying to add the skip feature to MDX and I need some help: 

First, I have MDX installed via spam, but when I try to check the version, it doesn't show. So, I decided to clone the MDX repo and followed the instructions in the `contributing.md` file. Now, when i run the test, it shows me the output. But how do i proceed to create the skip feature? 

From the `contributing.md` I'll have to use the `test\lib` for unit tests and the `test\bin` for end-to-end test. This is where i am confused. Do i need to create a separate file in the `test\lib` for my skip feature? And will i need to further add it to my `test\bin`?

I will appreciate any additional help. Thank you ;))

talex5:
You'll need to add a test to `test/bin` I would expect.

PR #781, opened 2024-11-20
Support `chown` in eio_posix and eio_linux

patricoferris:
This PR adds support to Eio for changing the ownership of a path using `fchownat`.

### DIFF
--- a/lib_eio/path.ml
+++ b/lib_eio/path.ml
@@ -266,3 +266,11 @@ let read_link t =
   with Exn.Io _ as ex ->
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "reading target of symlink %a" pp t
+
+let chown ~follow ?uid ?gid t =
+  let (Resource.T (dir, ops), path) = t in
+  let module X = (val (Resource.get ops Fs.Pi.Dir)) in
+  try X.chown ~follow ?uid ?gid dir path
+  with Exn.Io _ as ex ->
+    let bt = Printexc.get_raw_backtrace () in
+    Exn.reraise_with_context ex bt "changing ownership of %a" pp t
--- a/lib_eio/path.mli
+++ b/lib_eio/path.mli
@@ -247,3 +247,10 @@ val chmod : follow:bool -> perm:File.Unix_perm.t -> _ t -> unit
 (** [chmod ~follow ~perm t] allows you to change the file mode bits.
 
     @param follow If [true] and [t] is a symlink then change the target's mode bits. *)
+
+val chown : follow:bool -> ?uid:int64 -> ?gid:int64 -> _ t -> unit
+(** [chown ~follow ~uid ~gid t] changes the ownership of [t] to be [uid, gid].
+
+    [uid] or [gid] can be omitted to leave the current value unchanged.
+
+    @param follow If [t] is a symbolic link, change the ownership of its target. *)
--- a/lib_eio/unix/eio_unix.mli
+++ b/lib_eio/unix/eio_unix.mli
@@ -133,6 +133,8 @@ module Private : sig
   val read_link_unix : Unix.file_descr option -> string -> string
   val chmod : Fd.t -> string -> flags:int -> mode:int -> unit
   val chmod_unix : Unix.file_descr -> string -> flags:int -> mode:int -> unit
+  val chown : flags:int -> uid:int64 -> gid:int64 -> Fd.t -> string -> unit
+  val chown_unix : flags:int -> uid:int64 -> gid:int64 -> Unix.file_descr -> string -> unit
 end
 
 module Pi = Pi
--- a/lib_eio_linux/low_level.ml
+++ b/lib_eio_linux/low_level.ml
@@ -741,6 +741,15 @@ let chmod ~follow ~mode dir path =
     Eio_unix.run_in_systhread ~label:"chmod" (fun () -> Eio_unix.Private.chmod parent leaf ~mode ~flags)
   with Unix.Unix_error (code, name, arg) -> raise @@ Err.wrap_fs code name arg
 
+let chown ~follow ?(uid=(-1L)) ?(gid=(-1L)) fd path =
+  let module At = Uring.Linkat_flags in
+  let follow = if follow then At.(empty_path + symlink_follow) else At.empty_path in
+  let flags = (follow :> int) in
+  try
+    with_parent_dir_fd fd path @@ fun parent leaf ->
+    Eio_unix.run_in_systhread ~label:"chown" (fun () -> Eio_unix.Private.chown ~flags ~uid ~gid parent leaf)
+  with Unix.Unix_error (code, name, arg) -> raise @@ Err.wrap_fs code name arg
+
 (* https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml *)
 let getaddrinfo ~service node =
   let to_eio_sockaddr_t {Unix.ai_family; ai_addr; ai_socktype; ai_protocol; _ } =
--- a/lib_eio_windows/low_level.ml
+++ b/lib_eio_windows/low_level.ml
@@ -146,6 +146,12 @@ let read_link ?dirfd path =
   in_worker_thread @@ fun () ->
   Eio_unix.Private.read_link dirfd path
 
+let chown ?dirfd ~follow:_ ?(uid=(-1L)) ?(gid=(-1L)) path =
+  in_worker_thread @@ fun () ->
+  match dirfd with
+  | None -> failwith "Chown is unsupported on Windows"
+  | Some dirfd -> Eio_unix.Private.chown ~flags:0 ~uid ~gid dirfd path
+
 external eio_readv : Unix.file_descr -> Cstruct.t array -> int = "caml_eio_windows_readv"
 
 external eio_preadv : Unix.file_descr -> Cstruct.t array -> Optint.Int63.t -> int = "caml_eio_windows_preadv"
--- a/tests/fs.md
+++ b/tests/fs.md
@@ -939,6 +939,57 @@ Reading at the end of a file:
 - : unit = ()
 ```
 
+# Changing Ownership
+
+<!-- This test is written a little obscurely as we only want to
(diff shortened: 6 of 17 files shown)


lib_eio/path.ml
```diff
@@ -238,3 +238,11 @@ let read_link t =
...
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "reading target of symlink %a" pp t
+
+let chown ~follow ~uid ~gid t =
+  let (Resource.T (dir, ops), path) = t in
+  let module X = (val (Resource.get ops Fs.Pi.Dir)) in
+  try X.chown ~follow ~uid ~gid dir path
+  with Exn.Io _ as ex ->
+    let bt = Printexc.get_raw_backtrace () in
+    Exn.reraise_with_context ex bt "chowing file %a" pp t
```
talex5, line 248:
```suggestion
    Exn.reraise_with_context ex bt "changing ownership of %a" pp t
```
Users probably won't known what `chow(n)ing` means.

lib_eio/unix/eio_unix.mli
```diff
@@ -107,6 +107,9 @@ module Private : sig
 
   val read_link : Fd.t option -> string -> string
   val read_link_unix : Unix.file_descr option -> string -> string
+
+  val chown : flags:int -> uid:int64 -> gid:int64 -> Fd.t -> string -> unit
+  val chown_unix : flags:int -> uid:int64 -> gid:int64 -> Unix.file_descr -> string -> unit
```
talex5, line 112: I guess this is just copying the pattern above, but I think we only have `read_link_unix` because it's needed during path resolution and wrapping each step in `Fd.t` seemed unreasonable.

lib_eio/unix/stubs.c
```diff
@@ -52,3 +52,24 @@ CAMLprim value eio_unix_readlinkat(value v_fd, value v_path, value v_cs) {
...
+  caml_unix_error(EOPNOTSUPP, "fchownat not supported on Windows", v_path);
+#else
+  CAMLparam3(v_path, v_uid, v_gid);
+  char *path;
+  int fd = Int_val(v_fd);
+  int ret;
+  caml_unix_check_path(v_path, "fchownat");
+  path = caml_stat_strdup(String_val(v_path));
+  caml_enter_blocking_section();
+  ret = fchownat(fd, path, Int64_val(v_uid), Int64_val(v_gid), Int_val(v_flags));
```
talex5, line 67: You can't use `Int64_val` inside a blocking section (the GC may be moving it concurrently).

lib_eio_linux/low_level.ml
```diff
@@ -524,6 +524,15 @@ let read_link fd path =
     Eio_unix.run_in_systhread ~label:"read_link" (fun () -> Eio_unix.Private.read_link (Some parent) leaf)
   with Unix.Unix_error (code, name, arg) -> raise @@ Err.wrap_fs code name arg
 
+let chown ~follow ~uid ~gid fd path =
+  let module At = Uring.Linkat_flags in
+  let follow = if follow then At.empty_path else At.(empty_path + symlink_follow) in
```
talex5, line 529: This looks wrong. The flag is `nofollow`, not `follow`.

lib_eio_windows/low_level.ml
```diff
@@ -138,6 +138,12 @@ let read_link ?dirfd path =
   in_worker_thread @@ fun () ->
   Eio_unix.Private.read_link dirfd path
 
+let chown ?dirfd ~follow:_ ~uid ~gid path =
+  in_worker_thread @@ fun () ->
+  match dirfd with
+  | None -> failwith "Chown is unsupported on Windows"
```
talex5, line 144: Could just use `Unix.chown` here.

talex5 (commented):
Thanks!

Rebasing on `main` should fix the OpenBSD CI error.

lib_eio_windows/low_level.ml
patricoferris: Hmm I'm wondering whether `chown` should be in the cross-platform API at all. 

`Unix.chown` isn't implemented on Windows: 
https://github.com/ocaml/ocaml/blob/c484c6932fa2eae03ba0f5a7dbdb26e3eee65eb0/otherlibs/unix/unix_win32.ml#L450 -- it's no implemented in python's `os.chown` either?

Do we have a rule for what makes it into the cross-platform API and what doesn't ? I mean we already have `File.Unix_perm.t`.

lib_eio/unix/eio_unix.mli
patricoferris: Most of the low-level function in the posix backend use `Unix.file_descr` and this is in keeping with that? 

```ocaml
  Resolve.with_parent "<operation>" dirfd path @@ fun dirfd path ->
  let dirfd = Option.value dirfd ~default:at_fdcwd in
  ```
  
  They follow this pattern where `dirfd` ends up being a `Unix.file_descr` -- should we wrap them with `Fd.of_unix` + a switch?

File: lib_eio_linux/low_level.ml (hunk 4 above)
Comment (talex5) [reply lib_eio_linux/low_level.ml]: The Docker container has groups `opam` and `users`, so we could test it. Maybe just get the list of groups the current process is in and try switching the file to a different one?

File: lib_eio_windows/low_level.ml (hunk 5 above)
Comment (talex5) [reply lib_eio_windows/low_level.ml:144]: I think it's OK to keep it. We might want a proper operation-not-supported error at some point.

**Comment by talex5** [comment]
BTW, I was planning to make a new Eio release tomorrow. Do you want me to delay that to get this in, or should I go ahead and merge this later?

**Comment by patricoferris** [comment]
Oops sorry -- cut the release, no rush on this :))

File: lib_eio/path.mli [hunk 6]
```diff
@@ -220,3 +220,6 @@ val symlink : link_to:string -> _ t -> unit
     {[
       Eio.Path.symlink (dir / "current") ~link_to:"version-1.0"
     ]} *)
+
+val chown : follow:bool -> uid:int64 -> gid:int64 -> _ t -> unit
+(** [chown ~follow ~uid ~gid t] changes the ownership of [t] to be [uid, gid]. *)
```
talex5: Need to document `follow`. Also, `fchown` says "If the owner or group is specified as -1, then that ID is not changed". So maybe:

```suggestion
val chown : follow:bool -> ?uid:int64 -> ?gid:int64 -> _ t -> unit
(** [chown ~follow ~uid ~gid t] changes the ownership of [t] to be [uid, gid].

    [uid] or [gid] can be [None] to leave the current value unchanged.

    @param follow If [t] is a symbolic link, change the permission of its target. *)
```

talex5 (commented):
I squashed and rebased this. Just wondering if the uid and gid should be optional.

lib_eio/path.mli
patricoferris: Having a means to update only one of them seems sensible to me, thanks for taking a look!

talex5 (approved):
I rebased it and made `uid` and `gid` optional. Should be ready for merging now.

lib_eio_linux/low_level.ml
```diff
@@ -741,6 +741,15 @@ let chmod ~follow ~mode dir path =
     Eio_unix.run_in_systhread ~label:"chmod" (fun () -> Eio_unix.Private.chmod parent leaf ~mode ~flags)
   with Unix.Unix_error (code, name, arg) -> raise @@ Err.wrap_fs code name arg
 
+let chown ~follow ?(uid=(-1L)) ?(gid=(-1L)) fd path =
+  let module At = Uring.Linkat_flags in
+  let follow = if follow then At.(empty_path + symlink_follow) else At.empty_path in
```
avsm, line 746: are you sure this is right? Using `Linkat_flags` for a chown call.

lib_eio/path.mli
```diff
@@ -247,3 +247,10 @@ val chmod : follow:bool -> perm:File.Unix_perm.t -> _ t -> unit
 (** [chmod ~follow ~perm t] allows you to change the file mode bits.
 
     @param follow If [true] and [t] is a symlink then change the target's mode bits. *)
+
+val chown : follow:bool -> ?uid:int64 -> ?gid:int64 -> _ t -> unit
+(** [chown ~follow ~uid ~gid t] changes the ownership of [t] to be [uid, gid].
+
+    [uid] or [gid] can be omitted to leave the current value unchanged.
+
+    @param follow If [t] is a symbolic link, change the permission of its target. *)
```
avsm:
```suggestion
    @param follow If [t] is a symbolic link, change the ownership of its target. *)
```

tests/fs.md
```diff
@@ -939,6 +939,57 @@ Reading at the end of a file:
 - : unit = ()
 ```
 
+# Changing Ownership
+
+<!-- This test is written a little obscurely as we only want to
+     run it on macOS, but MDX only supports os_type (Unix, Win, Cygwin) -->
```
avsm, line 945: Don't we need chown tests on Linux as well?

talex5: I think the problem is it's hard to test chown as an unprivileged user.

lib_eio_linux/low_level.ml
talex5: The naming isn't great, but `linkat` and `chown` do both accept the same flags (`AT_EMPTY_PATH` and `AT_SYMLINK_NOFOLLOW`).

avsm: But they don't though; linkat(2) has AT_EMPTY_PATH and AT_SYMLINK_FOLLOW (not AT_SYMLINK_NOFOLLOW!). Doesnt that reverse the logic here?

talex5: Ah, I see. Yes, that does look wrong then!

lib_eio_linux/low_level.ml
talex5: This seems to have been dropped. I'll implement it now.

lib_eio_linux/low_level.ml
avsm: Attempted a test in https://github.com/ocaml-multicore/eio/pull/864 that did fail on the previous and works now (but my bandwidth is a little off in Greenland rn so i'll trust CI!)

PR #772, opened 2024-10-24
Implementing the ``chmod`` support

webbunivAdmin:
### Add `chmod` Support to Eio

This PR implements support for changing file permissions using the `chmod` operation in the Eio library.
#764 

#### Key Changes:
- **Added `chmod` function** that allows setting file permissions on a target resource:
  - **Parameters:**
    - `~follow`: Determines whether to follow symlinks.
    - `~perm`: Specifies the permission bits to set.
    - `t`: The target file or directory.
- **Exception Handling:**
  - Exceptions are handled to provide robust error reporting.
  - Context-aware re-raising of exceptions in case of failure for better debugging.
- **Seamless Integration** with Eio’s resource abstraction, supporting various permission modes.

#### Tests:
-

PR #785, opened 2024-11-29
Support `chmod` in Eio_linux and Eio_posix

patricoferris:
This PR is a tidied up version of #772 

I'm not really sure what to do on Windows -- `Unix.chmod` does use `_wchmod` but the Unix-y permissions don't really make much sense on Windows afaict. Perhaps this isn't a problem for this PR.

### DIFF
--- a/lib_eio/fs.ml
+++ b/lib_eio/fs.ml
@@ -72,6 +72,7 @@ module Pi = struct
     val rename : t -> path -> _ dir -> path -> unit
     val read_link : t -> path -> string
     val symlink : link_to:path -> t -> path -> unit
+    val chmod : t -> follow:bool -> perm:File.Unix_perm.t -> path -> unit
     val pp : t Fmt.t
     val native : t -> string -> string option
   end
--- a/lib_eio/path.ml
+++ b/lib_eio/path.ml
@@ -238,6 +238,14 @@ let symlink ~link_to source =
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "creating symlink %a -> %s" pp source link_to
 
+let chmod ~follow ~perm t =
+  let (Resource.T (dir, ops), path) = t in
+  let module X = (val (Resource.get ops Fs.Pi.Dir)) in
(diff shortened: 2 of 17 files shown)


patricoferris:
It seems `fchmodat` support is pretty flakey in terms of support on Linux, in particular `AT_SYMLINK_NOFOLLOW` is still not supported as far as I can tell (hence no test). It also seems some machines don't support it...

talex5:
Changing `Eio.Exn.Backend.show := false` to `true` in `fs.md` might show what's going on.

lib_eio_linux/include/discover.ml
```diff
@@ -0,0 +1,20 @@
+module C = Configurator.V1
+
+let () =
+  C.main ~name:"discover" (fun c ->
+      let c_flags = ["-D_LARGEFILE64_SOURCE"; "-D_XOPEN_SOURCE=700"; "-D_GNU_SOURCE";] in
+      let present_defs =
+        C.C_define.import c ~c_flags
+          ~includes:["fcntl.h"]
+          C.C_define.Type.[
+            "AT_SYMLINK_NOFOLLOW", Int;
```
talex5, line 10: You can get this from uring I think (`Statx.Flags.symlink_nofollow`).

avsm:
I took the liberty of rebasing this and:
- used uring nofollow flag in eabfc1eb3522ea863545ebd553d3f65a3d53d6f0
- switched to File.Unix_perm.t as a type alias instead of int in e29a8221e0cb5308d9198f21fbc50704ae7604d5
- dont leak a Failure exception if not implemented and do a ENOSUP unix error instead c295f58c64b6fe195f0d2639221cf9f88f1a207b

lib_eio_linux/include/discover.ml
avsm: done in eabfc1eb3522ea863545ebd553d3f65a3d53d6f0

talex5 (approved):
Thanks! I just added a fix to work with the released version of uring and a minor doc fix.

Issue #793, opened 2025-01-11
on Windows, reading from stdin blocks the whole domain.


bdodrem:
I tried to read a line from stdin with a timeout (code below). On Windows, the process blocks until I press Enter (not on Linux).
So I guess, reading the stdin on Windows blocks the whole domain.

```ocaml
open Eio.Std
open Eio.Buf_read
let _ = Eio_main.run (fun env -> 
    let readoneline () = traceln "%s" @@ parse_exn ~max_size:1024 line env#stdin  in
    Eio.Time.with_timeout_exn env#clock 2. readoneline
  )

```

Running the code : 
```bash
$ cat | dune exec ./test.exe
```
blocks untill `Enter` is pressed

talex5:
Probably needs a check for this. I see posix has one, but windows doesn't:

https://github.com/ocaml-multicore/eio/blob/fdd2593e33086b1a37d3afb21e669a6a77f6a37c/lib_eio_posix/low_level.ml#L50-L53

https://github.com/ocaml-multicore/eio/blob/fdd2593e33086b1a37d3afb21e669a6a77f6a37c/lib_eio_windows/low_level.ml#L41-L43

Issue #792, opened 2025-01-09
on Windows reading stdin from a pipe command throws"broken pipe" exception 


bdodrem:
On Windows, the following code :
```
     open Eio.Std
     open Eio.Buf_read  

     let _ = Eio_main.run (fun env ->  Seq.iter (traceln "%s" ) @@ lines @@ of_flow env#stdin ~max_size:1024  )
```

 when run in a terminal (bash console) with the command :
`    echo 1 | dune exec ./test.exe
`

Throws the exception : 
```
+1
Fatal error: exception Eio.Io Net Connection_reset Unix_error (Broken pipe, "stub_cstruct_read", "")

```
(I'm currently using ocaml 5.2.1 and eio 1.2 on MINGW64 )

talex5:
I did a search on the Internet, and it turns out I reported [the same bug in Lwt](https://github.com/ocsigen/lwt/issues/20) back in 2013. D'oh!

bdodrem:
The issue seemed to be solved when I replaced `Unix_cstruct.read` with `Unix.read_bigarray` in the function `read_cstruct` _(lib_eio_windows/low_level.ml)._

```diff
diff --git a/lib_eio_windows/low_level.ml b/lib_eio_windows/low_level.ml
index 85f25cd..962210d 100755
--- a/lib_eio_windows/low_level.ml
+++ b/lib_eio_windows/low_level.ml
@@ -42,9 +42,9 @@ let read fd buf start len =
   Fd.use_exn "read" fd @@ fun fd ->
   do_nonblocking Read (fun fd -> Unix.read fd buf start len) fd

-let read_cstruct fd buf =
+let read_cstruct fd (buf:Cstruct.t) =
   Fd.use_exn "read_cstruct" fd @@ fun fd ->
-  do_nonblocking Read (fun fd -> Unix_cstruct.read fd buf) fd
+  do_nonblocking Read (fun fd -> Unix.read_bigarray fd buf.buffer buf.off buf.len) fd

 let write fd buf start len =
   Fd.use_exn "write" fd @@ fun fd ->
```

Hopefully this will solve the bug

talex5:
This fix looks right to me.

@patricoferris how about removing `eio_windows_cstruct_stubs.c` and using the Unix functions instead?

talex5:
> Do I need to create a pull request ?

That would be easier, yes. Though I can do it if that's a problem for you.

dijkstracula:
[...]

PR #795, opened 2025-01-25
On Windows, fix stdin broken-pipe and blocking domain.  Issues #793 and #792.

bdodrem:
Hi,
 
   The purpose of this pull request is to 

* fix blocking stdin on Windows : issue #793. 
      Adding `await_readable` before reading fd 

* fix broken pipe exception : issue #792.  
      Use `Unix.read_bigarray` instead of `Unix_cstruct.read` 

* replace _eio_windows_cstruct_stubs.c_ by Unix functions.
     Since Ocaml 5.2, `Unix.read_bigarray` and `Unix.write_bigarray` are available.


talex5 (approved):
Looks good to me - thanks!

/cc @patricoferris, who added the cstruct stubs file

talex5:
I also bumped the minimum OCaml version to 5.2, since this PR needs it.

PR #796, opened 2025-02-09
Introduce Buf_write.of_flow

vog:
To serialize buffered into a `flow`, at the moment there is only `Buf_write.with_flow`, so those serializers are always expected to always run wrapper within a single function.

This PR introduces `Buf_write.of_flow` which allows a serializer to outlive its invocation, but instead being restricted by a surrounding `Switch`, as usual. Also, proper closing of the write is ensured by `Switch.on_release` in success as well as error cases.. This makes `Buf_write` more straight forward to use e.g. in network protocols, which otherwise need unfortunate workarounds like this one:

* https://github.com/m-click/pgx/blob/upgrade-ocaml5-eio/pgx_eio/src/pgx_eio.ml#L72-L81
* See also: https://github.com/arenadotio/pgx/pull/134

Moreover, `Buf_write.with_flow` is simplified by just calling `Buf_write.of_flow` within a local `Switch`.

### DIFF
--- a/lib_eio/buf_write.ml
+++ b/lib_eio/buf_write.ml
@@ -515,21 +515,16 @@ let copy t flow =
   try aux ()
   with End_of_file -> ()
 
-let with_flow ?(initial_size=0x1000) flow fn =
-  Switch.run ~name:"Buf_write.with_flow" @@ fun sw ->
+let of_flow ~sw ?(initial_size=0x1000) flow =
   let t = create ~sw initial_size in
+  Switch.on_release sw (fun () -> close t);
   Fiber.fork ~sw (fun () -> copy t flow);
-  match fn t with
-  | x ->
-    close t;
-    x
-  | exception ex ->
-    close t;
-    (* Raising the exception will cancel the writer thread, so do a flush first.
-       We don't want to flush if cancelled, but in that case the switch will
-       end the writer thread itself (and [flush] will raise). *)
-    flush t;
-    raise ex
+  t
+
+let with_flow ?initial_size flow fn =
+  Switch.run ~name:"Buf_write.with_flow" @@ fun sw ->
+  let t = of_flow ~sw ?initial_size flow in
+  fn t
 
 let rec serialize t writev =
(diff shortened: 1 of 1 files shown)


patricoferris:
Hi @vog,

Thank you for the PR.

Unfortunately this would change the semantics of `with_flow` compared to what was there before, which accounts for the failing tests. The forked `copy` action relies on the fact that the buffer will eventually be closed and this will raise an `End_of_file` exception. Moving this into the switch's `on_release` callbacks means this will now only succeed if the user _manually_ closes the buffer in `with_flow`. 

Also you will need to expose `of_flow` in the `.mli` file for people to be able to use it.

Note also that you can, with a little boiler plate, workaround this by doing things manually:

```ocaml
let of_flow ~sw flow =
  (* Copied from https://github.com/ocaml-multicore/eio/blob/62b9714f0e2ed8e72046c8e5808202e1e3ca9cd7/lib_eio/buf_write.ml#L507

     We have to do our own copy, because we can't [shift] until the write is complete. *)
  let copy t flow =
    let rec aux () =
      let iovecs = Eio.Buf_write.await_batch t in
      let wrote = Eio.Flow.single_write flow iovecs in
      Eio.Buf_write.shift t wrote;
      aux ()
    in
    try aux ()
    with End_of_file -> ()
  in
  let t = Eio.Buf_write.create ~sw 0x1000 in
  Eio.Switch.on_release sw (fun () -> Eio.Buf_write.close t);
  Fiber.fork ~sw (fun () -> copy t flow);
  t
```

Note that this will likely have the same problems as before (perhaps a `fork_daemon` might make more sense there) but it really depends on your use case. Hopefully that helps a bit.

talex5:
The Eio API is trying to avoid the mistake where you think you've finished writing but the buffer hasn't been flushed yet. Adding `of_flow` makes that mistake easy again (you have to wait for the switch to go out of scope to know the writes have taken effect).

In the PR you link, perhaps it would be simpler to return the flow and let the caller wrap it in a buffer?

Issue #595, opened 2023-07-28
Creating a process group in `Fork_action` (?)


smondet:
Would it make sense to add calls to `setpgid(0,0)` before the `execve`?

I think that would allow for better use of `signal` to kill entire process groups. 

https://pubs.opengroup.org/onlinepubs/9699919799/functions/setpgid.html

talex5:
Optional support would be good, yes.

PR #802, opened 2025-05-20
Add process groups to unix backends

patricoferris:
This is, I hope, most of the plumbing for #595. It remains to decide on how best to expose this in `Eio.Process` if at all.

Also if anyone has a good, reproducible test for this, that would be good too!

### DIFF
--- a/lib_eio/unix/fork_action.c
+++ b/lib_eio/unix/fork_action.c
@@ -237,3 +237,24 @@ static void action_dups(int errors, value v_config) {
 CAMLprim value eio_unix_fork_dups(value v_unit) {
   return Val_fork_fn(action_dups);
 }
+
+static void action_setpgid(int errors, value v_config) {
+  #ifdef _WIN32
+  eio_unix_fork_error(errors, "setpgid", "Unsupported operation on windows");
+  _exit(1);
+  #else
+  value vpid = Field(v_config, 1);
+  value vpgid = Field(v_config, 2);
+
+  int r;
+  r = setpgid(Int_val(vpid), Int_val(vpgid));
+  if (r != 0) {
+    eio_unix_fork_error(errors, "setpgid", strerror(errno));
+    _exit(1);
+  }
+  #endif
+}
+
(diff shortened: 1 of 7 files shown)


lib_eio/unix/fork_action.c
```diff
@@ -237,3 +237,23 @@ static void action_dups(int errors, value v_config) {
 CAMLprim value eio_unix_fork_dups(value v_unit) {
   return Val_fork_fn(action_dups);
 }
+
+static void action_setpgid(int errors, value v_config) {
+  #ifdef _WIN32
+  eio_unix_fork_error(errors, "set_cloexec", "Unsupported operation on windows");
```
talex5, line 243:
```suggestion
  eio_unix_fork_error(errors, "setpgid", "Unsupported operation on windows");
  _exit(1);
```

Perhaps `eio_unix_fork_error` should call `_exit` itself. Looks like everything that uses it either does that or is wrong.

talex5 (approved):
Looks reasonable to me.

lib_eio/unix/fork_action.c
patricoferris: Indeed!

lib_eio/unix/fork_action.mli
```diff
@@ -58,3 +58,8 @@ val inherit_fds : (int * Fd.t * [< blocking]) list -> t
     A mapping from an FD to itself simply clears the close-on-exec flag.
 
     After this, the new FDs may also be set as blocking or non-blocking, depending on [flags]. *)
+
+val setpgid : pid:int -> pgid:int -> t
+(** [setpgid ~pid ~pgid] sets the process group ID to [pgid] for the process [pid].
+
+    If [pid] is [0] then the process ID of the calling process shall be used. *)
```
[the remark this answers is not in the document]
patricoferris: Good point, I have made the API only take the `pgid` and added comments explaining the effect of `setpgid 0` in 1eb1197

talex5:
I've rebased it to fix the CI problems. OK to merge once that passes.

Issue #805, opened 2025-05-26
Nested Fiber.any race condition


adamchol:
Hey all, thanks for the awesome work on Eio!

I have a weird race condition with nested `Fiber.any` and I don't quite understand it.

Here is a simplified example of the code that reproduces the situation:
```ocaml
Eio_main.run (fun env ->
    let promise_r, promise_u = Eio.Promise.create () in

    let res =
      Eio.Fiber.any
        ~combine:(fun x y -> x + y)
        [
          (fun () ->
            (* function 1 *)

            (* let the second function start *)
            Eio.Time.sleep env#clock 1.0;

            Eio.Promise.resolve promise_u 1;
            Printf.printf "Promise resolved\n%!";
            10);
          (fun () ->
            Eio.Fiber.any
              ~combine:(fun x y -> x + y)
              [
                (fun () ->
                  (* function 2 *)
                  Eio.Time.sleep env#clock 10.;
                  2);
                (fun () ->
                  (* function 3 *)
                  let x = Eio.Promise.await promise_r in
                  Printf.printf "Promise returned\n%!";
                  x);
              ]);
        ]
    in
    Printf.printf "Result: %d\n" res)
```

The output from this is:
```
Promise resolved
Promise returned
Result: 10
```

Although, what I would expect from this is `Result: 11`. From my understanding, when the promise is resolved and "function 1" returns and other function are being checked if they are "unlocked". The "function 3" is unlocked and it returns, so it would seem like the nested `Fiber.any` should return in a whole and combine on the first `Fiber.any` should be invoked with values 10 and 1. 

But seems like the second `Fiber.any` is cancelled and only the first function actually returns.

Can someone explain what's happening here?

talex5:
It's due to `Fiber.any` explicitly checking whether it's been cancelled before returning:

https://github.com/ocaml-multicore/eio/blob/8f7f82d2c12076af8e9b8b365c58ebadaa963b8c/lib_eio/core/fiber.ml#L152-L154

I think it might be reasonable to change that to:

```ocaml
  match !r, Cancel.get_error parent_c with
  | OK r, _ -> r
  | New, Some ex -> raise ex
```

That would prioritise not losing a calculated value, rather than cancelling quickly.

wokalski:
This patch did solve our problem @talex5. Do you think there are any negative externalities of this change?

talex5:
I don't think so.

[...]

PR #806, opened 2025-05-27
Prioritize returning Fiber.any value over cancelling quickly

adamchol:
This makes changes in `Fiber.any` to prioritize returning the value rather than cancelling the whole `Fiber.any` from the outside if we already have the value from one of it's functions. A sample code showing a specific scenario and more detailed explanation are in the issue #805.

Let me know if this is reasonable to merge.

### DIFF
--- a/lib_eio/core/fiber.ml
+++ b/lib_eio/core/fiber.ml
@@ -150,8 +150,8 @@ let any_gen ~return ~combine fs =
       )
   in
   match !r, Cancel.get_error parent_c with
-  | OK r, None -> r
-  | (OK _ | New), Some ex -> raise ex
+  | OK r, _ -> r
+  | New, Some ex -> raise ex
   | Ex (ex, bt), None -> Printexc.raise_with_backtrace ex bt
   | Ex ex1, Some ex2 ->
     let bt2 = Printexc.get_raw_backtrace () in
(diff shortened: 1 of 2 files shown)


talex5:
The tests need updating too (CI is failing).

adamchol:
I changed the test from `fiber.md` and the description of it to showcase the priority on returning the value.

I'm not sure about the rest of the tests in the CI. One of them is new I think and I don't know how the rest is related to the code I modified. Tried to reproduce those locally but unsuccessfully because of some environment problems with docker. I might need some help with the rest of those CI tests.

talex5:
Rebasing on `main` should fix most of the CI problems (see #808). The `ERROR while compiling mdx.2.5.0` on OCaml 5.4-alpha1 is due to https://github.com/realworldocaml/mdx/issues/469 and can be ignored.

adamchol:
Thanks! Done!

adamchol:
Hi! Can this be merge now, please?

talex5:
I've rebased it and squashed (so there are no commits in the history with failing tests). I slightly modified the test to check that the parent cancellation takes effect on the next yield.

I don't have permission to update your branch (possibly because it's on `main`), so I've pushed the commit here: https://github.com/talex5/eio/pull/new/fiber-any-keep-result

If you reset your branch to that version I'm happy to merge it - thanks!

talex5 (approved):
Thanks!

Issue #808, opened 2025-07-02
posix test suite appears to be broken owing to inconsistent use of Unix_error and Eio.Io Fs


dijkstracula:
On OSX, running the test suite fails:

```
(cli) ntaylor@smgrp ~/c/eio> rm -rf _build && make                                                                     main
dune build @runtest @all
random ops: PASS

inherit_fds: PASS

random ops: PASS

File "tests/process.md", line 1, characters 0-0:
diff --git a/_build/default/tests/process.md b/_build/default/tests/.mdx/process.md.corrected
index 3a8483c..492521d 100644
--- a/_build/default/tests/process.md
+++ b/_build/default/tests/.mdx/process.md.corrected
@@ -138,7 +138,7 @@ Trying to access a path outside of the cwd:
'''ocaml
 # run @@ fun mgr env ->
   Process.run mgr ~cwd:(env#cwd / "..") [ "cat"; "test-cwd" ];;
-Exception: Eio.Io Fs Permission_denied _
+Exception: Unix.Unix_error(Unix.EACCES, "openat", "..")
'''

 If a command fails, we get shown the arguments (quoted if necessary):
File "lib_eio_posix/test/dune", line 7, characters 9-21:
7 |   (names open_beneath test_await)
             ^^^^^^^^^^^^
Fatal error: exception Eio.Io Fs Permission_denied Outside_sandbox ("..")
Raised at Eio_posix__Low_level.Resolve.resolve in file "lib_eio_posix/low_level.ml", line 254, characters 10-97
Called from Eio_posix__Low_level.Resolve.open_beneath_fallback.(fun).aux in file "lib_eio_posix/low_level.ml", line 309, characters 14-48
Called from Eio_posix__Low_level.Resolve.with_state in file "lib_eio_posix/low_level.ml", line 284, characters 10-18
Re-raised at Eio_posix__Low_level.Resolve.with_state in file "lib_eio_posix/low_level.ml", line 289, characters 6-41
Called from Dune__exe__Open_beneath.check.(fun) in file "lib_eio_posix/test/open_beneath.ml", line 15, characters 11-72
Called from Eio_unix__Rcfd.use in file "lib_eio/unix/rcfd.ml", line 170, characters 10-14
Re-raised at Eio_unix__Rcfd.use in file "lib_eio/unix/rcfd.ml", line 175, characters 6-41
[...]

PR #809, opened 2025-07-02
posix: spawn_unix: ensure spawn_unix wraps its call to `openat`

dijkstracula:
Addresses the first issue described in #808 .


dijkstracula:
I note that the MacOS CI job is failing here: https://ocaml.ci.dev/github/ocaml-multicore/eio/commit/f3cd8d57ad124851f7383d70454f01af8e3bd40f/variant/macos-homebrew-5.3_opam-2.3

It's failing with the same backtrace I observed as issue #2 in 808.

PR #810, opened 2025-07-03
Use O_RESOLVE_BENEATH on FreeBSD

talex5:
It needs `-D__BSD_VISIBLE` to be able to see this.

Fix the CI bug this revealed, which is now also affecting macos (reported by @dijkstracula in #808).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
         os:
           - macos-latest
         ocaml-compiler:
-          - 5.2.x
+          - 5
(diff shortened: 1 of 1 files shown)


dijkstracula:
Thank you, Thomas!  I recognise that this is unrelated to this patch; but, I notice that CI is only running against 5.2.  Should that also be bumped to 5.3 (or perhaps also the 5.4 alpha)?

talex5:
Good point. I've updated the macos and Windows CIs to use the latest compiler.

PR #813, opened 2025-07-15
Documentation fixes

jonludlam:
Largely these problems were found by running `odoc_driver`.

The changes to the opam files are to let docs-ci know that other packages need to be available to build the docs correctly. Dune will soon be able to insert these itself, but not yet.

There's still one warning remaining that I haven't fixed as I couldn't immediately see where the right element is - [`{!extension-FD}`](https://github.com/ocaml-multicore/eio/blob/d38a31ca7fa93daca50fb4cb0978d6cd995304e8/lib_eio/unix/eio_unix.mli#L30)

The change to put the canonical tag on the module alias as opposed to in the module itself is not at all obvious, and I need to write up the explanation for it elsewhere.


lib_eio/core/eio__core.mli
```diff
@@ -382,7 +382,7 @@ module Fiber : sig
 
   val get : 'a key -> 'a option
   (** [get key] reads [key] from the map of fiber local variables, returning its
-      value or {!None} if it has not been bound. *)
+      value or [None] if it has not been bound. *)
```
jonludlam, line 385: There's nowhere to link to for core types / core constructors.

lib_eio_linux/low_level.mli
```diff
@@ -109,7 +109,7 @@ val splice : fd -> dst:fd -> len:int -> int
 
     @return The number of bytes copied.
     @raise End_of_file [src] is at the end of the file.
-    @raise Unix.Unix_error(EINVAL, "splice", _) if splice is not supported for these FDs. *)
+    @raise Unix.Unix_error with args [(EINVAL, "splice", _)] if splice is not supported for these FDs. *)
```
jonludlam, line 112: We resolve the exception here, but this was being parsed as `Unix.Unix_error(EINVAL` which couldn't be resolved.

doc/odoc-config-eio.sexp
```diff
@@ -0,0 +1,3 @@
+(libraries unix)
+(packages eio_main bigstring)
```
jonludlam, line 2: This adds the unix library to all libraries build as part of the eio package. This was required because the [eio] library was referring to `Unix.SO_REUSEADDR` in net.mli

jonludlam:
The problem with the `@canonical` tag on `Fork_action` causes the problem that @avsm [pointed out](https://discuss.ocaml.org/t/new-odoc-3-generated-package-documentation-is-live-on-ocaml-org/16967/2)

talex5 (approved):
Haven't seen `odoc_driver` before - this looks great!

PR #821, opened 2025-10-06
Incremental directory reading

patricoferris:
When first providing `readdir` we shelved an incremental version. This PR starts the ball rolling on adding support for something like [`with_dir_entries`](https://github.com/ocaml-multicore/eio/pull/207#pullrequestreview-987295774). Opening early to get some feedback on how it should look, should `Eio.Path.walk` walk directory tree recursively or just a single directory, leaving it to the user to descend should they want to. We should perhaps have a `type entry` too.

I wrote a small benchmark:

```ocaml
open Eio.Std

let ( / ) = Eio.Path.( / )

let dirname i = Fmt.str "dir-%i" i
let filename i = Fmt.str "file-%i" i

let make_bench_dir ?(depth=1) ?(width=1000) dir =
  let rec loop dir depth =
    if depth <= 0 then ()
    else (
      Eio.Path.mkdir ~perm:0o700 (dir / dirname depth);
      Fiber.List.iter ~max_fibers:10 (fun fname ->
        Eio.Path.save ~create:(`If_missing 0o644) (dir / fname) "hello world"
      ) (List.init width filename);
      loop (dir / dirname depth) (depth - 1)
    )
  in
  loop dir depth

let read_dir_walk ~f dir =
  let rec aux acc path =
    match Eio.Path.kind ~follow:true path with
    | `Directory as k ->
       let acc2 = f acc (k, path) in
       let files = Eio.Path.read_dir path in
       List.fold_left (fun acc f -> aux acc (path / f)) acc2 files
    | k -> f acc (k, path)
    in
  let files = Eio.Path.read_dir dir in
  List.fold_left (fun acc f -> aux acc (dir / f)) [] files

let rec walk ~f acc dir =
  Eio.Path.walk dir @@ fun seq ->
  Seq.fold_left (fun acc -> function
    | (`Directory, path) ->
      let acc2 = f acc (`Directory, dir / path) in
      walk ~f acc2 (dir / path)
    | (kind, fname) -> f acc (kind, dir / fname)) acc seq

let time fn =
  let now = Mtime_clock.now_ns () in
  let v = fn () in
  let stop = Mtime_clock.now_ns () in 
  Eio.traceln "Took: %a" Fmt.uint64_ns_span (Int64.sub stop now);
  v

module Path = struct
  include Eio.Path
  type t = Eio.Fs.dir_ty Eio.Path.t 

  let compare a b = String.compare (Eio.Path.native_exn a) (Eio.Path.native_exn b)
end

module S = Set.Make (Path)

let () = 
  Eio_main.run @@ fun env ->
  let tmpdirf = Filename.temp_dir "eio-walk-" "-bench" in
  let tmpdir = env#fs / tmpdirf in
  make_bench_dir ~depth:15 ~width:20_000 tmpdir;
  let f acc (_, path) = path :: acc in
  let files2 = time (fun () -> walk ~f [] tmpdir) |> S.of_list in
  let files1 = time (fun () -> read_dir_walk ~f tmpdir) |> S.of_list in
  Eio.traceln "Files 1: %i\n" (S.cardinal files1);
  Eio.traceln "Files 2: %i\n" (S.cardinal files2);
  Eio.traceln "Same files? %b\n" (S.equal files1 files2);
  let _ : int = Sys.command (Fmt.str "rm -r %s" tmpdirf) in
  ()
```

It is far from perfect and we should take into account FS caching, but on my machine I'm getting the following.

For Eio_linux: 1.46s to 110ms
For Eio_posix: 2.02s to 1.58s

### DIFF
--- a/lib_eio/path.ml
+++ b/lib_eio/path.ml
@@ -106,6 +106,14 @@ let read_dir t =
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "reading directory %a" pp t
 
+let walk t fn =
+  let (Resource.T (dir, ops), path) = t in
+  let module X = (val (Resource.get ops Fs.Pi.Dir)) in
+  try X.with_dir_entries dir path fn
+  with Exn.Io _ as ex ->
+    let bt = Printexc.get_raw_backtrace () in
+    Exn.reraise_with_context ex bt "walking directory %a" pp t
+
 let stat ~follow t =
   let (Resource.T (dir, ops), path) = t in
   let module X = (val (Resource.get ops Fs.Pi.Dir)) in
--- a/lib_eio/path.ml
+++ b/lib_eio/path.ml
@@ -106,13 +106,19 @@ let read_dir t =
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "reading directory %a" pp t
 
-let walk t fn =
+let with_dir_entries t fn =
   let (Resource.T (dir, ops), path) = t in
   let module X = (val (Resource.get ops Fs.Pi.Dir)) in
-  try X.with_dir_entries dir path fn
+  (* Note: we don't handle exceptions here as that would include the user's [fn]
+     (and this is likely to be called recursively). *)
+  X.with_dir_entries dir path fn
+
+let read_dir_entries t =
+  try
+    with_dir_entries t List.of_seq |> List.sort (fun (_, a) (_, b) -> String.compare a b)
   with Exn.Io _ as ex ->
     let bt = Printexc.get_raw_backtrace () in
-    Exn.reraise_with_context ex bt "walking directory %a" pp t
+    Exn.reraise_with_context ex bt "reading directory %a" pp t
 
 let stat ~follow t =
   let (Resource.T (dir, ops), path) = t in
--- a/lib_eio/path.mli
+++ b/lib_eio/path.mli
@@ -155,6 +155,9 @@ val read_dir : _ t -> string list
 
     Note: The special Unix entries "." and ".." are not included in the results. *)
 
+val walk : _ t -> ((File.Stat.kind * string) Seq.t -> 'a) -> 'a
+(** [walk t] traverses the directory [t] producing a sequence of results. *)
+
 (** {1 Metadata} *)
 
 val stat : follow:bool -> _ t -> File.Stat.t
--- a/lib_eio/path.mli
+++ b/lib_eio/path.mli
@@ -149,14 +149,23 @@ val with_open_dir : _ t -> ([< `Close | dir_ty] t -> 'a) -> 'a
     it automatically when [fn] returns (if it hasn't already been closed by then). *)
 
 val read_dir : _ t -> string list
-(** [read_dir t] reads directory entries for [t].
+(** [read_dir t] reads directory entry names for [t].
 
     The entries are sorted using {! String.compare}.
 
     Note: The special Unix entries "." and ".." are not included in the results. *)
 
-val walk : _ t -> ((File.Stat.kind * string) Seq.t -> 'a) -> 'a
-(** [walk t] traverses the directory [t] producing a sequence of results. *)
+val read_dir_entries : _ t -> (File.Stat.kind * string) list
+(** [read_dir_entries] is like {!read_dir}, but also includes the kind of each entry.
+
+    This is typically much faster than calling {!kind} on each entry individually. *)
+
+val with_dir_entries : _ t -> ((File.Stat.kind * string) Seq.t -> 'a) -> 'a
+(** [with_dir_entries t fn] runs [fn items], where [items] is the entries as a sequence.
+
+    This is like {!read_dir_entries}, but loads the entries incrementally,
+    which may be more efficient. Unlike {!read_dir_entries}, it does not sort
+    the results, which may be returned in any order. *)
 
 (** {1 Metadata} *)
 
--- a/lib_eio/unix/eio_unix_stubs.c
+++ b/lib_eio/unix/eio_unix_stubs.c
@@ -3,11 +3,15 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <errno.h>
+#include <dirent.h>
 
 #include <caml/mlvalues.h>
 #include <caml/unixsupport.h>
 #include <caml/memory.h>
 #include <caml/bigarray.h>
+#include <caml/alloc.h>
+
+#define BUF_SIZE 4096
 
 static void caml_stat_free_preserving_errno(void *ptr) {
   int saved = errno;
@@ -38,7 +42,7 @@ CAMLprim value eio_unix_readlinkat(value v_fd, value v_path, value v_cs) {
   value v_ba = Field(v_cs, 0);
   value v_off = Field(v_cs, 1);
   value v_len = Field(v_cs, 2);
-  char *buf = (char *)Caml_ba_data_val(v_ba) + Long_val(v_off); 
+  char *buf = (char *)Caml_ba_data_val(v_ba) + Long_val(v_off);
   size_t buf_size = Long_val(v_len);
   int fd = Int_val(v_fd);
   int ret;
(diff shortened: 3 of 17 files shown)


patricoferris:
Thanks @avsm !

Note that the current implementation does _not_ descend into directories (see the recursive call to `walk` in the benchmark above). Happy to change the behaviour to recurse by default.

avsm:
Got it! Your approach makes more sense to me. We can build a recursive descender _over_ this interface in the future.

avsm:
Just chatting to @patricoferris, I think this version that _doesnt_ recurse is best, since we can then add a parallel or sequential recursing function over this.

lib_eio/unix/eio_unix_stubs.c
```diff
@@ -3,11 +3,15 @@
...
 #include <errno.h>
+#include <dirent.h>
 
 #include <caml/mlvalues.h>
 #include <caml/unixsupport.h>
 #include <caml/memory.h>
 #include <caml/bigarray.h>
+#include <caml/alloc.h>
+
+#define BUF_SIZE 4096
```
talex5: This doesn't seem to be used.

lib_eio/path.ml
```diff
@@ -106,6 +106,14 @@ let read_dir t =
     let bt = Printexc.get_raw_backtrace () in
     Exn.reraise_with_context ex bt "reading directory %a" pp t
 
+let walk t fn =
+  let (Resource.T (dir, ops), path) = t in
+  let module X = (val (Resource.get ops Fs.Pi.Dir)) in
+  try X.with_dir_entries dir path fn
+  with Exn.Io _ as ex ->
+    let bt = Printexc.get_raw_backtrace () in
+    Exn.reraise_with_context ex bt "walking directory %a" pp t
```
talex5: Not sure adding the exception handler here is a good idea. This is likely to be called recursively, and we don't want all the parents adding their own error context. Probably the backends should wrap just the actual call.

lib_eio/path.mli
```diff
@@ -155,6 +155,9 @@ val read_dir : _ t -> string list
 
     Note: The special Unix entries "." and ".." are not included in the results. *)
 
+val walk : _ t -> ((File.Stat.kind * string) Seq.t -> 'a) -> 'a
+(** [walk t] traverses the directory [t] producing a sequence of results. *)
```
talex5: Comment needs revising (it doesn't necessarily produce a sequence of results). Should say something about ordering here too.

lib_eio_linux/low_level.ml
```diff
@@ -513,11 +513,17 @@ let read_dir fd =
...
+      let files = List.filter_map (function (_, "..") | (_, ".") -> None | (_, f) -> Some f) files in
       read_all (files @ acc) fd
   in
   Eio_unix.run_in_systhread ~label:"read_dir" (fun () -> read_all [] fd)
 
+let read_some_dir fd =
+  Fd.use_exn "read_some_dir" fd @@ fun fd ->
+  Eio_unix.run_in_systhread ~label:"read_some_dir" @@ fun () ->
+  let files = eio_getdents fd in
+  List.filter_map (function _, ".." | _, "." -> None | v -> Some v) files
```
talex5:
```suggestion
  List.filter (function _, ".." | _, "." -> false | _ -> true) files
```
Maybe we should just filter these out in the C code, though?

lib_eio_posix/fs.ml
```diff
@@ -86,6 +86,9 @@ end = struct
     Err.run (Low_level.readdir t.fd) path
     |> Array.to_list
 
+  let with_dir_entries t path fn =
+    Err.run (Low_level.with_dir_entries t.fd path) fn
```
talex5: Probably don't want to wrap `fn` as well here, just the actual read.

lib_eio_posix/low_level.ml
```diff
@@ -389,6 +390,44 @@ let readdir dirfd path =
     Fd.use_exn "readdir" dirfd @@ fun dirfd ->
     use_confined (Some dirfd)
 
+let with_dir_entries dirfd path fn =
```
talex5: Most of this is duplicated from `readdir`. Could perhaps factor it out as `with_dir_handle` or similar?

lib_eio_posix/low_level.ml
```diff
@@ -389,6 +390,44 @@ let readdir dirfd path =
...
+  let rec read_entry h =
+    match eio_readdir h with _, "." | _, ".." -> read_entry h | v -> v
+  in
+  let it h () =
+    match in_worker_thread "with_dir_entries" @@ fun () -> read_entry h with
```
talex5: Seems a shame to use a thread job for every entry individually. I'd guess that reading 1000 at a time might speed things up a fair bit. The only problem I see is if you had a really slow network file-system and wanted to get the first entries without waiting.

lib_eio_posix/low_level.ml
```diff
@@ -389,6 +390,44 @@ let readdir dirfd path =
...
+    | exception End_of_file -> None
+    | exception ex ->
+        let bt = Printexc.get_raw_backtrace () in
+        Unix.closedir h;
+        Printexc.raise_with_backtrace ex bt
+  in
+  let use h =
+    let seq = Seq.of_dispenser (it h) in
+    let v = fn seq in
+    Unix.closedir h;
```
talex5, line 394: If `fn` raises, `h` doesn't get closed. Would probably be simpler to do all the closing in one place.

talex5 (commented):
Thanks - this looks really good!

I agree we should start with this non-recursive version. Might be good to rename it though (Python uses `walk` for the recursive version: https://docs.python.org/3/library/os.html#os.walk).

(also, you need to install `lintcstubs_arity` to get `primitives.h` to update)

talex5:
I've rebased and updated this as follows:

- Don't wrap `walk` with an exception handler, as that would include the user's callback.
- Improve documentation for `walk`.
- Add a simple test for it.
- Rename it as `with_dir_entries`.
- Add `read_dir_entries` helper that still does sorting and error reporting.

eio_linux:
- Filter out `.` and `..` earlier.

eio_posix:
- Handle errors from `readdir`.
- Open directory in a worker thread too.
- Share opening logic.
- Ensure handle is always closed.
- Don't wrap user's function in our error handling.
- Batch reading directory entries (much faster).

eio_windows:
- Add (very unoptimised) implementation.
- Fix eio/unix stubs build on Windows.

@patricoferris if you're happy with this, I think it's ready to squash and merge once CI is done.

lib_eio_posix/low_level.ml
talex5: On a test I did (recursively scanning `/usr`), using a batch size of 100 reduced the total time from 3.14s to 1.04s. Increasing the batch size to 1000 didn't give any further improvement.

lib_eio/unix/eio_unix_stubs.c
```diff
@@ -52,3 +54,21 @@ CAMLprim value eio_unix_readlinkat(value v_fd, value v_path, value v_cs) {
   CAMLreturn(Val_int(ret));
   #endif
 }
+
+CAMLprim value eio_unix_file_type_of_dtype (int d_type) {
+  #ifdef _WIN32
+  return caml_hash_variant("Unknown");
+  #else
+  switch (d_type) {
+    case DT_REG: return caml_hash_variant("Regular_file");
```
avsm, line 63: caml_hash_variant involves computation, so I think this'll happen every time the function is called. Is it worth assigning these to static variables so this conversion is faster? https://github.com/ocaml/ocaml/blob/b5ec56d7afe77b4e57df4ce369c1281a92c6303d/runtime/hash.c#L313

avsm:
Read rest of diff, lgtm

lib_eio/unix/eio_unix_stubs.c
talex5: I don't mind, but I tried it in godbolt and gcc and clang both evaluated it at compile time:

https://godbolt.org/z/hcMnnKzo4

avsm: Wow, that's impressive (and good enough for me!)

talex5: Oh, except it doesn't work in the actual code (because `caml_hash_variant` is defined in a .c file, not a header, I think). I'll expand it.

talex5: It's actually not obvious how to do this. C doesn't seem to have a way to force compile-time evaluation, even though real compilers do it anyway, so it won't let you use `caml_hash_variant` in a static initialiser.

We could get `discover.ml` to write a header file with the values, but getting the value from OCaml isn't easy.
In https://github.com/ocaml/ocaml/issues/4751, the compiler devs say:

> You can already do it for fixed strings by (Obj.magic `Variant : int).
> You can also copy the corresponding function from ocaml sources
> (it is supposed to be fixed forever).

Seems messy, though.

(also, we now seem to have three functions like this: in eio/unix, eio_posix and uring)

talex5:
Since the `caml_hash_variant` problem was also there before this PR, let's merge this as it is and optimise the hashing separately. It's strictly an improvement on the previous system of calling `kind` separately for each entry.

talex5:
I tried pre-calculating the hashes here: https://github.com/talex5/eio/commit/575f0b923ab62378eee4030a4da8318595bef22e

However, it didn't make it noticeably faster, so probably not worth it.

Issue #757, opened 2024-10-01
Crash on writing to /dev/null on macOS


aaronjeline:
Attempting to use `Eio.Flow.copy_string` when stdout points to `/dev/null` crashes.
`main.ml`:
```ocaml
open Eio

let main env =
    let stdout = Stdenv.stdout env in
    Eio.Flow.copy_string "Hello world!" stdout;
    ()

let () = Eio_main.run main
```
`dune`:
```
(executable
 (public_name demo)
 (name main)
 (libraries eio_main demo))
```

Interacting with it:
```bash
% dune exec demo
Hello world!
% dune exec demo > out
% cat out -p
Hello world!
% dune exec demo > /dev/null
Fatal error: exception File "lib_eio_posix/sched.ml", line 155, characters 2-8: Assertion failed

Eio version: 1.1

aaronjeline:
Looks like this is an issue with MacOS `poll()`, doesn't seem like it supports block devices. (https://nathancraddock.com/blog/macos-dev-tty-polling/). 
I'm happy to work on supporting this, I see two ways of going about it, not sure which you think is best?

1) Special case the handling of block devices, and use `select` for them.
2) Spawn a new thread for monitoring blocks devices, use `select` in that thread, and pipe data from that to the main thread, monitor that second thread as normal with `poll`

If you have better ideas I'm also happy to hear that.

aaronjeline:
> sorted in Leopard
lol.
Like I said, happy to hack on this if you'd like me to

aaronjeline:
Blog post on the trick `uv` does, I'll start down this road: https://code.saghul.net/2016/05/libuv-internals-the-osx-select2-trick/

talex5:
Sounds good!

talex5:
Fixed in #893.

PR #827, opened 2026-01-08
Support writing to block devices on MacOS

aaronjeline:
If we get `POLLNVAL`, wake all of our readers/writers. MacOS will report `POLLNVAL` for any block device. Solves #757

### DIFF
--- a/lib_eio_posix/sched.ml
+++ b/lib_eio_posix/sched.ml
@@ -21,20 +21,10 @@ module Fiber_context = Eio.Private.Fiber_context
 module Trace = Eio.Private.Trace
 module Rcfd = Eio_unix.Private.Rcfd
 module Poll = Iomux.Poll
+module Eventfd = Posix_eventfd
+open Sched_common
 
 type exit = [`Exit_scheduler]
-
-(* The type of items in the run queue. *)
-type runnable =
-  | IO : runnable                                       (* Reminder to check for IO *)
-  | Thread : 'a Suspended.t * 'a -> runnable            (* Resume a fiber with a result value *)
-  | Failed_thread : 'a Suspended.t * exn -> runnable    (* Resume a fiber with an exception *)
-
-(* For each FD we track which fibers are waiting for it to become readable/writeable. *)
-type fd_event_waiters = {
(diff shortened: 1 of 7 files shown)


lib_eio_posix/sched.ml
```diff
@@ -152,16 +152,26 @@ let resume t node =
 
 (* Called when poll indicates that an event we requested for [fd] is ready. *)
 let ready t _index fd revents =
-  assert (not Poll.Flags.(mem revents pollnval));
   if fd == t.eventfd_r then (
     clear_event_fd t
     (* The scheduler will now look at the run queue again and notice any new items. *)
   ) else (
+    (* Watiters is the set of computations waiting on this fd, split into readers and writers *)
```
talex5:
```suggestion
    (* Waiters is the set of computations waiting on this fd, split into readers and writers *)
```

talex5 (commented):
I assume this means that, if the resource isn't ready, then it will spin? That's not a problem for `/dev/null` but it might be for other things.

Possibly that's still better than the current behaviour, but needs a comment in the code that it should be improved if it turns out to be a problem.

aaronjeline:
> but needs a comment in the code that it should be improved if it turns out to be a problem.

 Added a note about this.

talex5:
> This results in us blocking on block devices

Does it block? I'd expect it to keep trying at 100% CPU instead (since we normally open things as non-blocking, it should get EAGAIN and try to wait again).

I wonder if we should just use `select` on macOS, since apparently you can avoid the 1024 FD limit there (with `_DARWIN_UNLIMITED_SELECT`), and I doubt people are using macOS to run e.g. high-performance servers with huge numbers of connections anyway.

Any other mac users have thoughts about this?

aaronjeline:
> I wonder if we should just use select on macOS

Do you mean just switch from `poll()` to `select()` for all files, or keep a separate thread up for monitoring non-pollable files with `select`.

I believe (2) is what libuv does

talex5:
Either would be fine I think.

aaronjeline:
Latest commits spawns a new thread to handle block devices on macOS. Let me know if you want anything organized differently. Happy to adjust as you'd like.

PR #832, opened 2026-02-27
Simplify custom "env" type in example

vog:
Since we are already using the "constraint 'a =" syntax, the "as 'a" at the end is not needed, and even confusing.


talex5 (approved):
Thanks for the clean up!

PR #840, opened 2026-05-20
eio_linux: use newer uring flags for performance

talex5:
This shows a large speed improvement on the HTTP benchmark on my machine (will be interesting to see what the benchmark CI says).

- `IORING_SETUP_DEFER_TASKRUN` only does completion work as part of `io_uring_enter`. This avoids interrupting the application. The uring docs say "This flag should be considered the default mode for applications setting up a ring"

- `IORING_SETUP_SINGLE_ISSUER` hints to uring to do some locking optimisations assuming that each ring is only written by one thread. Only a small improvement in the HTTP tests on my machine, but needed for `IORING_SETUP_DEFER_TASKRUN`.

`IORING_SETUP_COOP_TASKRUN` showed similar improvements, but preferring `DEFER_TASKRUN` as the docs say it's a good default.

### DIFF
--- a/lib_eio_linux/sched.ml
+++ b/lib_eio_linux/sched.ml
@@ -510,9 +510,15 @@ let with_eventfd fn =
     close ();
     Printexc.raise_with_backtrace ex bt
 
+let uring_create ~queue_depth ?polling_timeout () =
+  let flags = Uring.Setup_flags.(single_issuer + coop_taskrun) in (* Requires Linux >= 6.0 *)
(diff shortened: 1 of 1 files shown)


talex5:
Ah. The benchmark machine is still running Linux 5.15 and that's too old to use these flags.

talex5:
Closing in favour of #842. It's a separate PR mainly to avoid confusion in the benchmark results, as the benchmark machine got updated after this one was opened.

PR #843, opened 2026-05-27
eio_posix: use `MSG_CMSG_CLOEXEC` when receiving file descriptors

talex5:
(eio_linux was fixed by https://github.com/ocaml-multicore/ocaml-uring/pull/132)


lib_eio_posix/eio_posix_stubs.c
```diff
@@ -535,7 +535,7 @@ CAMLprim value caml_eio_posix_recv_msg(value v_fd, value v_max_fds, value v_bufs
   msg.msg_iov = iov;
 
   caml_enter_blocking_section();
-  r = recvmsg(Int_val(v_fd), &msg, 0);
+  r = recvmsg(Int_val(v_fd), &msg, MSG_CMSG_CLOEXEC);
```
avsm: This won't work on macOS; this needs to be Eio_linux specific.

talex5: It's in POSIX: https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/sys_socket.h.html

> The <sys/socket.h> header shall define the following symbolic constants with distinct values for use as the valid values for the msg_flags field in the msghdr structure, or the flags parameter in [recv()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/recv.html), [recvfrom()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/recvfrom.html), [recvmsg()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/recvmsg.html), [send()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/send.html), [sendmsg()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sendmsg.html), or [sendto()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sendto.html) calls:
>
> MSG_CMSG_CLOEXEC
>    Atomically set the FD_CLOEXEC flag on any file descriptors created via SCM_RIGHTS during [recvmsg()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/recvmsg.html).

FreeBSD, NetBSD and OpenBSD all have it. But it does indeed seem to be missing from macos. I'll make it conditional.

avsm: Indeed, I'd missed its in POSIX now, but only under the "Implementations **may** provide additional socket creation flags". So this PR looks right to me now with the conditional.

talex5: (I think that's referring to the flags above it (e.g. `SOCK_NONBLOCK`). `MSG_CMSG_CLOEXEC` isn't a socket creation flag.)

PR #848, opened 2026-06-04
Add Eio_linux.Low_level.Fixed

talex5:
Previously, Eio used the `Uring.Region` API. However, it probably makes more sense to handle regions inside Eio (possibly `Region` should be removed from uring completely).

This is an API change for anyone using fixed buffers.

A minor temporary benefit of this change is that it makes Eio compatible with the current Git version of uring too.

It might allow some future simplifications. For example, `Uring.Region` maintains a free pool, but we could use `Eio.Pool` instead and get the waiting logic for free.

@avsm - what do you think?


avsm:
I'm in favour, because I was about to suggest the same change as part of a move towards bytes! Eio itself needs to manage the pool of buffers (including tracking if they are page aligned, fixed, or >1KB to be immovable) to supply to uring.

PR #849, opened 2026-06-05
eio_linux: move read/write retries out of scheduler

talex5:
The goal here is to simplify the scheduler by removing some special cases that are only used for fixed-buffer reads and writes.

Before, a fiber could ask to read exactly n bytes, and the scheduler would keep issuing reads until that many had been read. It would also resubmit on EINTR, etc.

Now, the scheduler simply returns the number of bytes read or written and lets the fiber make another call if it wants more.

This adds a couple of fiber switches and delays the issuing of the next request. However, fiber switches are fast and if the kernel does a short read or write then it's because it's not ready to do more yet anyway.

It simplifies the code in several ways:

- The scheduler no longer needs to know about fixed-buffer operations.

- The `Read` and `Write` job types are gone; the regular `Job` works fine.

- We no longer need to distinguish between "read at offset 0" and "read on non-seekable FD", since the action doesn't need to update it. This avoids the need for the `file_offset` type.

- The `amount` type is gone too.

### DIFF
--- a/lib_eio_linux/low_level.ml
+++ b/lib_eio_linux/low_level.ml
@@ -60,13 +60,19 @@ let uring_file_offset t =
   if Fd.is_seekable t then Optint.Int63.minus_one else Optint.Int63.zero
 
 let file_offset t = function
-  | Some x -> `Pos x
-  | None when Fd.is_seekable t -> `Seekable_current
-  | None -> `Nonseekable_current
+  | Some x -> x
+  | None when Fd.is_seekable t -> Optint.Int63.minus_one
+  | None -> Optint.Int63.zero
 
-let enqueue_read st action (file_offset,fd,buf,len) =
-  let req = { Sched.op=`R; file_offset; len; fd; cur_off = 0; buf; action } in
-  Sched.submit_rw_req st req
+let rec enqueue_read st action req =
+  let retry = Sched.with_cancel_hook ~action st (fun () ->
+      let (file_offset,fd,buf,len) = req in
+      let off = Fixed.to_offset buf in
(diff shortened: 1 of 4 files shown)


lib_eio_linux/low_level.ml
```diff
@@ -165,19 +177,27 @@ let sleep_until time =
...
 
-let read ?file_offset:off fd buf amount =
-  let off = file_offset fd off in
-  Fd.use_exn "read" fd @@ fun fd ->
-  let res = Sched.enter "read" (fun t k -> enqueue_read t k (off, fd, buf, amount)) in
-  if res < 0 then (
-    raise @@ Err.wrap (Uring.error_of_errno res) "read" ""
-  ) else res
+(* TODO bind from unixsupport *)
+let errno_is_retry = function -62 | -11 | -4 -> true |_ -> false
```
avsm, line 181: could expose errnos from the probing in uring too, will take a look

talex5: Yes, could be useful (but this change is just existing code being moved from `sched.ml`).

avsm:
> if the kernel does a short read or write then it's because it's not ready to do more yet anyway

I agree that this is a neat heuristic to fibre switch on!

talex5:
> > if the kernel does a short read or write then it's because it's not ready to do more yet anyway
> I agree that this is a neat heuristic to fibre switch on!

Actually, I think I was wrong about this. It's using `Enter`, not `Suspend`, so `schedule` will immediately resume the application fiber, which will immediately re-enter the scheduler and reissue the request. So it should behave the same as before.

PR #853, opened 2026-06-10
Rename Path.open_dir to open_subtree

talex5:
This more accurately describes what it's for (every time I show some code with the old names, I feel I have to explain what it does).

### DIFF
--- a/README.md
+++ b/README.md
@@ -809,12 +809,12 @@ The checks also apply to following symlinks:
 - : unit = ()
 ```
 
-You can use `open_dir` (or `with_open_dir`) to create a restricted capability to a subdirectory:
+You can use `open_subtree` (or `with_subtree`) to create a restricted capability to a subdirectory:
 
 ```ocaml
 # Eio_main.run @@ fun env ->
   let cwd = Eio.Stdenv.cwd env in
-  Eio.Path.with_open_dir (cwd / "dir1") @@ fun dir1 ->
+  Eio.Path.with_subtree (cwd / "dir1") @@ fun dir1 ->
   try_save (dir1 / "file4") "D";
   try_save (dir1 / "../file5") "E";;
 +save <dir1:file4> : ok
@@ -822,13 +822,13 @@ You can use `open_dir` (or `with_open_dir`) to create a restricted capability to
 - : unit = ()
 ```
 
-You only need to use `open_dir` if you want to create a new sandboxed environment.
+You only need to use `subtree` if you want to create a new sandboxed environment.
 You can use a single base directory object to access all paths beneath it,
 and this allows following symlinks within that subtree.
 
 A program that operates on the current directory will probably want to use `cwd`,
 whereas a program that accepts a path from the user will probably want to use `fs`,
-perhaps with `open_dir` to constrain all access to be within that directory.
+perhaps with `open_subtree` to constrain all access to be within that directory.
 
 On systems that provide the [cap_enter][] system call, you can ask the OS to reject accesses
 that don't use capabilities.
--- a/examples/capsicum/main.ml
+++ b/examples/capsicum/main.ml
@@ -28,7 +28,7 @@ let () =
   in 
   if not (Eio.Path.is_directory path) then Fmt.failwith "%a is not a directory" Eio.Path.pp path;
   (* Get access to resources before calling cap_enter: *)
-  Eio.Path.with_open_dir path @@ fun dir ->
+  Eio.Path.with_subtree path @@ fun dir ->
   traceln "Opened directory %a" Eio.Path.pp path;
   (* Switch to capability mode, if possible: *)
   begin match Eio_unix.Cap.enter () with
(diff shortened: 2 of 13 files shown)


avsm:
I'm not sure the new name's entirely clearer... isn't this just a doc problem with `open_dir`?

talex5:
The docs said "This can be passed to functions to grant access only to the subtree [t]", which is fine. But `open_dir` doesn't suggest anything about confining to subtrees to me, and sounds more like an API for listing directories.

avsm:
I don't understand how renaming will help to clarify the user confusion, though. There's no mechanism to open a directory _without_ this restriction is there? Perhaps if we show both side by side in the Eio.Path docs it would be easier for a user to figure out which function to use.

patricoferris:
This is like forgetting the `open_dir` here right: https://github.com/mirage/ocaml-tar/pull/177/changes#diff-c00a6ad125fbfa55e6c8e3467eabe832056f84cb8e60846558f7c1a152ac767dR165.

It is very easy to just run with user provided `_ Eio.Path.t`s and not "open the subtree" where you might need it. I eluded to some "Eio for Systems Programmers" docs which might help with this sort of thing maybe?

avsm:
Yes this line 165 is insidiously easy to miss! https://github.com/mirage/ocaml-tar/pull/177/changes#diff-c00a6ad125fbfa55e6c8e3467eabe832056f84cb8e60846558f7c1a152ac767dL157-R165

That wouldn't be addressed by renaming the function, but it does need to be addressed ;-)

talex5:
@patricoferris yes, that's a good example.

If someone sees `with_open_dir` then they might wonder why it needs to be opened, try it without, and find it seems to work. Whereas I think `with_subtree` makes it more obvious that this is about confinement.

(and in the other direction, if you see `with_subtree` then it's easy to remember that it's confined and you don't need to check further, whereas `with_open_dir` will likely need reading the docs to double-check)

avsm:
The Linux uses `RESOLVE_BENEATH`; how about `open_beneath[_dir]`? I think I dislike that we dont distinguish between "open a directory that's a subtree" vs "open a directory _and constrain future lookups_ to that subtree".

talex5:
The problem is that Linux doesn't have a flag to open a directory that can only be used as a subtree. Opening a directory with `RESOLVE_BENEATH` controls whether you can open the directory in the first place, not what you can do with it afterwards.

Also, note that this function isn't really giving you a directory, it's giving you another `Path.t` (which happens to contain a directory).

avsm:
> Also, note that this function isn't really giving you a directory, it's giving you another Path.t (which happens to contain a directory).

Isn't this the root of the problem (pun intended)? A directory that's a root should probably have a different type to one that's traversable 'out of it'.

talex5:
> A directory that's a root should probably have a different type to one that's traversable 'out of it'.

Why? What would you do with that information?

Ultimately, every path is limited as to how far you can go up from it (though Unix has a special case for `/`, allowing you to do `/..` for some reason). You could encode "directory that lets you go up at most three levels" in the type, but you wouldn't be able to maintain the type when combining it with user-provided strings or symlinks.

e.g. if `t` doesn't allow going up then you can't tell whether `t / "foo"` does, since `foo` could be a symlink.

But anyway, the point is that `open_dir` is badly named because it isn't about opening directories (that's just an implementation detail), it's about restricting file access to a subtree.

avsm:
> Why? What would you do with that information?

In the tar example above, the user intention is to open the tar output directory as a root so it could never be escaped. If the type had reflected that, code review would have spotted that it's just a normal path.

I'm not particularly bothered about the renaming here (for or against), except for intensely disliking deprecation warnings in existing code that is perfectly fine except for its name. Could we just make this a non-deprecated alias and quietly remove it in a future release?

talex5:
> In the tar example above, the user intention is to open the tar output directory as a root so it could never be escaped. If the type had reflected that, code review would have spotted that it's just a normal path.

`Path.t` is defined as `type 'a t = 'a Fs.dir * path`, and tar-eio could have taken just a `Fs.dir` to force the issue. But I wouldn't recommend it in general - in most cases it's reasonable to allow the use of symlinks.

I also think any reasonably security-conscious user of tar would restrict it themselves. When doing a security audit you'll start at `Eio_main.run` and track where `env#fs` goes. If you see it going into a tar extractor with untrusted input then that should be a big warning sign.

(I do wonder if `Path.(/)` should be renamed to `Path.extend_unconfined` or something. Seeing `let (/) = Path.extend_unconfined` at the top of a module would make things a bit clearer.)

> Could we just make this a non-deprecated alias and quietly remove it in a future release?

Sure. I was undecided myself about whether to deprecate now or later - I'll remove the attribute.

avsm:
I hadn't considered using Fs.dir directly! There's a reasonable case that symlinks should be resolved before extracting a tar, I think. Diff looks good to me with the deprecation removed!

PR #854, opened 2026-06-13
Propagate errno back from fork_actions into Eio.Process

### LINKED ISSUE #556: "execve: Argument list too long" should not be a `Failure`

avsm:
The idea follows @talex5's review of #561 by @Innf107.  When a fork action fails, the child now reports the errno rather than the strerror. The parent decodes it and raises a Unix.Unix_error instead of a Failure (7f01035a2df0c29d1c0b42a66f61e36d4f81ae7c)

The subsequent commits then add a Eio.Process.Argument_list_too_long (fixes #556 and #561) from E2BIG, and we also map ENOENT/EACCES/ENOEXEC to new Eio.Process errors (d42c0ef8dec3b145450e562970418d3798e06f64).

### DIFF
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -28,11 +28,11 @@ let () =
     | E e ->
       Fmt.string f "Process ";
       begin match e with
-        | Executable_not_found e -> Fmt.pf f "Executable %S not found" e;
+        | Executable_not_found e -> Fmt.pf f "Executable_not_found %S" e;
         | Child_error e -> Fmt.pf f "Child_error %a" pp_status e;
-        | Argument_list_too_long -> Fmt.pf f "Argument list too long"
-        | Permission_denied e -> Fmt.pf f "Permission denied when executing %S" e
-        | Executable_format_error e -> Fmt.pf f "Executable %S has an invalid format" e
+        | Argument_list_too_long -> Fmt.pf f "Argument_list_too_long"
+        | Permission_denied e -> Fmt.pf f "Permission_denied %S" e
+        | Executable_format_error e -> Fmt.pf f "Executable_format_error %S" e
       end;
       true
     | _ -> false
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -15,6 +15,9 @@ let pp_status ppf = function
 type error =
   | Executable_not_found of string
   | Child_error of exit_status
+  | Argument_list_too_long
+  | Permission_denied of string
+  | Executable_format_error of string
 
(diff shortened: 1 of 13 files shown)


avsm:
Also layered pty support in #531 over this PR

lib_eio/process.ml
```diff
@@ -27,6 +30,9 @@ let () =
       begin match e with
         | Executable_not_found e -> Fmt.pf f "Executable %S not found" e;
         | Child_error e -> Fmt.pf f "Child_error %a" pp_status e;
+        | Argument_list_too_long -> Fmt.pf f "Argument list too long"
+        | Permission_denied e -> Fmt.pf f "Permission denied when executing %S" e
+        | Executable_format_error e -> Fmt.pf f "Executable %S has an invalid format" e
```
talex5: The intention for `register_pp` was to print the variant name in the error so you can you add a handler for it easily (but `Executable_not_found` wasn't doing this and that pattern got copied for the others).
```suggestion
        | Executable_not_found e -> Fmt.pf f "Executable_not_found %S" e;
        | Child_error e -> Fmt.pf f "Child_error %a" pp_status e;
        | Argument_list_too_long -> Fmt.pf f "Argument_list_too_long"
        | Permission_denied e -> Fmt.pf f "Permission_denied %S" e
        | Executable_format_error e -> Fmt.pf f "Executable_format_error %S" e
```

talex5 (approved):
Looks good. I pushed another commit to simplify the tests slightly (might as well take advantage of the new exceptions!).

Not sure about the error tests though. e.g. I don't think the kernel is *required* to reject long arguments, even if most do, and we can't be sure an executable is a bad format.

lib_eio/process.ml
avsm: That makes sensE!

avsm:
It is true that a binfmt_misc might conceivably be registered and not trigger the bad executable format, but it seems unlikely. These tests all need to run on a reasonably vanilla kernel, I think. I'll see what happens in Windows though...

PR #855, opened 2026-06-15
eio_posix: open confined paths in non-blocking mode too

talex5:
Previously, paths opened using `fs` were opened in non-blocking mode, but paths opened with `cwd` were not.


avsm:
Clearly correct to match behaviour, although I'm not sure that nonblocking has any practical effect when file-backed fds are involved.

talex5:
Yes; on Linux, the man-page says:
> Note that this flag has no effect for regular files and block devices

but also that it "might eventually be implemented".

However, it does matter for FIFOs.

avsm:
I hadn't considered FIFOs!

PR #857, opened 2026-06-15
eio_posix: wait for a writer when first reading from a FIFO

### LINKED ISSUE #856: eio_posix: fifos don't work properly
The blocking behaviour of fifos (named pipes) on Unix is quite interesting.

When opening for reading in blocking mode, it waits until a writer is available:
```ocaml
$ mkfifo fifo
$ utop
utop # #require "unix";;
utop # let r = Unix.openfile "./fifo" [O_RDONLY] 0;;
[ waits ]
```

Once a writer has opened it, the reader's open completes (even though there's no data to be read yet). Reading then blocks until there's data available, as expected.

If opened with `O_NONBLOCK`, the open completes immediately even without a writer:

```ocaml
utop # let r = Unix.openfile "./fifo" [O_RDONLY; O_NONBLOCK] 0;;
val r : Unix.file_descr = <abstr>
```
`select` reports the FD as not ready, but attempting to read from it immediately returns end-of-file:
```ocaml
utop # Unix.select [r] [] [] 0.;;
- : Unix.file_descr list * Unix.file_descr list * Unix.file_descr list =
([], [], [])

utop # Unix.read r (Bytes.create 10) 0 10;;
- : int = 0
```

If there is a writer (but no data), then it's still not reported as ready, but reading returns `EAGAIN`. If there is data ready, `select` reports it as ready and you can read it, as expected.

If a writer opens the FIFO and then closes it without writing anything, then `select` starts reporting the reading end as ready.

So, there seems to be a special case: `select` returns non-ready before the first writer connects.

For Eio, I think we need to open in non-blocking mode, because opening can take any amount of time and we don't want to block the whole scheduler (and we'd need to be able to cancel, even if it ran in another thread).

eio_posix tries to do a read directly and only suspends the fiber if it gets `EAGAIN`. This doesn't work with fifos because it gets end-of-file instead.

So, to mimic the blocking behaviour, I think that eio_posix needs to notice that you've opened a FIFO for reading and wait for it to become readable before resuming the opening fiber.

Opening for writing is also troublesome. Normally, the open will wait for a reader, but with `O_NONBLOCK` it instead returns an error. There seems to be a Linux-specific workaround of opening for reading and writing.

On eio_linux, the open succeeds immediately, but the first read waits until data is available.

talex5:
This matches the ordinary blocking behaviour.

Without this, eio_posix would immediately return end-of-file.

Curiously, eio_linux (uring) had the open return immediately, but waited on the first read. Possibly that's a bug. I've added an explicit wait there too.

Fixes #856.

### DIFF
--- a/lib_eio_posix/fs.ml
+++ b/lib_eio_posix/fs.ml
@@ -32,6 +32,26 @@ let as_posix_dir (Eio.Resource.T (t, ops)) =
   | None -> None
   | Some fn -> Some (fn t)
 
+module Fifo_reader = struct
+  include Flow.Impl
+
+  let single_read t buf =
+    Low_level.await_readable "single_read" t;
+    Flow.Impl.single_read t buf
+
+  let pread t ~file_offset bufs =
+    Low_level.await_readable "pread" t;
+    Flow.Impl.pread t ~file_offset bufs
+end
+
+(* After opening a FIFO with no writers, it is not readable according to select,
+   but [read] returns end-of-file rather than EAGAIN!
+   Therefore, use a slightly slower implementation that always checks readability first.
+   See https://github.com/ocaml-multicore/eio/issues/856 *)
+let fifo_reader =
+  let handler = Eio_unix.Pi.flow_handler (module Fifo_reader) in
+  fun fd -> (Eio.Resource.T (fd, handler) :> Eio.File.ro_ty Eio.Resource.t)
+
 module rec Dir : sig
   include Eio.Fs.Pi.DIR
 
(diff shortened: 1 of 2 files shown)


talex5:
Actually, I think it's not possible to match the blocking behaviour. Ideally, we want the `open_in` to return once a writer is connected, even though no data is available. But we can only wait for data to become available.

(i.e. we want to say "wait until `read` would start returning EAGAIN", which is the opposite of what `select` does)

Also, it would be nice to be able to open a file, see that it's a FIFO, and then close it immediately. Waiting on open prevents that.

talex5:
I've pushed a new approach which leaves the eio_linux behaviour as it was and makes eio_posix match it by always waiting for the fifo to become readable before trying to read it. With this, opening a fifo always completes immediately, without waiting for a writer to connect.

However, there's another problem: on macos it seems that you don't get a kevent when there are no more writers, so it will just hang (see e.g. https://github.com/golang/go/issues/63937). Probably we should switch to using `select` on macos (apparently you can avoid the 1024 FD limit there with `_DARWIN_UNLIMITED_SELECT`). macos kevent seems to be very buggy.

For now, I've changed the test so it doesn't wait for EOF, but that should be fixed separately.

PR #858, opened 2026-06-23
Initial support for setting/getting socket options

talex5:
This adds a `Eio.Net.Sockopt.t` extensible variant for socket options.

This is based on @avsm's #575. To make review easier, I removed all the portable options. I'll rebase that PR on this one next. Other changes from #575:

- In `Eio.Net`, `Sockopt` is the type and the functions are outside as e.g. `getsockopt`, but elsewhere we had e.g. `Eio_unix.Net.Sockaddr.get`. I changed everything to use the `getsockopt` pattern.
- Before, you had to call the correct pretty-printer depending on the option. Here, I added a `register_printer` function so it works like OCaml's exception printers: `Eio.Net.Sockopt.pp` now works everywhere.
- `pp` now just prints the name. `pp_binding` shows the value too.
- Trying to set or get an option that isn't supported now raises `Io Net Invalid_option`. Before it raised `Invalid_argument` ("the given arguments do not make sense"), but it's not a programming bug to try to set an option that isn't supported.
- I added error context to getters and setters so you can see what it was trying to set on error.
- The documentation and examples implied that you should use `Eio.Net.setsockopt` for portable options and `Eio_unix.Net.Sockopt.set` for Unix-specific ones, but this is misleading. `Eio.Net.setsockopt` works for both; the `Eio_unix` function is just to handle common options for backends.

### DIFF
--- a/lib_eio/unix/net.ml
+++ b/lib_eio/unix/net.ml
@@ -91,3 +91,55 @@ let socketpair_datagram ~sw ?(domain=Unix.PF_UNIX) ?(protocol=0) () =
 
 let fd socket =
   Option.get (Resource.fd_opt socket)
+
+type _ Eio.Net.Sockopt.t +=
+  | Sockopt_bool : Unix.socket_bool_option -> bool Eio.Net.Sockopt.t
+  | Sockopt_int : Unix.socket_int_option -> int Eio.Net.Sockopt.t
+  | Sockopt_optint : Unix.socket_optint_option -> int option Eio.Net.Sockopt.t
+  | Sockopt_float : Unix.socket_float_option -> float Eio.Net.Sockopt.t
+
+let () =
+  let get : type a. a Eio.Net.Sockopt.t -> (string * a Fmt.t) option = function
+    | Sockopt_bool SO_DEBUG -> Some ("Unix.SO_DEBUG", Fmt.bool)
+    | Sockopt_bool SO_BROADCAST -> Some ("Unix.SO_BROADCAST", Fmt.bool)
+    | Sockopt_bool SO_REUSEADDR -> Some ("Unix.SO_REUSEADDR", Fmt.bool)
+    | Sockopt_bool SO_KEEPALIVE -> Some ("Unix.SO_KEEPALIVE", Fmt.bool)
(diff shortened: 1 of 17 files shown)


talex5:
@avsm: The rebased branch with the actual options is temporarily at https://github.com/talex5/eio/tree/sockopt-impl. I didn't force push your branch as there are quite a lot of changes and you might not agree with them. But this does demonstrate that this PR is a reasonable base for whichever version we pick.

avsm:
I have no particularly strong opinions here, esp. about my PR, so feel free to override me :-)

lib_eio/unix/net.ml
```diff
@@ -91,3 +91,55 @@ let socketpair_datagram ~sw ?(domain=Unix.PF_UNIX) ?(protocol=0) () =
...
+    | Sockopt_int SO_RCVBUF -> Some ("Unix.SO_RCVBUF", Fmt.int)
+    | Sockopt_int SO_ERROR -> Some ("Unix.SO_ERROR", Fmt.int)
+    | Sockopt_int SO_TYPE -> Some ("Unix.SO_TYPE", Fmt.int)
+    | Sockopt_int SO_RCVLOWAT -> Some ("Unix.SO_RCVLOWAT", Fmt.int)
+    | Sockopt_int SO_SNDLOWAT -> Some ("Unix.SO_SNDLOWAT", Fmt.int)
+    | Sockopt_optint SO_LINGER -> Some ("Unix.SO_LINGER", Fmt.(option ~none:(any "<none>") int))
+    | Sockopt_float SO_RCVTIMEO -> Some ("Unix.SO_RCVTIMEO", Fmt.float)
+    | Sockopt_float SO_SNDTIMEO -> Some ("Unix.SO_SNDTIMEO", Fmt.float)
+    | _ -> None
+  [@@alert "-deprecated"]
```
avsm, line 123: What's this deprecation thing for?

avsm:
All these changes look excellent to me! I like the toplevel printer approach, much better.

lib_eio/unix/net.ml
talex5: You added it, but it appears to be needed because SO_ERROR is deprecated.

PR #859, opened 2026-06-24
Add portable names for common network options

### LINKED ISSUE #376: Support retrieving socket options

talex5:
This is the next part of @avsm's https://github.com/ocaml-multicore/eio/pull/575 PR.

It adds `Eio.Net.Sockopt.TCP_NODELAY` etc, since these options are also useful outside of Unix processes (e.g. in unikernels or mocks). This is also a nicer API (`Unix`'s API predates GADTs).

Closes #376 and #322.

### DIFF
--- a/tests/network.md
+++ b/tests/network.md
@@ -573,6 +573,53 @@ let try_getsockopt sock opt =
   | exception ex -> traceln "%a -> %a" Eio.Net.Sockopt.pp opt Fmt.exn ex
 ```
 
+Test portable socket options on a TCP socket:
+
+```ocaml
+# run @@ fun ~net sw ->
+  let server = Eio.Net.listen net ~sw ~reuse_addr:true ~backlog:5 addr in
+  let client = Eio.Net.connect ~sw net (Eio.Net.listening_addr server) in
+  Eio.Net.setsockopt client Eio.Net.Sockopt.TCP_NODELAY true;
+  assert (Eio.Net.getsockopt client Eio.Net.Sockopt.TCP_NODELAY);
+  Eio.Net.setsockopt client Eio.Net.Sockopt.SO_KEEPALIVE true;
+  assert (Eio.Net.getsockopt client Eio.Net.Sockopt.SO_KEEPALIVE);
+  Eio.Net.setsockopt client Eio.Net.Sockopt.SO_SNDBUF 32768;
+  let sndbuf = Eio.Net.getsockopt client Eio.Net.Sockopt.SO_SNDBUF in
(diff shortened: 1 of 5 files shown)


tests/network.md
```diff
@@ -573,6 +573,55 @@ let try_getsockopt sock opt =
...
+  traceln "SO_SNDBUF: %s" (if sndbuf > 0 then "positive" else "error");
+  assert (sndbuf > 0);
+  try_getsockopt client Eio.Net.Sockopt.SO_TYPE;
+  Eio.Net.setsockopt client Eio.Net.Sockopt.SO_LINGER (Some 10);
+  try_getsockopt client Eio.Net.Sockopt.SO_LINGER;
+  Eio.Net.setsockopt client Eio.Net.Sockopt.SO_RCVTIMEO 5.0;
+  let timeout = Eio.Net.getsockopt client Eio.Net.Sockopt.SO_RCVTIMEO in
+  traceln "SO_RCVTIMEO: %.1f seconds" timeout;;
++SO_SNDBUF: positive
++SO_TYPE = 1
```
talex5: Is returning the type as a number useful? Does it vary by platform?

avsm: We really ought to just resolve SO_TYPE into a proper type, shouldn't we. Exposing the integer was there for completeness, but not that useful.

avsm (approved):

tests/network.md
avsm: Took a look at doing the reverse conversion. The issue is that we would have to map back to Unix.socket_type, and core Eio doesnt have a `unix` dependency.  We could shift socket types to polymorphic variants so this is no longer an issue, but that's a bigger change (but, I think, nicer from an Eio API perspective?)

avsm: I think my preference is to remove SO_TYPE from the portable list until we can figure this out, rather than expose a raw integer which isn't really that much use from a cross-platform perspective (even though it happens to be the same value as far as I can tell on Linux and macOS)

talex5: Agreed. The useless version has been in OCaml since 2001 and no one has noticed, so I doubt anyone will miss it. https://sherlocode.com/?q=SO_TYPE finds only one use:

```ocaml
let _socktype = Unix.getsockopt_int fd Unix.SO_TYPE in
```

It seems to be just checking whether something is a socket or not. `fstat` would work for that too though, and I guess so would any other socket option.

PR #863, opened 2026-07-03
Fix follow/nofollow confusion in chown

talex5:
- `AT_SYMLINK_FOLLOW` only works with `linkat`
- `AT_SYMLINK_NOFOLLOW` is the flag for `fchownat`

This also updates the tests so they would detect this problem.

Spotted by @avsm in https://github.com/ocaml-multicore/eio/pull/781

/cc @patricoferris

Note: I removed the `AT_EMPTY_PATH` as it doesn't seem to be needed for anything and was only used on Linux anyway. Could be added back later (with a test) if required.


avsm:
I have a smaller fix in #864 since the statx AT_* flag detection should work, but I have no strong preference here.

talex5:
Closing in favour of #864.

PR #864, opened 2026-07-03
use the right flags for chown and add a linux/other test

avsm:
See https://github.com/ocaml-multicore/eio/pull/781#discussion_r3520929700


talex5 (approved):
OK, that is simpler than my version!

Is there any point in keeping the old test? That only tried setting the group to "everyone", so it doesn't test anything that the new version doesn't (except possibly setting the uid to itself).

talex5:
Also, it needs to clean up the symlink; macos test is failing with:
```
+Eio.Io Fs Already_exists _,
+  creating symlink <cwd:symlink.txt> -> link.txt
```

PR #865, opened 2026-07-04
switch fstat to statx on linux and fetch blksize

avsm:
I opted not to add a fallback for pre-5.18 kernels (where the statx is apparently a bit unreliable). The actual feature I need is blksize/blocks filled in (for disk space usage), which is not currently filled in.

I _think_ that fstat using statx with `empty_path` is the correct thing to do. I'm still a little sad that we don't expose a way to selectively fetch just the stat field I need (i.e. why can't I specify _just_ blksize and plumb that through to statx, as in https://github.com/ocaml-multicore/eio/pull/599#issuecomment-1729778486), so I may explore an optional argument for this in the future (once I confirm that statx itself shows a perf improvement if fewer fields are selected; it might ignore them)


avsm:
rebased now that uring 2.15 is out and #867 is merged

talex5 (approved):
> I opted not to add a fallback for pre-5.18 kernels (where the statx is apparently a bit unreliable).

I've pushed a commit requiring Linux 5.18 to use uring. I don't think there's any point trying to use it and then failing to stat things.

PR #866, opened 2026-07-05
Report errors from getaddrinfo

### LINKED ISSUE #776: Eio.Net.getaddrinfo_stream fails to resolve host after 1019 successful calls on linux
I'm writing an application using cohttp-eio that repeatedly connects to a remote host every 3s. The initial 1019 requests succeed, but all the following requests fail to resolve the host. I can simplify the code to the following:

```ocaml
let connect ~sw ~env addr =
  ignore @@ Eio.Net.connect ~sw env#net addr

let print addr =
  Format.printf "@[%a@]@." Eio.Net.Sockaddr.pp addr

let connect_and_print ~sw ~env addr =
  print addr;
  (match addr with | `Tcp (a,_) -> Eio.Net.Ipaddr.fold ~v4:(fun _ -> connect ~sw ~env addr) ~v6:(fun _ -> ()) a | _ -> ())

let getaddrinfo ~sw ~env uri =
  List.iter (connect_and_print ~sw ~env) @@
  Eio.Net.getaddrinfo_stream ~service:"http" env#net
    (Uri.host_with_default uri)

let google_uri = Uri.of_string "http://google.com"

let rec loop ~sw ~env clock count =
  Format.printf "%d:@." count;
  getaddrinfo ~sw ~env google_uri ;
  loop ~sw ~env clock @@ (count + 1)

let () =
  Eio_main.run @@ fun env ->
  Eio.Switch.run @@ fun sw ->
  let clock = env#clock in
  loop ~sw ~env clock 0
```

This initially given the correct output

```
0:
tcp:142.250.186.206:80
tcp:[2a00:1450:401b:80d::200e]:80
1:
tcp:142.250.186.206:80
tcp:[2a00:1450:401b:80d::200e]:80
```

But after 1019 requests `Eio.Net.getaddrinfo_stream` returns no addresses:

```
1018:
tcp:142.250.186.206:80
tcp:[2a00:1450:401b:80d::200e]:80
1019:
1020:
1021:
1022:
1023:
1024:
1025:
```

I get the same behaviour every single time. I tested this on three different machines: Manjaro, Ubuntu and FreeBSD. It only reproduces on linux systems, so it's likely an issue in `eio_linux`.
Also, it reproduces only when `Eio.Net.connect` is called.

Possibly related https://github.com/ocaml-multicore/eio/issues/351

talex5:
Before, `Net.getaddrinfo` returned `[]` on error, but this isn't very useful:

- Callers must remember to check for this and generate an error.
- The actual error is lost, so the reported error isn't very accurate.

Fixes #776 ("Eio.Net.getaddrinfo_stream fails to resolve host after 1019 successful calls on linux"), where the process has too many files open and starts returning an empty list of addresses rather than giving a useful error.

Notes:

- The error `Connection_failure No_matching_addresses` has been replaced by `Address_lookup_failed _` (it wasn't the connect that was failing anyway).

- We now pass `AI_ADDRCONFIG` on eio_linux and eio_posix, which means that we only return IPv4 or IPv6 addresses if the local system is configured to use them. This is the default behaviour for `getaddrinfo` on Linux.

This is based on Christiano's #352. However, there the C stubs were based on the ones in OCaml and the OCaml developers did not want to relicense them so I've written my own C stub instead.

### DIFF
--- a/lib_eio/unix/eio_unix_stubs.c
+++ b/lib_eio/unix/eio_unix_stubs.c
@@ -5,12 +5,18 @@
 #include <errno.h>
 #include <dirent.h>
 #include <sys/stat.h>
+#include <sys/types.h>
+#ifndef _WIN32
+# include <sys/socket.h>
+# include <netdb.h>
+#endif
 
 #include <caml/mlvalues.h>
 #include <caml/unixsupport.h>
 #include <caml/memory.h>
 #include <caml/bigarray.h>
 #include <caml/alloc.h>
+#include <caml/socketaddr.h>
 
 static void caml_stat_free_preserving_errno(void *ptr) {
   int saved = errno;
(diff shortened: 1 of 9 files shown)


lib_eio/unix/eio_unix_stubs.c
```diff
@@ -113,3 +119,113 @@ CAMLprim value eio_unix_fchownat(value v_fd, value v_path, value v_uid, value v_
...
+      }
+    }
+
+    freeaddrinfo(res);
+
+    v_result = caml_alloc(1, 0);        // Ok res
+    Store_field(v_result, 0, v_list);
+    CAMLreturn(v_result);
+  } else if (r == EAI_SYSTEM) {
+    uerror("getaddrinfo", v_node);
```
avsm, line 242: I think the getaddrinfo errno has been thoroughly clobbered by this point, so you probably want to save it earlier

lib_eio_posix/net.ml
```diff
@@ -87,27 +87,9 @@ let datagram_handler = Eio_unix.Pi.datagram_handler (module Datagram_socket)
...
-  Err.run (Eio_unix.run_in_systhread ~label:"getaddrinfo") @@ fun () ->
-  let rec aux () =
-    try
-      Unix.getaddrinfo node service []
-      |> List.filter_map to_eio_sockaddr_t
-    with Unix.Unix_error (EINTR, _, _) -> aux ()
-  in
-  aux ()
+  Eio_unix.run_in_systhread ~label:"getaddrinfo" @@ fun () ->
+  Eio_unix.Private.getaddrinfo ~service node
```
avsm: This is a bit different to the previous code in that it raises Unix_error instead of going through Err.run; is the exception behaviour still the same?

lib_eio/unix/eio_unix_stubs.c
```diff
@@ -113,3 +119,113 @@ CAMLprim value eio_unix_fchownat(value v_fd, value v_path, value v_uid, value v_
...
+  int r;
+  int have_node = caml_string_length(v_node) > 0;
+  int have_service = caml_string_length(v_service) > 0;
+
+  {
+    struct addrinfo hints = {
+      .ai_family = AF_UNSPEC,
+      .ai_socktype = 0,
+      .ai_protocol = 0,
+      .ai_flags = AI_ADDRCONFIG,
```
avsm, line 171: I think setting AI_ADDRCONFIG is the right thing, but it's a change in behaviour so ought to be documented. I'm wondering if this'll be noticeable in Docker environments, where ipv6 is very weird.

talex5: The Linux getaddrinfo man-page says:
>  Specifying hints as NULL is equivalent to setting ... `ai_flags` to `(AI_V4MAPPED | AI_ADDRCONFIG)`.

So, I think `AI_ADDRCONFIG` should be safe on Linux and Docker (and `AI_V4MAPPED` doesn't do anything unless "ai_family equals `AF_INET6`", which it doesn't).

But yes, it is a change relative to OCaml's `Unix.getaddrinfo` and should be noted in the changelog.

lib_eio/unix/eio_unix_stubs.c
talex5: Good point. Fixed in 206c1b30.

lib_eio_posix/net.ml
talex5: You're right - this needs to be wrapped. Fixed in 206c1b30.

Annoyingly, glibc's `getaddrinfo` seems to have a bug that prevents it from reporting system errors correctly anyway.

talex5 (commented):
Updated - thanks! I also changed the C stubs a bit to create the Eio addresses manually, rather than hoping they had the same memory layout as OCaml's ones.

avsm (approved):
lgtm!

PR #867, opened 2026-07-05
uring: update min to 2.15.0+ so we can use new functionality

avsm:


Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM ocaml/opam:debian-13-ocaml-5.4
 RUN sudo ln -sf /usr/bin/opam-2.5 /usr/bin/opam
 # Ensure opam-repository is up-to-date:
-RUN cd opam-repository && git pull -q origin 85e16f902396ddeb2ceae04e43fca88275e7a626 && opam update
+RUN cd opam-repository && git pull -q origin 68acdc87b95f10206b36751f535034556b3cb606 && opam update
```
talex5, line 4: Will need the same change in `bench.Dockerfile` too.

avsm: done

avsm:
also bumped to use ocaml 5.5 in both dockerfiles

PR #868, opened 2026-07-07
eio_linux: use Uring.Res module for safer error handling

talex5:


lib_eio_linux/low_level.ml
```diff
@@ -282,50 +278,51 @@ let splice src ~dst ~len =
...
+  | Ok 0 -> raise End_of_file
+  | Ok n -> n
+  | Error e -> raise @@ Err.wrap e "splice" ""
 
 let connect fd addr =
   Fd.use_exn "connect" fd @@ fun fd ->
   let res = Sched.enter "connect" (enqueue_connect fd addr) in
-  if res < 0 then (
+  match Uring.Res.int_result res with
+  | Ok _ -> ()
```
avsm:
```suggestion
  | Ok 0 -> ()
  | Ok _ -> raise (Err.unclassified (Eio_unix.Unix_error (Unix.EINVAL, "connect", "")))
```

talex5: I've added a `unit_result` function that returns `Error ERANGE` if the result is positive (`ERANGE` is "Result too large", which seems more appropriate than `EINVAL`).

As well as `connect` (which previously ignored positive results), this also affects `noop`, `statx` and `unlink` (which previously would try to interpret the number as a positive error code). All of these functions are only supposed to return 0 on success, so we shouldn't ever see this error.

avsm: Yes this sounds great, thanks!

Issue #835, opened 2026-03-17
Eio.Path.load has race condition and fails on /proc


vog:
The current implementation of `Eio.Path.load` first determines the file size, then reads the file contents into a buffer of that size. This approach has two major downsides:

1. This introduces a race condition if the file is changing between `stat` and `open/read`. It can fail in two directions:
    * If the file became larger in between, `Eio.Path.load` will load too few data without noticing
    * If the file became smaller in between, `Eio.Path.load` will reach EOF earlier than expected
2. This makes reading from `/proc` largely impossible via `Eio.Path.load`, as those "files" are generated at the moment they are read, so the kernel simply reports size 0 for them.

Especially the second case caused quite some debugging time for me, which I'd like to save future Eio users from.

For example:

```ocaml
# Eio_main.run @@ fun env ->
  Eio.Path.load Eio.Path.(env#fs / "/proc/sys/fs/file-nr")
- : string = ""
```

but:

```shell
$ cat /proc/sys/fs/file-nr
16320   0       9223372036854775807
```

I propose to:
* change `Eio.Path.load` so that it uses the file size only as a hint for the initial buffer size, and to
* make `Eio.Path.load path` otherwise largely equivalent to `Eio.Path.with_open_in path Eio.Flow.read_all`

What do you think?

vog:
Some background information:
* https://linuxvox.com/blog/any-idea-why-my-c-code-can-t-read-from-proc/

patricoferris:
Thanks @vog!

Both of these problems are touched upon in https://github.com/ocaml-multicore/eio/issues/798 I think, though the raciness is only eluded to by @copy in:

>...has the advantage of being more robust against other stat problems, at the cost of an extra syscall.

Perhaps we could update the title of this bug to be explicitly about the raciness and keep #798 about `/proc`?

Issue #798, opened 2025-02-27
Eio.Path.load doesn't work for files in /proc (which don't report their size in stat)


copy:
To reproduce:

```ocaml
utop [1]: #require "eio_posix";;
utop [2]: open Eio.Path;;
utop [3]: Eio_posix.run (fun env -> Eio.Path.load (env#fs / "/proc/loadavg"));;
val _5 : string = ""
```

patricoferris:
Thanks for the report @copy ! It appear the issue here is that `load` does a `Eio.File.size` and uses that to create an internal buffer to read into. The files in `proc` aren't really "real" so have a 0 file size which is leading to this problem.

I think the right approach here would be to document this limitation of `Eio.Path.load` and in this case to explicitly collect the data? Maybe something like:

```ocaml
  Eio_posix.run @@ fun env ->
  let loadavg =
    let buf = Buffer.create 128 in
    Eio.Path.with_open_in (env#fs / "/proc/loadavg") @@ fun src ->
    Eio.Flow.copy src (Eio.Flow.buffer_sink buf);
    Buffer.contents buf
  in
  Eio.traceln "loadavg: %s" loadavg
```

Alternatively we could update the implementation of `load` to not check the size of the file before trying to read it.

patricoferris:
((Only after writing all of that did I see your comment in parentheses ^^"))

talex5:
We already have:
```ocaml
  let read_all flow =
    Buf_read.(parse_exn take_all) flow ~max_size:max_int
```

Perhaps `if size = 0 then read_all flow` would do here?

copy:
> Perhaps if size = 0 then read_all flow would do here?

Alternatively, we could do a 1-byte read and if it doesn't return eof, continue reading until eof. That has the advantage of being more robust against other `stat` problems, at the cost of an extra syscall.

talex5:
Fixed in #869.

PR #869, opened 2026-07-07
Allow Eio.Path.load to handle files with zero size

patricoferris:
This is the suggestion in #798 (and related to #835).

Using the buffered reader is probably ok given the initial size defaults to 4096 given `~max_size:max_int` which should fit most files in `/proc`.

Is it worth adding a note in `Eio.Path.load`'s docs that it is racy w.r.t to `stat` and `read`?

### DIFF
--- a/lib_eio/path.ml
+++ b/lib_eio/path.ml
@@ -156,23 +156,21 @@ let with_lines path fn =
   let buf = Buf_read.of_flow flow ~max_size:max_int in
   fn (Buf_read.lines buf)
 
-let load (t, path) =
-  with_open_in (t, path) @@ fun flow ->
+let load t =
+  with_open_in t @@ fun flow ->
   try
     let size = File.size flow in
     if Optint.Int63.(compare size (of_int Sys.max_string_length)) = 1 then
       raise @@ Fs.err File_too_large;
-    let buf = Cstruct.create (Optint.Int63.to_int size) in
-    let rec loop buf got =
-      match Flow.single_read flow buf with
-      | n -> loop (Cstruct.shift buf n) (n + got)
-      | exception End_of_file -> got
+    let buf =
+      Buf_read.of_flow flow
+        ~initial_size:(Optint.Int63.to_int size + 1)
+        ~max_size:(Sys.max_string_length + 1)
     in
-    let got = loop buf 0 in
-    Cstruct.to_string ~len:got buf
+    Buf_read.take_all buf
   with Exn.Io _ as ex ->
     let bt = Printexc.get_raw_backtrace () in
-    Exn.reraise_with_context ex bt "loading %a" pp (t, path)
+    Exn.reraise_with_context ex bt "loading %a" pp t
 
 let save ?append ~create path data =
   with_open_out ?append ~create path @@ fun flow ->


avsm (approved):
I think this is a better fix than reading one byte as it doesnt disadvantage a normal file. A function like load will always be racy in the presence of concurrent writes so no need to document this specially I think.

talex5:
The original version of `load` was:
```ocaml
let load path =
  with_open_in path @@ fun flow ->
  let buf = Buf_read.of_flow flow ~max_size:max_int in
  Buf_read.take_all buf
```
https://github.com/ocaml-multicore/eio/pull/339 changed it because it keeps doubling the size of the internal buffer, which was a bit inefficient.

But I'm not sure why we ended up with the current solution. Something like this would work better I think:
```ocaml
let load t =
  with_open_in t @@ fun flow ->
  try
    let size = File.size flow in
    if Optint.Int63.(compare size (of_int Sys.max_string_length)) = 1 then
      raise @@ Fs.err File_too_large;
    let buf =
      Buf_read.of_flow flow
        ~initial_size:(Optint.Int63.to_int size + 1)
        ~max_size:(Sys.max_string_length + 1)
    in
    Buf_read.take_all buf
  with Exn.Io _ as ex ->
    Exn.reraise_with_context ex bt "loading %a" pp t
```

patricoferris:
That does seem more sensible. Not perfect, but a small benchmark (reading a bunch of small/large files from a directory) using both approaches (your suggestion and the original `Eio.Path.load`) suggests they are pretty similar:

```
% dune exec -- ./main.exe
                          name,   major-allocated,   minor-allocated,    monotonic-clock
  bench/original-small      10,          2.477670,       8055.862394,      199379.483836
  bench/original-small      50,         20.489823,      44927.049155,     1028498.381240
  bench/original-small     100,         44.218219,      91234.650607,     2841583.009717
  bench/original-small    1000,        473.505495,     987563.989011,    24233410.923077
  bench/original-small    5000,       2391.357143,    4977909.642857,   102770825.857143
  bench/proposed-small      10,          3.069692,       8007.191261,      277614.621664
  bench/proposed-small      50,         20.614740,      43508.057894,     1176147.350266
  bench/proposed-small     100,         42.675202,      92549.451873,     2879218.874822
  bench/proposed-small    1000,        531.219780,     987594.153846,    24483348.340659
  bench/proposed-small    5000,       2833.071429,    5033206.357143,   101209270.071429
    bench/original-big      10,      21293.108148,      10381.547170,      315164.342339
    bench/original-big      50,     100957.090147,      49962.001451,     1486204.798097
    bench/original-big     100,     199986.397759,      99543.789916,     3506254.924370
    bench/original-big    1000,    2002639.186813,     998555.758242,    25592058.901099
    bench/original-big    5000,   10013301.857143,    5027031.785714,   134342568.714286
    bench/proposed-big      10,      21257.235061,      10376.696960,      301365.891765
    bench/proposed-big      50,     101081.912987,      50418.304329,     1159692.769120
    bench/proposed-big     100,     200461.811498,      99619.739721,     2240835.164460
    bench/proposed-big    1000,    2002467.000000,    1004815.076923,    27429702.714286
    bench/proposed-big    5000,   10012326.000000,    5053058.714286,   120991733.214286
  ``` 
  
  I have updated the file with the proposed

talex5:
Thanks for the benchmarks - that looks reasonable to me too.

PR #870, opened 2026-07-07
fork_action: retry on eintr when writing to pipe

avsm:
No harm doing this I think, in case the child is interrupted


talex5 (approved):
Seems reasonable (though I guess the error pipe isn't ever going to be full, so the write shouldn't be blocking).

avsm:
Yes I think the only case here that matters is a rare signal interrupt causing the EINTR

PR #871, opened 2026-07-07
eio_linux: these read/write functions didnt hold the fd open

avsm:
All the other fd functions (like the readv/writev) hold the fd open before entering the scheduler, but these run within it. Fix the inconsistency by lifting them Fd use out of the sched closure.


talex5 (approved):
Looks good. The old code was clearly wrong - we need to keep the FD alive at least until the `Uring.submit`.

Issue #845, opened 2026-05-29
eio_linux should disable fixed buffers by default


talex5:
On old versions of ZFS, using fixed buffers triggers a kernel bug and the machine needs to be rebooted. Since fixed buffers rarely provide any performance benefit, we should disable them by default (or possibly detect buggy kernel versions).

Originally reported in #797.

avsm:
I do wonder if we should just restrict io_uring by default to kernel 7+ or higher (for the default selection in Eio_main; just let the user manually pick it if they want). There have been so many bugs fixed...

avsm:
I was talking to @patricoferris this afternoon, and he's also quite up for restricting the uring backend to kernel 6.17 or 7.x+. 
This would avoid a _lot_ of op probes in Eio for the newer features I've been added from uring, and in general any performance-sensitive user is going to be on a recent kernel anyway. "Normal" eio users would do fine on the posix backend.

talex5:
There can be bugs in eio_posix too though (e.g. #856), and if the user can still select eio_linux manually then you still need the probes anyway.

PR #872, opened 2026-07-07
chmod: prevent sandbox escape using symlinks

avsm:
There appears to be a mess involving AT_SYMLINK_NOFOLLOW and chmod on Linux, see https://www.phoronix.com/news/fchmodat2-For-Linux-6.6 for background.

This testcase does fail but I haven't figured out a fix! This does increase the case for #845 and a modern glibc/kernel minimum I think.

### DIFF
--- a/lib_eio_linux/low_level.ml
+++ b/lib_eio_linux/low_level.ml
@@ -741,13 +741,24 @@ let read_link fd path =
     Eio_unix.run_in_systhread ~label:"read_link" (fun () -> Eio_unix.Private.read_link (Some parent) leaf)
   with Unix.Unix_error (code, name, arg) -> raise @@ Err.wrap_fs code name arg
 
-let chmod ~follow ~mode dir path =
-  let module X = Uring.Statx in
-  let flags = if follow then X.Flags.empty else X.Flags.symlink_nofollow in
-  let flags = (flags :> int) in
+let chmod ~follow ~mode dirfd path =
   try
-    with_parent_dir_fd dir path @@ fun parent leaf ->
-    Eio_unix.run_in_systhread ~label:"chmod" (fun () -> Eio_unix.Private.chmod parent leaf ~mode ~flags)
+    Switch.run @@ fun sw ->
+    let fd =
+      openat ~sw ~access:`R ~perm:0 dirfd path
+        ~flags:Uring.Open_flags.(cloexec + path + if follow then empty else nofollow)
+    in
+    Eio_unix.run_in_systhread ~label:"chmod" (fun () ->
+        try
+          Eio_unix.Private.chmod ~mode fd ""
+            ~flags:(Uring.Statx.Flags.empty_path :> int)
+        with
+        | Unix.Unix_error ((ENOSYS | EOPNOTSUPP | EINVAL), _, _) ->
+          (* For Linux before 6.6 *)
+          Eio_unix.Fd.use_exn "chmod" fd @@ fun fd ->
+          let fd : int = Obj.magic fd in
+          Unix.chmod (Printf.sprintf "/proc/self/fd/%d" fd) mode
+      )
   with Unix.Unix_error (code, name, arg) -> raise @@ Err.wrap_fs code name arg
 
 let chown ~follow ?(uid=(-1L)) ?(gid=(-1L)) dirfd path =
--- a/tests/fs.md
+++ b/tests/fs.md
@@ -898,6 +898,64 @@ Chmod works.
 - : unit = ()
 ```
 
+`chmod ~follow:true` follows the leaf symlink within the sandbox.
+A leaf symlink pointing out of the sandbox must not be followed:
+
+```ocaml
+# run ~clear:["sandbox"; "outside.txt"] @@ fun env ->
+  let cwd = Eio.Stdenv.cwd env in
+  Switch.run @@ fun sw ->
+  Path.save ~create:(`Exclusive 0o644) (cwd / "outside.txt") "outside";
+  try_mkdir (cwd / "sandbox");
+  Path.save ~create:(`Exclusive 0o644) (cwd / "sandbox/inside.txt") "inside";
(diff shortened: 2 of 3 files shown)


talex5:
> This does increase the case for https://github.com/ocaml-multicore/eio/issues/845 and a modern glibc/kernel minimum I think.

Blocking eio_linux doesn't help here - eio_posix will have the same problem. But this shouldn't be failing on a modern system anyway.

For a confined path, `chmod ~follow:true` should *always* use no-follow in the system call, and should manually resolve the symlink and retry if that fails. In general, `Resolve.with_parent` only makes sense for operations on the parent directory, such as unlinking or renaming, which don't follow symlinks anyway.

Probably it needs something like this (for eio_posix):

```ocaml
let chmod ~follow ~mode dirfd path =
  Err.run (in_worker_thread "chmod") @@ fun () ->
  let use_confined dirfd =
    Resolve.with_parent_loop ?dirfd path @@ fun dirfd path ->
    let dirfd = Option.value dirfd ~default:at_fdcwd in
    match Eio_unix.Private.chmod_unix dirfd path ~mode ~flags:Config.at_symlink_nofollow with
    | () -> Ok ()
    | exception (Unix.Unix_error (EOPNOTSUPP, _, _) as e) when follow -> (* TODO: check error on other platforms *)
      Error (`Symlink (Some e))
  in
  match dirfd with
  | Fs ->
    let flags = if follow then 0 else Config.at_symlink_nofollow in
    Eio_unix.Private.chmod_unix at_fdcwd path ~mode ~flags
  | Cwd -> use_confined None
  | Fd dirfd -> Fd.use_exn "chmod" dirfd @@ fun dirfd -> use_confined (Some dirfd)
```

lib_eio_linux/low_level.ml
```diff
@@ -741,13 +741,15 @@ let read_link fd path =
...
-  let flags = (flags :> int) in
+let chmod ~follow ~mode dirfd path =
   try
-    with_parent_dir_fd dir path @@ fun parent leaf ->
-    Eio_unix.run_in_systhread ~label:"chmod" (fun () -> Eio_unix.Private.chmod parent leaf ~mode ~flags)
+    Switch.run @@ fun sw ->
+    let flags = Uring.Open_flags.(cloexec + path + if follow then empty else nofollow) in
+    let fd = openat ~sw ~flags ~access:`R ~perm:0 dirfd path in
+    Eio_unix.run_in_systhread ~label:"chmod" (fun () ->
+        let flags = (Uring.Statx.Flags.empty_path :> int) in
```
talex5: The Linux man-page doesn't mention `AT_EMPTY_PATH`, but it seems to work.

talex5 (commented):
I've rebased this on top of #877 (fixing the problem for `chown`), and added a fix for `chmod` here. The fix is essentially the same, except that the race in the eio_posix version goes away on Linux because it doesn't support chmod on symlinks!

lib_eio_linux/low_level.ml
avsm: I think the AT_EMPTY_PATH requires kernel 6.6 (with fchmodat2) and glibc 2.39+ (afaict), which is reasonably modern. It'll fail with an ENOTSUPP (or maybe EINVAL) before that.

This seems reasonable to me... let's just document the lower bounds here.

tests/fs.md
```diff
@@ -898,6 +898,33 @@ Chmod works.
 - : unit = ()
 ```
 
+`chmod ~follow:true` and `chown ~follow:true` follow the leaf symlink within the sandbox.
```
avsm: We should  add a `chmod ~follow:false` test as well I think. This is sufficiently complex that checking the portable behaviour of both paths on the CI would be helpful.

talex5: I've added one. Note that we can't test chmodding a symlink easily, as Linux doesn't support that.

lib_eio_linux/low_level.ml
talex5: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=5daeb41a6fc9d0d81cb2291884b7410e062d8fa1 notes that on older systems you can do it via `/proc/self/fd`, but not sure it's worth it.

talex5: Actually, it's easy enough. I've pushed an attempt. Do you have an old version of Linux to test it on?

talex5: Tested in a Linux nixos 6.1.159 VM and it works. strace shows:
```
1114  fchmodat2(5, "", 0600, AT_EMPTY_PATH) = -1 ENOSYS (Function not implemented)
1114  chmod("/proc/self/fd/5", 0600)    = 0
```

tests/fs.md
avsm: FreeBSD does support it, so ocaml-ci should exercise it on there

lib_eio_linux/low_level.ml
avsm: I resisted mentioned in the /proc/self trick as it adds another Obj.magic ;-) This is for a followup PR but I think we should purge those and have a C binding for this, as the risk of miscompilation with OxCaml is higher.

avsm: Does the ENOSYS in your trace above get converted to a EOPNOTSUPP/EINVAL somewhere, so that your pattern match catches it?

avsm:
I cant actually approve "my own PR" on the GitHub UI, but this looks right to me.

lib_eio_linux/low_level.ml
talex5: That's a very good question! Looks like glibc catches `ENOSYS` and calls `fchmodat_fallback`, which returns `EINVAL` due to the `empty_path`. I've added `ENOSYS` to the list in case other libraries/versions work differently.

tests/fs.md
talex5: I've added an extra test for that now.

PR #873, opened 2026-07-07
getsockopt_string: fix truncation check

avsm:
Previously it was checking against the static size of the buffer, so switch the returned size.

### DIFF
--- a/lib_eio_linux/eio_stubs.c
+++ b/lib_eio_linux/eio_stubs.c
@@ -289,8 +289,9 @@ CAMLprim value caml_eio_sockopt_string_get(value v_fd, value v_level, value v_op
   int ret = getsockopt(Int_val(v_fd), Int_val(v_level), Int_val(v_option),
                        buffer, &len);
   if (ret == -1) uerror("getsockopt", Nothing);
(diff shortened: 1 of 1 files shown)


avsm:
I think I confused myself here; optlen is actually modified isn't it, so the original code is right. A quick pair of eyes from someone else to confirm this isn't necessary would be great.

avsm:
I'm pretty sure the existing code is right on a third re-read ;-)

talex5:
> Previously it was checking against the static size of the buffer, so switch the returned size.

The previous check was: if the returned string fills the static buffer (i.e. there wasn't room for the terminator) then the buffer was too small. That looks correct to me.

The obvious test here would be to check if `len == sizeof(buffer)`, but I have an idea that doesn't work (Linux appears to set it to 16, the max length).

The new version proposed in this PR checks if the returned string has the size Linux claims. If Linux was working correctly, this would always report an overflow, which is wrong.

avsm:
Thanks; I always forget that the len passed into getsockopt was mutated after the call, which is what confused me! And my new version was wrong too. Keep me away from buffer lengths.

PR #850, opened 2026-06-05
add ftruncate/fdatasync and use uring for fsync/shutdown/socket/bind/listen

avsm:
Needs trunk uring

PR #875, opened 2026-07-07
Pre-connect socket options (follow up to #713)

avsm:
Only 92d2924b1824b723610fa69ce54d6b1d03a3b31b is relevant here; it is layered over #850 

This is based on work by @art-w in #713 and revised based on review comments in that PR.  It exposes a single `options` GADT that can specify either bindings or setting options. Could be simplified more (to remove the convenience functions like `reuse_addr` if deemed too verbose) but I wanted to check on the basic interface first.

### DIFF
--- a/lib_eio_linux/net.ml
+++ b/lib_eio_linux/net.ml
@@ -93,10 +93,17 @@ let socket_domain_of = function
       ~v4:(fun _ -> Unix.PF_INET)
       ~v6:(fun _ -> Unix.PF_INET6)
 
-let connect ~sw connect_addr =
+let connect ~bind_to ~options ~sw connect_addr =
   let addr = Eio_unix.Net.sockaddr_to_unix connect_addr in
   let sock = Low_level.socket ~sw (socket_domain_of connect_addr) Unix.SOCK_STREAM 0 in
-  Low_level.connect sock addr;
+  let rec set_options : Eio.Net.Sockopt.settings -> unit = function
+    | [] -> ()
+    | (o, v) :: tl -> Low_level.setsockopt sock o v; set_options tl
+  in
+  Err.run (fun () ->
+    set_options options;
+    Option.iter (fun a -> Low_level.bind sock (Eio_unix.Net.sockaddr_to_unix a)) bind_to;
+    Low_level.connect sock addr) ();
   (Flow.of_fd sock :> _ Eio_unix.Net.stream_socket)
 
 module Impl = struct
@@ -132,7 +139,8 @@ module Impl = struct
     Low_level.listen sock backlog;
     (listening_socket sock :> _ Eio.Net.listening_socket_ty r)
 
-  let connect () ~sw addr = (connect ~sw addr :> [`Generic | `Unix] Eio.Net.stream_socket_ty r)
+  let connect () ~bind_to ~options ~sw addr =
+    (connect ~bind_to ~options ~sw addr :> [`Generic | `Unix] Eio.Net.stream_socket_ty r)
 
   let datagram_socket () ~reuse_addr ~reuse_port ~sw saddr =
     if reuse_addr then (
--- a/lib_eio_linux/net.ml
+++ b/lib_eio_linux/net.ml
@@ -93,9 +93,15 @@ let socket_domain_of = function
       ~v4:(fun _ -> Unix.PF_INET)
       ~v6:(fun _ -> Unix.PF_INET6)
 
-let connect ~sw connect_addr =
+let connect ~bind_to ~options ~sw connect_addr =
   let addr = Eio_unix.Net.sockaddr_to_unix connect_addr in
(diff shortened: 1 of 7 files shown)


talex5:
Hmm. Looking at the example, it is getting a bit messy now, turning into a little DSL:

```ocaml
Eio.Net.connect ~sw net (Eio.Net.listening_addr server)
  ~options:Eio.Net.Connect.[
      reuse_addr;
      set Eio.Net.Sockopt.SO_KEEPALIVE true;
      bind_to (`Tcp (Eio.Net.Ipaddr.V4.loopback, 0));
    ]
```

`Net.connect` and `Net.listen` are quite convenient, but maybe it's time to provide a low-level API as well, that matches the BSD sockets API more closely? e.g.

```ocaml
  let x = Eio.Net.stream_socket ~sw net in
  Eio.Net.setsockopt x Eio.Net.Sockopt.SO_REUSEADDR true;
  Eio.Net.setsockopt x Eio.Net.Sockopt.SO_KEEPALIVE true;
  Eio.Net.bind_socket x (`Tcp (Eio.Net.Ipaddr.V4.loopback, 0));
  Eio.Net.connect_socket x (Eio.Net.listening_addr server);
```

avsm:
Indeed, I've been mulling over these little DSLs that we have in Eio (including Fork_actions). A design principle emerging here is that Eio is where the portable expression (i.e. the DSL) of application intent goes, which is then interpreted by the backends into specific calls.

This is only useful if an Eio backend exists that can usefully do something different from the simple raw calls, by virtue of being able to iterate over the user DSL and not see every low-level call one-by-one.

For filesystem work, the advantages of batching seem obvious for performance.  For networking, it's a little less obvious.

- Performance: I really like having a 'default fast and sensible' parallel-friendly set of network calls. I'd like to be able to call accept with a domain manager, and have it do the right thing in the backend (accept in parallel, dispatch to domains). This seems particularly applicable to Eio_linux, where management of fixed buffers and so on needs to be per-domain due to the ring being domain-affine.
- Portability: @samoht has been working on a mirage-eio, so avoiding baking the sockets API into the Eio frontend is desirable here, I'd say
- Tracing: It's quite difficult to know when to set certain socket options. I could imagine us building in a backpressure API in Eio_linux that keeps track of socket buffers and can balloon them. It feels like this would be easier if we keep the frontend a bit more abstract.
- Safety: Does the frontend DSL forbid any code that might be useful when written manually? Fork_action does forbid some patterns since we have to patch Eio itself to add one. But a sockopt DSL does full passthrough of the underlying options, so it seems ok...

avsm:
I'm just peering at the Windows support for these sock options; they seem mostly-but-not-quite similar so far. Will update when I get through the interfaces.

avsm:
Simplified to make bind_to a parameter and the options a heterogenous flat list of sockopts to set, after discussion with @talex5

lib_eio_linux/net.ml
```diff
@@ -93,10 +93,17 @@ let socket_domain_of = function
...
-let connect ~sw connect_addr =
+let connect ~bind_to ~options ~sw connect_addr =
   let addr = Eio_unix.Net.sockaddr_to_unix connect_addr in
   let sock = Low_level.socket ~sw (socket_domain_of connect_addr) Unix.SOCK_STREAM 0 in
-  Low_level.connect sock addr;
+  let rec set_options : Eio.Net.Sockopt.settings -> unit = function
+    | [] -> ()
+    | (o, v) :: tl -> Low_level.setsockopt sock o v; set_options tl
+  in
+  Err.run (fun () ->
```
talex5: Do we need `Err.run` here?

`Low_level.setsockopt`, `Low_level.bind` and `Low_level.connect` all raise Eio errors already.

tests/network.md
```diff
@@ -647,6 +647,64 @@ Exception: Eio.Io Net Invalid_option,
...
+      ~bind_to:(`Tcp (Eio.Net.Ipaddr.V4.loopback, 0))
+      ~options:Eio.Net.Sockopt.[
+          (SO_REUSEADDR, true);
+          (SO_KEEPALIVE, true);
+        ]
+  in ();;
++mock-net: connect to tcp:127.0.0.1:8081
++mock-net: setsockopt SO_REUSEADDR = true
++mock-net: setsockopt SO_KEEPALIVE = true
++mock-net: bind tcp:127.0.0.1:0
```
talex5: This output is a bit confusing. I've changed it to set the options on the socket rather than the network, and display the bind-to address along with the connect address.

talex5 (approved):

lib_eio_linux/net.ml
avsm: I think you're right. It's a little confusing, as some backends raise Unix_errors and others wrap in Eio exns!

avsm: removed and rebased in 399aa7aa656e37b920e9ef2e0cb5b725798190e7

PR #876, opened 2026-07-08
Be more careful about unterminated string socket options

talex5:
Following on from @avsm's #873, this is another attempt to make the code more robust, as well as documenting why it's difficult.


avsm:
This is hilariously complicated.

PR #877, opened 2026-07-08
chown: prevent sandbox escape using symlinks

talex5:
Before, we opened the parent directory and then did `chown`, following symlinks if called with `~follow:true`. This is not safe, since the symlink may point outside of the sandbox.

The new code works as follows:

On eio_linux:
- Instead of opening the parent directory, just open the target itself with `O_PATH` and use `AT_EMPTY_PATH` to set it.

On eio_posix with follow=false:
- Open the parent directory and do `chown ~follow:false` on the child.

On eio_posix with follow=true:
- Open the parent directory.
- If the child is a symlink, follow it.
- Otherwise, `chown ~follow:false` on it.

For this last case (eio_posix+follow), there is a small race:
- If a symlink changes into a non-symlink while we're chowning it, we'll raise an error from `readlink`.
- If a non-symlink changes into a symlink, we'll change the ownership of the symlink even though we're supposed to be following links.

We don't escape the sandbox in either case.

(this addresses just the `chown` part of @avsm's #872)

/cc @patricoferris 

(note: chown support hasn't yet appeared in a release)

### DIFF
--- a/lib_eio_linux/low_level.ml
+++ b/lib_eio_linux/low_level.ml
@@ -750,13 +750,15 @@ let chmod ~follow ~mode dir path =
     Eio_unix.run_in_systhread ~label:"chmod" (fun () -> Eio_unix.Private.chmod parent leaf ~mode ~flags)
   with Unix.Unix_error (code, name, arg) -> raise @@ Err.wrap_fs code name arg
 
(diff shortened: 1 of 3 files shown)


talex5:
Merging this as the previous behaviour is clearly wrong, but more testing in this area wouldn't be a bad idea.

PR #879, opened 2026-07-10
Clarify Eio as a possible concurrency library

patricoferris:
Based on a thread on discuss -- seems fair enough to clarify that if people wanna keep using Lwt they certainly can! Or they can jump in with Eio (or some mix of the two!).

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@ For example, you can use Eio to read and write files, make network connections,
 or perform CPU-intensive calculations, running multiple operations at the same time.
 It aims to be easy to use, secure, well documented, and fast.
 A generic cross-platform API is implemented by optimised backends for different platforms.
-Eio replaces existing concurrency libraries such as Lwt
-(Eio and Lwt libraries can also be used together).
+Eio implements similar functionality to Lwt or Async, but using effects rather than monadic concurrency
+(though it is possible to [use Eio and Lwt libraries together][Lwt_eio]).
 
 ## Contents
 


avsm:
I think the existing "Eio replaces" is more accurate, since it doesn't interoperate with the other libraries! Nothing in this statement implies that users should switch, only that if they do switch, they should replace their existing library.

I was thinking we should just add '[...] Lwt or Async', as those are the two main existing monadic concurrency libraries involved.

avsm:
@redianthus, I notice you gave a thumbs down to my comment, but no constructive feedback. Would you like to explain your reasoning further?

patricoferris:
I agree that, when read in full, the "replaces" is about the fact that it "can also be used with Lwt", though I still think a casual passerby might misinterpret it. I've kept the "can" in to be safe.  Adding in `async` seems sensible since there's a good amount of code in the wild using that particular IO monad!

redianthus:
@avsm because as the PR correctly suggests, it is wrong to say that eio replaces existing concurrency libraries.

For instance, it failed to replace Lwt, see: https://discuss.ocaml.org/t/why-ocsigen-is-staying-on-lwt-for-now

avsm:
I'm trying to home in on advice that'll be most useful to our users. I don't think anyone's going to more enlightened when a 'replaces' turns into a 'can replace'.

I've suggested to @patricoferris that instead of the existing sentence, we indicate that Eio implements similar functionality to Async or Lwt, but using effects, and to add a warning that there are known difficulties with getting multiple effect-using libraries to interop (due to the problem of stacking handlers in the right order). Users can bear that in mind while selecting their library base.

redianthus:
Yes, I agree with you, I think it does not replace Async or Lwt and we should be more explicit by writing something along the lines of "Eio was a tentative to replace Async or Lwt, unfortunately we discovered it might not fit many use cases so have this in mind before choosing this library and you might prefer staying with Lwt" would be clearer for users. WDYT?

patricoferris:
I have updated with that suggestion. I think we could follow up with something about using multiple effect libraries together.

avsm:
Thanks Patrick, and @redianthus for the helpful suggestions!

PR #881, opened 2026-07-11
win32: declare unixsupport functions as CAMLextern

avsm:
This is necessary under Windows to prevent flexdll errors when dynamically loading "caml_unix_error RELOC_REL32, target is too far" and with the fix allows mdx tests to get further

Solution pointed out by @dra27

### DIFF
--- a/lib_eio/unix/pty.c
+++ b/lib_eio/unix/pty.c
@@ -19,6 +19,11 @@
 #include <caml/fail.h>
 #include <caml/unixsupport.h>
 
+#ifdef _WIN32
+CAMLnoret CAMLextern
+void caml_unix_error (int errcode, const char * cmdname, value arg);
+#endif
+
 /* Returns [pty_fd], opened close-on-exec. */
 CAMLprim value eio_unix_open_pty(value v_unit) {
 #ifdef _WIN32
--- a/lib_eio/unix/eio_unix_stubs.c
+++ b/lib_eio/unix/eio_unix_stubs.c
@@ -12,11 +12,21 @@
 #endif
 
 #include <caml/mlvalues.h>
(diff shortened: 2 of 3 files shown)


avsm:
This also fixes the Windows build following #866

lib_eio/unix/pty.c
```diff
@@ -19,6 +19,11 @@
 #include <caml/fail.h>
 #include <caml/unixsupport.h>
 
+#ifdef _WIN32
+CAMLnoret CAMLextern
+void caml_unix_error (int errcode, const char * cmdname, value arg);
```
talex5, line 24: Doesn't this come from ` <caml/unixsupport.h>`, which is still imported above?

talex5 (commented):
Is this a byte-code thing then? The Windows build is passing in CI for the `main` branch.

lib_eio/unix/pty.c
avsm: Not the CAMLExtern, it's declared in that header file as 

```
CAMLnoret extern
void caml_unix_error (int errcode, const char * cmdname, value arg);
```

...which seems to be a possible upstream bug, but I'm just following @dra27's advice here to use redeclare it CAMLextern for now.

avsm:
@talex5 wrote:
>  The Windows build is passing in CI for the main branch.

Aha, the unixsupport.h removal for Windows removed the getaddrinfo constant defines, which broke the build only in this PR. I've got a followon patch to use GetAddrInfoW on Windows, so there's no regression here as getaddrinfo as eio_unix_getaddrinfo returns ENOTSUPP right now.

talex5 (approved):
OK, looks good then!

PR #882, opened 2026-07-12
add braces around switch case block

olafhering:


talex5:
Thanks - good to fix the warnings on macos!

avsm:
I was wondering what that warning was about, thanks!

PR #884, opened 2026-07-13
eio_linux: add Low_level.{socket,bind,listen} with uring support

talex5:
This is the networking part of @avsm's #850, split out to help me review it.

The first commit adds `Low_level.{socket,bind,listen}` helpers to clean up the code and the second gets them to use uring, if available. `shutdown` now also uses uring, and errors are wrapped in more cases.

I made a couple of minor changes:

- Instead of recording a bool for each feature, I added a commit to store the `probe` itself. `opcode_supported` is `noalloc`, so I wouldn't expect this to be a performance problem. Avoids having to modify `sched.ml` every time we need to query a new feature.
- I removed `test_tcp_loopback`. It didn't seem to be testing anything Linux-specific, and we already have loopback tests in the main `tests` directory.

### DIFF
--- a/lib_eio_linux/eio_linux.ml
+++ b/lib_eio_linux/eio_linux.ml
@@ -116,8 +116,7 @@ let socket_domain_of = function
 
 let connect ~sw connect_addr =
   let addr = Eio_unix.Net.sockaddr_to_unix connect_addr in
-  let sock_unix = Unix.socket ~cloexec:true (socket_domain_of connect_addr) Unix.SOCK_STREAM 0 in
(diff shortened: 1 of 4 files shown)


avsm (approved):
Changes all look good to me, thanks!

I used a version of the tcp listen to stress test the change, but didn't include the stress test binary. I can add those in as non-test-case-but-useful-binaries independently of this PR.

The op change looks good; the only other mechanism I thought of was to define a FCM that had each of the function calls (either the uring one or the Unix fallback or exception), but couldn't find an ergonomic way to thread that through.

PR #885, opened 2026-07-14
Remove all use of `[@@@alert "-unstable"]`

talex5:
We must have needed these at some point, but I can't work out when. The `unstable` tag was added to `stdlib/effect.mli` when it was first added in OCaml 5.0.0, but OCaml 5.0.0 already had that alert disabled by default.

I also tried building Eio 0.12 with dune 3.10 and OCaml 5.0.0, and that didn't show the alerts either, so I suspect that disabling them isn't needed now.


talex5:
Ah - it was to work around an old Merlin bug: https://github.com/ocaml/merlin/issues/1558 (fixed in 2024)

PR #887, opened 2026-07-15
Add Eio_unix.Err module

talex5:
This is convenient whenever wrapping a function that raises `Unix_error`, so it makes sense for it to be public. Also, it avoids duplicating the logic across eio_linux, eio_posix and eio_windows, with possible inconsistencies.

I renamed `Err.wrap` to `Err.v`, which seems more sensible. I didn't include `Err.run` in the public API as I don't really like it (it's too easy to partially apply it). This also removes the distinction between `wrap` and `wrap_fs` in eio_linux. eio_posix didn't do this, so it probably wasn't necessary.

This will also be useful if we want to make other functions in `Eio_unix` raise `Eio.Io` exceptions in future.

The new `discover.ml` is nearly empty, but I except it will be a useful place for more shared logic in future.

### DIFF
--- a/lib_eio/unix/config.ml.windows
+++ b/lib_eio/unix/config.ml.windows
@@ -1,0 +1,1 @@
+let enotcapable = None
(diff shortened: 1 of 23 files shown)


avsm:
Combining `wrap` and `wrap_fs` is useful; I was always applying them wrongly

PR #889, opened 2026-07-16
Move sockopt stubs to eio.unix

talex5:
This allows the extended socket options to be shared between eio_linux and eio_posix (which can still be used on Linux), and allows the options to be implemented for other platforms too.

The C stubs are almost the same, except that they should also compile on Windows (based on @avsm's code in capnp-rpc, but untested).


lib_eio/unix/eio_unix_stubs.c
```diff
@@ -23,9 +30,8 @@
 #ifdef _WIN32
 CAMLnoret CAMLextern
 void caml_unix_error (int errcode, const char * cmdname, value arg);
-#define Nothing ((value) 0)
-#else
-#include <caml/unixsupport.h>
+CAMLnoret CAMLextern
+void caml_uerror (const char * cmdname, value arg);
```
talex5, line 34: I don't know if this is needed, but I assume if it was needed for `caml_unix_error` then it's needed here too?

avsm: It's needed if eio_windows ever links against the module in question, which it does here for eio.unix

PR #893, opened 2026-07-18
eio_posix: fall back to select() for fds that poll() fails with POLLNVAL

avsm:
macOS poll() returns POLLNVAL for some device FDs such as /dev/tty or /dev/null. Following the discussions in #757, this is the simplest fix I could find that doesn't affect other platforms like Linux.

We now resume the waiters that fail with POLLNVAL with an [Unpollable] exception, which then uses select() calls via a systhread for these fds only. I'm just not sure if the select is too aggressive at 0.1s timeouts...

Without this fix the new regression test fails with "Fatal error: Assertion failed, sched.ml line 155"

### DIFF
--- a/lib_eio_posix/eio_posix_stubs.c
+++ b/lib_eio_posix/eio_posix_stubs.c
@@ -1,8 +1,14 @@
+#ifdef __APPLE__
+/* Must be defined before including any system headers */
+#define _DARWIN_UNLIMITED_SELECT
+#endif
+
 #include "primitives.h"
 
 #define _FILE_OFFSET_BITS 64
 
 #include <sys/types.h>
+#include <sys/select.h>
 #include <sys/socket.h>
 #ifdef __linux__
 #if __GLIBC__ > 2 || __GLIBC_MINOR__ > 24
@@ -21,6 +27,7 @@
 
 #include <caml/mlvalues.h>
 #include <caml/memory.h>
+#include <caml/fail.h>
 #include <caml/alloc.h>
 #include <caml/unixsupport.h>
 #include <caml/bigarray.h>
@@ -43,6 +50,34 @@ static void caml_stat_free_preserving_errno(void *ptr) {
   int saved = errno;
   caml_stat_free(ptr);
   errno = saved;
+}
+
+/* Wait for a single FD to become readable or writable, with a timeout. */
+CAMLprim value caml_eio_posix_select_one(value v_fd, value v_write, value v_timeout) {
+#ifndef __APPLE__
+  caml_unix_error(EOPNOTSUPP, "This work-around is only for macOS", Nothing);
+#else
+  CAMLparam1(v_timeout);
+  int fd = Int_val(v_fd);
+  double timeout = Double_val(v_timeout);
+  struct timeval tv;
+  int ret;
+
+  fd_set *fds = caml_stat_calloc_noexc((size_t)fd / NFDBITS + 1, sizeof(fd_mask));
+  if (fds == NULL) caml_raise_out_of_memory();
+  FD_SET(fd, fds);
+  tv.tv_sec = (time_t)timeout;
+  tv.tv_usec = (suseconds_t)((timeout - (double)tv.tv_sec) * 1e6);
+  caml_enter_blocking_section();
+  ret = select(fd + 1,
+               Bool_val(v_write) ? NULL : fds,
+               Bool_val(v_write) ? fds : NULL,
+               NULL, &tv);
+  caml_leave_blocking_section();
+  caml_stat_free_preserving_errno(fds);
+  if (ret == -1) uerror("select", Nothing);
+  CAMLreturn(Val_bool(ret > 0));
+#endif
 }
 
 CAMLprim value caml_eio_posix_getrandom(value v_ba, value v_off, value v_len) {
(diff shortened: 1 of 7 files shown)


lib_eio_posix/low_level.ml
```diff
@@ -22,25 +22,48 @@ type dir_fd =
...
+   systhread instead. TODO avsm: is 0.1 too low here? *)
+let unpollable_select_timeout = 0.1
+
+let rec await_unpollable ty fd =
+  let ready =
+    in_worker_thread "select" @@ fun () ->
+    try
+      match (ty : ty) with
+      | Read ->
+        let readable, _, _ = Unix.select [fd] [] [] unpollable_select_timeout in
```
talex5: We probably need our own `select` stub, compiled with `-D_DARWIN_UNLIMITED_SELECT`. Otherwise, this will fail on FDs >= 1024 I think.

Or could perhaps just note in a comment that OCaml's `Unix.select` should be doing this (doesn't look like it is, but I haven't tested it on macOS).

lib_eio_posix/test/test_devices.ml
```diff
@@ -0,0 +1,38 @@
+(* Regression test for https://github.com/ocaml-multicore/eio/issues/757.
+   macOS poll rejects /dev/null/tty with POLLNVAL. *)
+
+open Eio.Std
+
+let with_blocking_fd path flags ~sw fn =
+  let fd = Unix.openfile path flags 0 in
+  fn (Eio_unix.Fd.of_unix ~sw ~blocking:true ~close_unix:true fd)
```
talex5: Minor, but this should either create the switch itself (and thus free the FD promptly when `fn` finishes), or just return the `fd` attached to the switch (and not bother with the `fn` argument).

```suggestion
let with_blocking_fd path flags fn =
  Switch.run @@ fun sw ->
  let fd = Unix.openfile path flags 0 in
  fn (Eio_unix.Fd.of_unix ~sw ~blocking:true ~close_unix:true fd)
```

talex5 (commented):
Seems reasonable as a temporary work-around until Apple fixes this properly.

talex5 (approved):
I tidied up the switches in the test slightly. OK to merge this now - it's clearly an improvement on the current behaviour.

lib_eio_posix/low_level.ml
avsm: Yes, I've pushed cdb79466ab294964358bba4ce5ae877515321e47 (i couldnt find a caml_stat_calloc that has a raising variety, for some reason). Also tweaked timeout to 0.2s

lib_eio_posix/eio_posix_stubs.c
```diff
@@ -45,6 +52,31 @@ static void caml_stat_free_preserving_errno(void *ptr) {
   errno = saved;
 }
 
+/* Wait for a single FD to become readable or writable, with a timeout. */
+CAMLprim value caml_eio_posix_select_one(value v_fd, value v_write, value v_timeout) {
+  CAMLparam0();
```
talex5:
```suggestion
  CAMLparam1(v_timeout);
```
As it's a float (though we're not doing any allocation before reading it, so this is just to be safe against future changes).

lib_eio_posix/low_level.ml
talex5: If the timeout is now 0.2, then the test timeout might need raising too.

lib_eio_posix/eio_posix_stubs.c
```diff
@@ -45,6 +52,31 @@ static void caml_stat_free_preserving_errno(void *ptr) {
...
+  int fd = Int_val(v_fd);
+  double timeout = Double_val(v_timeout);
+  struct timeval tv;
+  fd_set *fds;
+  int ret;
+
+  fds = caml_stat_calloc_noexc((size_t)fd / NFDBITS + 1, sizeof(fd_mask));
```
talex5: Is `fd_mask` etc standard? Though it seems to compile OK on every platform we're testing in CI, so it's probably OK.

avsm: It's been in every select.h I've used, but now that you mention it I checked and it's not defined anywhere (only the macros are). Investigating...

talex5: I've pushed a commit to use a simple stack-allocated fd_set on other platforms (and also use `FD_ZERO` in all cases, although unlikely to make a difference with `calloc`), just in case it fails somewhere.

avsm:
Given how apple specific this workaround is, I'm wondering if we should just fail with an exception on non __APPLE__ platforms entirely if that stub is called.

talex5:
> Given how apple specific this workaround is, I'm wondering if we should just fail with an exception on non APPLE platforms entirely if that stub is called.

Could do. I'm happy with it either way.

avsm:
Just one other thing I'm investigating is whether we can just always stack allocate the fdset even with large fds. We should have enough, but I remember @mdales attempting to have 100k fds open for his geospatial tile stitcher! That should be a followup PR I think; this is fine as is from my perspective.

talex5:
> I'm investigating is whether we can just always stack allocate the fdset even with large fds.

Apparently that won't work: https://news.ycombinator.com/item?id=38237121

talex5:
> I'm wondering if we should just fail with an exception on non APPLE platforms entirely if that stub is called.

OK, I made it macOS-only. Saves worrying about what happens if the FD is out-of-range on a non-macOS system.

PR #894, opened 2026-07-20
add null sink that consumes/discards all data written to it until eof

avsm:
Noticed this might be useful from Forester's eio utils via @jonsterling at https://git.sr.ht/~jonsterling/ocaml-forester/tree/main/item/lib/compiler/Eio_util.ml


talex5:
Probably worth adding read support too; then it can just be called `null`.

talex5 (approved):
I've renamed it to `null` and added read support too. If that looks right to you, it's ready to merge.

avsm:
This looks spot on; thanks!

PR #895, opened 2026-07-20
fork_action: reject ids that don't fit uid_t/gid_t/pid_t

avsm:
The setuid/setgid/setpgid stubs passed OCaml ints to 32-bit C types that would truncate out-of-range values. THe length check now makes sure the OCaml cast isn't truncated.

My macro cast needs to be double checked, but it seems better to do it here as proof against uid_t/gid_t/pid_t becoming 64 bit in the future

### DIFF
--- a/lib_eio/unix/fork_action.c
+++ b/lib_eio/unix/fork_action.c
@@ -250,6 +250,12 @@ CAMLprim value eio_unix_fork_dups(value v_unit) {
   return Val_fork_fn(action_dups);
 }
 
+/* Reject IDs that don't fit the C type rather than silently truncating */
+#define check_id_range(t, v, name) \
+  if ((long)(t)Long_val(v) != Long_val(v)) { \
+    eio_unix_fork_error(errors, name, EINVAL); \
+  }
+
 static void action_setpgid(int errors, value v_config) {
   #ifdef _WIN32
   eio_unix_fork_error(errors, "setpgid", ENOSYS);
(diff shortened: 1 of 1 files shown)


lib_eio/unix/fork_action.c
```diff
@@ -266,6 +273,8 @@ static void action_setpgid(int errors, value v_config) {
   value vpgid = Field(v_config, 2);
 
   int r;
+  check_id_range(pid_t, vpid, "setpgid");
+  check_id_range(pid_t, vpgid, "setpgid");
   r = setpgid(Int_val(vpid), Int_val(vpgid));
```
talex5: It's strange that we do the check with `Long_val` and `pid_t`, but then use `Int_val` at the point of use.

Perhaps it would be better to have the macro return the value itself? e.g.

```c
pid_t pid = checked_range_val(pid_t, vpid, "setpgid");
```

lib_eio/unix/fork_action.c
```diff
@@ -257,6 +257,13 @@ CAMLprim value eio_unix_fork_dups(value v_unit) {
   return Val_fork_fn(action_dups);
 }
 
+/* Reject IDs that don't fit the C type rather than silently truncating */
+#define check_id_range(t, v, name) \
+  if ((long)(t)Long_val(v) != Long_val(v)) { \
+    eio_unix_fork_error(errors, name, EINVAL); \
+    _exit(1); \
```
talex5: I've opened https://github.com/ocaml-multicore/eio/pull/896 so we can avoid the `_exit` here.

talex5 (approved):
I rebased it, removed the exit (which is no longer needed) and changed `Int_val` to `Long_val` to match the tests.

avsm:
All looks good to me; I thought Int_val was equivalent to Long_val here, but matching the tests is certainly sensible.

PR #897, opened 2026-07-21
eio_linux: don't return an empty getdents batch before end-of-file

avsm:
Since the ./.. filtering moved into the C stub in 1f7b135810af663b75, a batch containing _only_ those entries returned an empty list, which also marks the end of the directory structure. While this is harmless for local filesystems, it can be a problem on FUSE where the protocol allows for shorter returns. 

Therefore, we now loop within the stub until we have at least one real dirent or eof. Could also fix it by signaling eof by other means, but that involves an allocation or an exception. Blocking in a loop seems reasonable.


talex5 (approved):
Thanks!

PR #907, opened 2026-07-27
Fix handling of empty paths when using fs unconfined

### LINKED ISSUE #906: Eio.Path.mkdirs fails on a relative path whose first component doesn't exist
It seems that `Eio.Path.mkdirs cannot create a nested relative directory tree while the same call with an absolute path succeeds.

Small repro:

```ocaml
Eio_main.run @@ fun env ->
let fs = Eio.Stdenv.fs env in
Unix.chdir (Filename.temp_dir "repro" "");
Eio.Path.mkdirs ~exists_ok:true ~perm:0o755 Eio.Path.(fs / "_build/x")
```

talex5:
On eio_posix, `fstat fs` failed because it called `fstatat(AT_FDCWD, "", ...)` and that doesn't allow empty paths. 

The other cases look OK:

- `fstat cwd` was OK because our path resolving logic handles that.
- Eio_linux was OK because we have `AT_EMPTY_PATH` there.

This was breaking `mkdirs` if only the base directory already existed.

Fixes #906, reported by @samoht.

(Windows has some problems here too, but I think @avsm is rewriting its path handling at the moment anyway)

### DIFF
--- a/lib_eio_linux/fs.ml
+++ b/lib_eio_linux/fs.ml
@@ -35,7 +35,7 @@ let with_dir dir_fd path fn =
     ~access:`R
     ~perm:0
     ~flags:Uring.Open_flags.(cloexec + path + directory)
-    dir_fd (if path = "" then "." else path)
+    dir_fd path
   |> fn
 
 module rec Dir : sig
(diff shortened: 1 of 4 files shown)


avsm (approved):

talex5:
I also updated several other places with similar problems and added some more tests for that.

PR #911, opened 2026-07-30
posix: add device mknod support

avsm:
This allows for the creation of a range of devices more so than just FIFOs.

Draft PR until it runs over the BSD CIs to test there, only manually tested on macOS and Linux


avsm:
Works across the board now!

lib_eio_posix/eio_posix_stubs.c
```diff
@@ -19,6 +19,9 @@
 #endif
 #include <sys/uio.h>
 #include <sys/stat.h>
+#ifdef __linux__
+#include <sys/sysmacros.h>
```
avsm: This is the best I could find; I'm surprised that Linux needs a special include but its like this since glibc 2.28 https://www.man7.org/linux/man-pages/man3/makedev.3.html

lib_eio_posix/low_level.ml
```diff
@@ -486,6 +486,47 @@ let symlink ~link_to new_dir new_path =
...
+    if major t <> maj || minor t <> mnr then
+      Fmt.invalid_arg
+        "Dev.make: %d:%d is out of range for this platform (it became %a)"
+        maj mnr pp t;
+    t
+end
+
+external eio_mknodat : Unix.file_descr -> string -> int -> Dev.t -> unit = "caml_eio_posix_mknodat"
+
+type node_kind = [ `Fifo | `Sock | `Reg | `Chr of Dev.t | `Blk of Dev.t ]
```
talex5: These should probably match the names in `Eio.File.Stat.kind` (e.g. `Regular_file`).

lib_eio_posix/low_level.mli
```diff
@@ -44,6 +44,31 @@ val lseek : fd -> Optint.Int63.t -> [`Set | `Cur | `End] -> Optint.Int63.t
...
+        to be represented on this platform. *)
+
+  val major : t -> int
+  (** [major t] is the major number of [t]. *)
+
+  val minor : t -> int
+  (** [minor t] is the minor number of [t]. *)
+
+  val to_int64 : t -> int64
+  (** [to_int64 t] is the raw [dev_t] value of [t], as the OS represents it. *)
```
talex5: Needs a `from_int64` too.

(and in the past I've had a `from_raw` for Wayland, which doesn't assume devices fit in 64 bits: https://github.com/talex5/libdrm-ocaml/blob/main/lib/dev_t.mli)

talex5 (commented):
It's strange that this is only for eio_posix and not available with eio_linux. Probably it needs moving so it can be shared.

`Eio.File.Stat.t` uses `Int64.t` for devices. We should probably use the same type here. We could add an `Eio.File.Dev.t` type (just an alias to int64 at first, then maybe private later).

The conversions to/from major/minor pairs could go in `Eio_unix` somewhere.

avsm:
Not sure where this should go so it can work with Eio_linux as well. `eio.unix` would be pulled in on Windows too so not appropriate. Can Eio_linux depend on Eio_posix?

lib_eio_posix/low_level.ml
avsm: Yeah, the problem with exposing this in Eio.File is that it should be portable to Windows as well, so now I'm spelunking in there to find out what's going on :-)

lib_eio_posix/low_level.mli
avsm: Oh the Wayland >uint64_t is a new one on me, I'll look at that...

talex5: Actually maybe it's not that it's larger, just that the size varies by platform, so it sends it as an array of bytes:

https://wayland.app/protocols/linux-dmabuf-v1#zwp_linux_dmabuf_feedback_v1:event:main_device

lib_eio_posix/low_level.ml
avsm: Done in b8e67693035abf27fbc109de332f02d70f41cc8d, with the slight annoyance that the types aren't exactly the same since the Eio.File.Stat.kind doesn't have a payload for Dev.t (since it can be constructed by getdents which doesn't supply the full device number)

lib_eio_posix/low_level.mli
avsm: Done in b8e67693035abf27fbc109de332f02d70f41cc8d

avsm:
I've moved it into Eio_unix in b8e67693035abf27fbc109de332f02d70f41cc8d. One breaking change to look at is that the Eio.File.t now switches to a Dev.t, and also a Dev.t option for rdev.  This is because the Windows Unix layer in OCaml sets rdev to dev during its 'emulation'. It seems better for our portable layer to explicitly mark this as None. For the dev layer, it is indeed set to the volume serial number on Windows, so that works well as a private int64

avsm:
Rebased and tweaked to use alcotest in eio_posix as well

lib_eio_linux/low_level.ml
```diff
@@ -721,6 +722,14 @@ let chmod ~follow ~mode dirfd path =
       )
   with Unix.Unix_error (code, name, arg) -> raise @@ Err.v code name arg
 
+let mknod kind ~perm dirfd path =
+  try
+    with_parent_dir_fd dirfd path @@ fun parent leaf ->
+    Eio_unix.run_in_systhread ~label:"mknod" (fun () ->
+        Eio_unix.Private.mknod parent leaf ~kind ~perm
+      )
+  with Unix.Unix_error (code, name, arg) -> raise @@ Err.v code name arg
```
talex5: This can be moved to `Eio_unix`. Not all functions there do this yet, but that was the idea behind adding `Eio_unix.Err`.

lib_eio_posix/low_level.mli
```diff
@@ -58,6 +58,7 @@ external uid     : stat -> (int64 [@unboxed]) = "ocaml_eio_posix_stat_uid_bytes"
 external gid     : stat -> (int64 [@unboxed]) = "ocaml_eio_posix_stat_gid_bytes" "ocaml_eio_posix_stat_gid_native" [@@noalloc]
 external ino     : stat -> (int64 [@unboxed]) = "ocaml_eio_posix_stat_ino_bytes" "ocaml_eio_posix_stat_ino_native" [@@noalloc]
 external size    : stat -> (int64 [@unboxed]) = "ocaml_eio_posix_stat_size_bytes" "ocaml_eio_posix_stat_size_native" [@@noalloc]
+
```
talex5, line 61: Was this blank line intentional?

lib_eio_posix/low_level.mli
```diff
@@ -83,6 +84,11 @@ val symlink : link_to:string -> dir_fd -> string -> unit
 (** [symlink ~link_to dir path] will create a new symlink at [dir / path]
     linking to [link_to]. *)
 
+val mknod : Eio_unix.Private.node_kind -> perm:int -> dir_fd -> string -> unit
+(** [mknod kind ~perm dir path] creates a fs node of type [kind] at [dir / path].
+
+    Creating a device node normally requires privileges. *)
```
talex5: Not for pipes, though.

talex5 (commented):
Looks pretty good!

I think it would be useful to add `Eio.Path.mknod` in the cross platform API too. Otherwise, it's hard to use this without forcing the choice of eio_linux or not. And it would avoid needing to duplicate the tests.

PR #913, opened 2026-07-30
Move path operations to backends

talex5:
This is split off from @avsm's #908 to make reviewing easier. This PR just moves the existing code without changing the logic or adding Windows support or new tests.

Compared to #908:
- I made `PATH` implement `split` and `join` and all the existing logic moves there. The original PR had some handling of `.` in the frontend, which seemed odd.
- `DIR` now has `module Path : PATH`, not `include PATH`, as that seemed a bit cleaner.
- `posix_path.mli` now just does `include Eio.Fs.Pi.PATH`.
- `tests/fs.md` just tests `Posix_path` directly. There doesn't seem any point running a main loop and a backend now that the frontend is just calling the functions directly, and the tests really are checking POSIX syntax, not the host syntax.

### DIFF
--- a/lib_eio_linux/fs.ml
+++ b/lib_eio_linux/fs.ml
@@ -169,6 +169,8 @@ end = struct
 
   let native t path =
     Some (native_internal t path)
+
+  include Eio_utils.Posix_path
 end
 and Dir_handler : sig
   val v : (Dir.t, [`Dir | `Close]) Eio.Resource.handler
--- a/lib_eio/fs.ml
+++ b/lib_eio/fs.ml
@@ -50,6 +50,15 @@ type 'a dir = ([> dir_ty] as 'a) r
 (** {2 Provider Interface} *)
 
 module Pi = struct
+
+  module type PATH = sig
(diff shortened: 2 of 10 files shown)


avsm:
Thanks as always for the review!

> I made PATH implement split and join and all the existing logic moves there. The original PR had some handling of . in the frontend, which seemed odd.

This was because I wasn't entirely sure if that was part of the frontend Eio guarantee or not (as #907 was going on). It is cleaner to have them all in the backend as long.

lib_eio_linux/fs.ml
```diff
@@ -169,6 +169,8 @@ end = struct
 
   let native t path =
     Some (native_internal t path)
+
+  module Path = Eio_utils.Posix_path
```
avsm: I thought `include PATH` was cleaner here since it doesn't add yet another name (Path) that's unused for any purpose in the module implementation of the signature.  I have no strong feelings either way though

talex5: Thinking about it more, I agree with you. I've reverted that bit.

PR #916, opened 2026-07-31
Implement Eio.Path.native as a Pi backend and add Windows support

talex5:
This is split off from @avsm's #908 and rebased on top of #913 and #915.

Incorporates the approach from ocaml-multicore/eio#738.

I haven't reviewed it yet.

### DIFF
--- a/lib_eio/utils/eio_utils.ml
+++ b/lib_eio/utils/eio_utils.ml
@@ -7,3 +7,4 @@ module Suspended = Suspended
 module Zzz = Zzz
 module Dla = Dla
 module Posix_path = Posix_path
+module Nt_path = Nt_path
--- a/lib_eio/utils/nt_path.ml
+++ b/lib_eio/utils/nt_path.ml
@@ -1,0 +1,99 @@
+(* Windows path syntax. *)
+
+let is_drive_letter = function 'A' .. 'Z' | 'a' .. 'z' -> true | _ -> false
+let is_sep c = c = '\\' || c = '/'
+
+(* A recognizer matches at position [i] of [s] and returns the position
+   one past the match, or [None] if it doesn't match. *)
+type recognizer = string -> int -> int option
+
+let charp f : recognizer =
+  fun s i -> if i < String.length s && f s.[i] then Some (i + 1) else None
+
+let ( *> ) (p : recognizer) (q : recognizer) : recognizer =
+  fun s i -> Option.bind (p s i) (q s)
+
+let ( <|> ) (p : recognizer) (q : recognizer) : recognizer =
+  fun s i -> match p s i with Some _ as r -> r | None -> q s i
+
+let opt p : recognizer = p <|> (fun _ i -> Some i)
+
+(* Zero or more matches of [p]. *)
+let rec many p : recognizer = fun s i ->
+  match p s i with Some j -> many p s j | None -> Some i
+
+let chr c = charp (Char.equal c)
+let sep = charp is_sep
+let bslash = chr '\\'
+let qmark = chr '?'
+
+(* The (possibly empty) run of non-separator characters at [i]. *)
+let component = many (charp (fun c -> not (is_sep c)))
+
+(* A component whose text satisfies [f]. *)
+let component_is f : recognizer = fun s i ->
+  match component s i with
+  | Some e when f (String.sub s i (e - i)) -> Some e
+  | _ -> None
+
+(* The component after this one. [opt sep] rather than [sep] so that a
+   malformed prefix is cut at whatever exists. *)
+let next = opt sep *> component
+
+let drive = charp is_drive_letter *> chr ':'
+let q_or_dot = component_is (fun c -> c = "?" || c = ".")
+let unc_kw = component_is (fun c -> String.uppercase_ascii c = "UNC")
+
+(* The "C:", "\\server\share", "\\.\device", "\\?\..." or "\??\..."
+   volume prefix of a path. *)
+let volume_prefix =
+  drive                                                  (* C: *)
+  <|> (sep *> sep *> q_or_dot *> opt sep *> unc_kw
+       *> next *> next)                                  (* \\?\UNC\server\share *)
+  <|> (sep *> sep *> component *> next)                  (* \\server\share, \\?\C: or \\.\device *)
+  <|> (bslash *> qmark *> qmark *> component *> next)    (* \??\C: - the NT object-manager form (backslash only) *)
+
+let volume_end s = Option.value (volume_prefix s 0) ~default:0
+let volume s = String.sub s 0 (volume_end s)
+
+(* Win32 does no normalization in the verbatim and NT namespaces. *)
+let verbatim_prefix = bslash *> (bslash <|> qmark) *> qmark
+let verbatim s = Option.is_some (verbatim_prefix s 0)
+
+(* [is_relative p] is [true] unless [p] begins with a volume or a separator. *)
+let is_relative s = Option.is_none ((volume_prefix <|> sep) s 0)
+
+let split p =
+  let vend = volume_end p in
+  let sep_at = if verbatim p then Char.equal '\\' else is_sep in
+  let sep_at i = sep_at p.[i] in
+  (* Trailing separators are ignored; one is kept for a bare root. *)
+  let rec trim i = if i > vend + 1 && sep_at (i - 1) then trim (i - 1) else i in
+  let stop = trim (String.length p) in
+  if stop <= vend || (stop = vend + 1 && sep_at vend) then None
+  else
+    let rec rsep i = if i < vend then None else if sep_at i then Some i else rsep (i - 1) in
+    match rsep (stop - 1) with
+    | None -> Some (volume p, String.sub p vend (stop - vend))
+    | Some idx ->
+      let basename = String.sub p (idx + 1) (stop - idx - 1) in
+      let dirname =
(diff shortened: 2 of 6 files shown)


lib_eio/utils/test/test_paths.ml
```diff
@@ -0,0 +1,134 @@
...
+    check_split "rooted" (W.split "\\x") (Some ("\\", "x"));
+    check_split "unc" (W.split "\\\\srv\\share\\x") (Some ("\\\\srv\\share\\", "x"));
+    check_split "unc root" (W.split "\\\\srv\\share") None;
+    (* In verbatim paths, '/' is an ordinary character. *)
+    check_split "verbatim slash literal" (W.split "\\\\?\\C:\\a/b")
+      (Some ("\\\\?\\C:\\", "a/b"));
+    check_split "nt" (W.split "\\??\\C:\\a\\b") (Some ("\\??\\C:\\a", "b"));
+    (* The NT prefix is backslash-only: with forward slashes this is just a
+       rooted path with ordinary components. *)
+    check_split "not nt with fwd seps" (W.split "/??/C:/x") (Some ("/??/C:", "x"));
```
talex5: If `/` is just an ordinary character, shouldn't the basename here be `/x`? We had `a/b` as a basename above.

(just guessing; I don't know much about Windows paths)

lib_eio/utils/test/test_paths.ml
```diff
@@ -0,0 +1,134 @@
...
+    (* A drive prefix means nothing in POSIX syntax. *)
+    check_bool "drive" (P.is_step "C:\\x")
+
+  let test_join () =
+    check_str "join" (P.join "a" "b") "a/b";
+    check_str "join root" (P.join "/" "a") "/a";
+    check_str "join empty" (P.join "" "b") "b";
+    check_str "join trailing" (P.join "a/" "b") "a/b"
+
+  let test_split () =
```
talex5: Note we already have a bunch of POSIX `split` tests in fs.md, e.g.

https://github.com/ocaml-multicore/eio/blob/8a2052cded62ea3c91bfeff2f5161a3561b8a17f/tests/fs.md#L280-L284

avsm:
I think this is broken now on Windows due to the changes in `step` handling from #913; I'll fix it up first before you need to review.

lib_eio/fs.ml
```diff
@@ -52,6 +52,10 @@ type 'a dir = ([> dir_ty] as 'a) r
 module Pi = struct
 
   module type PATH = sig
+    val is_step : string -> bool
+    (** [is_step s] is [true] if [s] extends a path when joined by {!Path.( / )},
+        and [false] if it would replace it (e.g. an absolute path). *)
```
talex5: I don't like the name `step` here. It suggests a single step / component, but in fact it's another path.

(this confusion might be due to the comment on `Path.(/)` which calls the second argument `step` - that should probably be renamed to avoid confusion)

lib_eio/fs.ml
```diff
@@ -84,7 +88,9 @@ module Pi = struct
     val chmod : t -> follow:bool -> perm:File.Unix_perm.t -> path -> unit
     val chown : follow:bool -> ?uid:int64 -> ?gid:int64 -> t -> path -> unit
     val pp : t Fmt.t
+
     val native : t -> string -> string option
+    (** [native t p] is the OS-native path for [p] within [t], if available. *)
```
talex5: Not sure it's worth commenting these. The meaning is the same as in the main API and they need to be kept in sync.

lib_eio/utils/test/test_paths.ml
```diff
@@ -0,0 +1,137 @@
...
+    check_split "nt" (W.split "\\??\\C:\\a\\b") (Some ("\\??\\C:\\a", "b"));
+    (* The NT prefix is backslash-only: with forward slashes this is just a
+       rooted path with ordinary components. *)
+    check_split "not nt with fwd seps" (W.split "/??/C:/x") (Some ("/??/C:", "x"));
+    check_split "empty" (W.split "") None
+
+  let test_join () =
+    check_str "empty" (W.join "a" "") "a\\";
+    check_str "abs" (W.join "a" "C:\\b") "C:\\b";
+    check_str "dot" (W.join "." "b") "b";
```
talex5: I added a few more tests, now that the `join` logic has moved to the backend.

talex5:
> I think this is broken now on Windows due to the changes in step handling from https://github.com/ocaml-multicore/eio/pull/913; I'll fix it up first before you need to review.

Should be fixed now (I spotted it too).

lib_eio/fs.ml
avsm: I'm trying to get across the fact that it wont replace the root, but some paths in Windows are really confusing (like `C:foo` since they are relative but still rooted in a volume.  Definitely open to new names though, and I wonder if we should just simplify all this without trying to support every last corner case.

avsm: I've discovered an important fact! I thought that Windows maintains a _per drive_ cwd (e.g. `C:foo` and `D:foo` could be different relative paths on their respective drives).  However, it appears to be [legacy behaviour](https://devblogs.microsoft.com/oldnewthing/20101011-00/?p=12563), and language frameworks like Powershell and .NET maintain a single per-process cwd.  This implies in turn that we could reject drive-relative paths and just assume a process-global cwd. I'm spelunking into Powershell in my next hacking session to figure out what it does with `C:foo`.  If we get rid of that support, then the rest of this patch is greatly simplifable.

talex5: Ah, I thought that too! Now I'm curious what the win32 APIs do with a path like `C:foo`. Does it just treat it as `C:\foo`? If so, the old `is_relative` name still works.

However, we're not actually using `is_step` for anything except the unit-tests, so I've just removed it from the public API for now.

lib_eio/utils/test/test_paths.ml
talex5: I merged the new POSIX tests into fs.md, which removes some duplicates (#918). We could move the NT ones too, or leave them here (I don't mind).

avsm:
I've been reading up more on Windows paths and have a commit to slightly simplify the nt path parser, but also split out the tests. Sadly my Windows machine keeps rebooting for reasons unknown so I'll have to push it once I get back to it in person later today I hope :-)

lib_eio/fs.ml
avsm: As far as I can tell, Win32 uses the process cwd _or_ queries some hidden env variables (maintained by cmd.exe)

i.e.:
 - If the process cwd is already on C:, then `C:foo = <cwd>\foo`
 - win32 reads the hidden env var `=C:` and joins against that. In my test the process cwd was `D:\` 

From our perspective, I think a drive-relative path can never be resolved relative to a dirfd in win32 due to the second case, so treating it as rooted (and so rejecting it in a sandboxed operation) is the way. I could really do with a sanity check of this from @dra27!

Agreed that is_step is no longer neededd in public interface since the move of the logic from the Pi `join`

lib_eio/utils/test/test_paths.ml
avsm: I think it's right as-is; we scan from the right and select the first separator (in this case, it's the `/` before `x).

In the earlier one, `a/b\c` we select `\` as the first separator (since Windows) and end up with `c` as the basename.

lib_eio/fs.ml
avsm: I've renamed to is_relative in 7e16678e2f4c01c7469de641b80ea2507db93c46

avsm:
With apologies for the code churn, I've moved to mdx for the windows tests in 1fae694cbc0a4d4f9908d14f3e3ac4f4ab73102d which is a bit clearer to read than the denser alcotest. I'm trying to get all the portable mdx working again on Windows, but this is just the fs bit. It also drops an alcotest dep on Eio as a result.

lib_eio/utils/test/test_paths.ml
talex5: Ah right, I thought "verbatim path" and "NT prefix is backslash-only" both meant that `/` wasn't a separator. But I see it means that the prefix itself is backslash only.

I hadn't noticed the difference between `\\?\` and `\??\`.

avsm:
my CRLF is messed up, sorry! Can revert 1fae694cbc0a4d4f9908d14f3e3ac4f4ab73102d (but my Windows machine has just rebooted again!!)

talex5:
No problem; I squashed them and converted the newlines in `fs.md` (I left the existing test files alone).

Hmm, that doesn't work (failed in CI). I've switched it back to CRLF.

talex5:
I think we've hit https://github.com/realworldocaml/mdx/pull/294 here.

> Some git installations are configured to translate newlines, some do not. If they are translated, the LFs in this repository will be converted to CR LF and the tests will fail. Conversely this translation can also hide problems in the other direction by making CRs disappear.

I tried using `.gitattributes` to force it to `lf`, which didn't work. Then I forced it to `crlf`, that didn't work either.

talex5:
> Can revert https://github.com/ocaml-multicore/eio/commit/1fae694cbc0a4d4f9908d14f3e3ac4f4ab73102d (but my Windows machine has just rebooted again!!)

I partly reverted this, but moved the tests of `Nt_path` to the main `tests/fs.md`, next to the POSIX ones. That means they get tested on Linux and macOS at least. The remaining tests (that use `Eio_windows.run`) are still using Alcotest and only run on Windows.

PR #919, opened 2026-08-05
Add Eio.Path.of_dir

talex5:
This is perhaps slightly clearer than creating a `(dir, "")` tuple manually, and avoids having to think about whether to use "." or "".

It also opens up the possibility of making `Path.t` `private` at some point. The extra variance annotations are just to make testing that easier.

I updated the backends to use it for `fs` and `cwd` and made them a bit more similar to each other.


lib_eio_posix/fs.ml
```diff
@@ -164,6 +164,7 @@ end = struct
     ]
 end
 
-(* Full access to the filesystem. *)
-let fs = Eio.Resource.T (Dir.v ~label:"fs" ~path:"." Fs, Handler.v)
-let cwd = Eio.Resource.T (Dir.v ~label:"cwd" ~path:"." Cwd, Handler.v)
+let dir ~label ~path fd = Eio.Resource.T (Dir.v ~label ~path fd, Handler.v)
+
+let fs = Eio.Path.of_dir (dir ~label:"fs" ~path:"." Fs)
```
avsm, line 169: The use of "." in the posix backend and "" in the Linux one probably deserves a comment to explain it.

avsm (approved):

lib_eio_posix/fs.ml
talex5: I don't think it actually matters (the tests pass either way). The only thing that uses it is generating native paths, and `native_internal` checks for "." and treats it as "" anyway. But it was like that before - it's not part of this change. I'll make a separate PR cleaning it up.

PR #920, opened 2026-08-05
Clean up dir_path handling

talex5:
It wasn't clear whether `dir_path` would be "" or "." for the current directory:

- `eio_linux` was using "" for `cwd` and `fs`.
- `eio_posix` was using "." for `cwd` and `fs`.

Force it to "" in both cases so there are fewer things to test.

The only difference should be printing of native paths for the subtree "./.":

    ./          # Before with eio_linux
    ././        # Before with eio_posix
    .           # After (both)

(from https://github.com/ocaml-multicore/eio/pull/919#discussion_r3722069825)

Note: I didn't touch `eio_windows` as I don't have an easy way to test that.

### DIFF
--- a/lib_eio_linux/fs.ml
+++ b/lib_eio_linux/fs.ml
@@ -49,11 +49,14 @@ module rec Dir : sig
 end = struct
   type t = {
     fd : Low_level.dir_fd;
+    dir_path : string;
     label : string;
-    path : string;
   }
 
-  let v ~label ~path fd = { fd; label; path }
+  let v ~label ~path:dir_path fd =
+    (* Avoid having "" and "." meaning the same thing, so there are fewer cases to test *)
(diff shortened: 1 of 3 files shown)


lib_eio_linux/fs.ml
```diff
@@ -49,11 +49,14 @@ module rec Dir : sig
 end = struct
   type t = {
     fd : Low_level.dir_fd;
+    dir_path : string;
     label : string;
-    path : string;
```
talex5, line 53: I renamed `path` to `dir_path` and re-ordered the record so it's the same as in `eio_posix`.

avsm (approved):
This looks fine. For Windows I believe "" means relative, and I'll have a test for that once I submit the forthcoming Windows Eio.Process PR...

PR #921, opened 2026-08-18
Add "available" to opam files

talex5:
- `eio_windows` should only install on Windows. This was previously commented out due to a CI limitation, but ocaml-ci now has Windows support.
- `eio_posix` should not install on Windows.

### DIFF
--- a/eio_posix.opam
+++ b/eio_posix.opam
@@ -32,6 +32,7 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/ocaml-multicore/eio.git"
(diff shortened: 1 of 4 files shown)


avsm (approved):
Grudgingly approved, since I've never understood how to `opam pin .` with these availability constraints. Do people really individually `pin add -n` every single package?

talex5:
It doesn't affect pinning for me. I usually use:
```
opam pin add -yn --with-version=dev .
opam install eio_main
```

Issue #922, opened 2026-08-19
Access environment variables from Eio


patricoferris:
Although they probably do not need to be made non-blocking, it would be useful for Eio to expose a portable interface for accessing environment variables. I am thinking of something like:

```ocaml
open Std

type ty = [ `System ]
type 'a t = ([> ty] as 'a) r

val env : _ t -> (string * string) list
(** [env system] gets the full list of environment variables. *)

val get_env : _ t -> string -> string
(** [get_env system name] will get the environment variable called [name].
    If it does not exist, this function raises [Not_found]. *)

val put_env : _ t -> name:string -> value:string -> unit
(** [put_env system ~name ~value] adds [name=value] to the environment. *)

module Pi : sig
  module type SYSTEM = sig
    type t

    val env : t -> (string * string) list

    val get_env : t -> string -> string

    val put_env : t -> name:string -> value:string -> unit
  end

  val system : (module SYSTEM with type t = 't) -> ('t, ty) Resource.handler
end
```

Other things could get added to this system interface in the future perhaps (e.g., some signal handling logic). I have a branch which implements this interface, here is the diff: https://github.com/ocaml-multicore/eio/compare/main...patricoferris:eio:eio-system -- happy to open that as PR or discuss the design and whether this feature is wanted more here.

talex5:
Managing environment variables seems reasonable to me. I'm not keen on the name `System` though: `Stdenv` itself represents the system. But we can certainly have more things in there, including environment variables. `env#env` would be a bit confusing though. Maybe `env#vars`?

patricoferris:
> I'm not keen on the name System though: Stdenv itself represents the system. But we can certainly have more things in there, including environment variables. env#env would be a bit confusing though. Maybe env#vars?

Agreed! I've opened the linked PR using the  `Vars` terminology, also makes the functions nicer e.g. `Eio.Vars.get` :-)

avsm:
How could you resist not calling it `env#comonad` ?! https://bartoszmilewski.com/2017/01/02/comonads/

patricoferris:
Completed in #923 #930

PR #923, opened 2026-08-19
Add environment functions to Eio.Process

patricoferris:
Access to environment variables after some discussion in #922

### DIFF
--- a/lib_eio/process.mli
+++ b/lib_eio/process.mli
@@ -52,6 +52,8 @@ type 'a mgr = 'a r
  constraint 'a = [> [> `Generic] mgr_ty]
 (** A process manager capable of spawning new processes. *)
 
+(** {2 Environment variables} *)
+
 module Env : sig
   (** A list of environment variable entries.
 
@@ -79,9 +81,8 @@ module Env : sig
       This just adds [xs] to {!empty} using {!override}. *)
 
   val get_opt : string -> t -> string option
-  (** [get_opt name t] is the value of [name] in [t], or [None] if there is no such binding.
-
-      @raise Invalid_argument if [name] is not a valid name. *)
+  (** [get_opt name t] is the value of [name] in [t], or [None] if there is no such binding
+      (or if [name] is not a valid name). *)
 
   val override : (string * string option) list -> t -> t
   (** [override bindings t] is a new environment which is like [t]
@@ -108,6 +109,13 @@ module Env : sig
 
   val pp : t Fmt.t
 end
+
+val environment : _ mgr -> Env.t
+(** [environment t] returns a snapshot of this process's current environment. *)
+
+val getenv_opt : _ mgr -> string -> string option
+(** [getenv_opt t name] will get the environment variable called [name].
+    Returns [None] if [name] does not exist. *)
 
 (** {2 Processes} *)
 
@@ -227,6 +235,9 @@ module Pi : sig
   module type MGR = sig
     type tag
     type t
+
+    val environment : t -> Env.t
+    val getenv_opt : t -> string -> string option
 
     val pipe :
       t ->
--- a/lib_eio/unix/process.ml
+++ b/lib_eio/unix/process.ml
@@ -136,6 +136,9 @@ end) = struct
 
   type tag = [ `Generic | `Unix ]
 
+  let environment _t = Unix.environment () |> Eio.Process.Env.of_array
+  let getenv_opt _t name = Sys.getenv_opt name
+
   let pipe _ ~sw =
     (Private.pipe sw :> ([Eio.Resource.close_ty | Eio.Flow.source_ty] r *
     [Eio.Resource.close_ty | Eio.Flow.sink_ty] r))
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -165,6 +165,9 @@ module Pi = struct
     type tag
     type t
 
+    val environment : t -> Env.t
+    val getenv_opt : t -> string -> string option
+
     val pipe :
       t ->
       sw:Switch.t ->
@@ -222,6 +225,16 @@ let signal (type tag) (t : [> tag ty] r) s =
   let (Resource.T (v, ops)) = t in
   let module X = (val (Resource.get ops Pi.Process)) in
   X.signal v s
+
+let environment (type tag) (t : [> tag mgr_ty] r) =
+  let (Resource.T (v, ops)) = t in
+  let module X = (val (Resource.get ops Pi.Mgr)) in
+  X.environment v
+
+let getenv_opt (type tag) (t : [> tag mgr_ty] r) =
+  let (Resource.T (v, ops)) = t in
+  let module X = (val (Resource.get ops Pi.Mgr)) in
+  X.getenv_opt v
 
 let spawn (type tag) ~sw (t : [> tag mgr_ty] r) ?cwd ?stdin ?stdout ?stderr ?env ?executable args : tag ty r =
   let (Resource.T (v, ops)) = t in
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -93,13 +93,15 @@ module Env = struct
     Printf.sprintf "%s=%s" name value
 
   let get_opt name t =
-    Name.validate name;
-    let prefix = name ^ "=" in
-    Array.find_opt (Name.starts_with ~prefix) t
-    |> Option.map (fun e ->
-        let i = String.length prefix in
-        String.sub e i (String.length e - i)
-      )
+    if name = "" then None
+    else (
+      let prefix = name ^ "=" in
+      Array.find_opt (Name.starts_with ~prefix) t
+      |> Option.map (fun e ->
+          let i = String.length prefix in
+          String.sub e i (String.length e - i)
+        )
+    )
 
   let override bindings t =
     List.iter validate_binding bindings;
--- a/lib_eio_linux/tests/spawn.md
+++ b/lib_eio_linux/tests/spawn.md
@@ -7,8 +7,6 @@ open Eio.Std
 
 module Env = Eio.Process.Env
 module Process = Eio_linux.Low_level.Process
-
-let default_env = Unix.environment () |> Env.of_array
 ```
 
 ## Spawning processes
@@ -31,7 +29,8 @@ FOO=bar
 Changing directory:
 
 ```ocaml
-# Eio_linux.run @@ fun _env ->
+# Eio_linux.run @@ fun env ->
+  let default_env = Eio.Stdenv.process_mgr env |> Eio.Process.environment in
   Switch.run @@ fun sw ->
   let child = Process.spawn ~sw Process.Fork_action.[
     chdir "/";
@@ -47,7 +46,8 @@ Changing directory:
 Changing directory using a file descriptor:
 
 ```ocaml
-# Eio_linux.run @@ fun _env ->
+# Eio_linux.run @@ fun env ->
+  let default_env = Eio.Stdenv.process_mgr env |> Eio.Process.environment in
   Switch.run @@ fun sw ->
   let root =
     Eio_linux.Low_level.openat2 ~sw "/"
@@ -71,7 +71,8 @@ Changing directory using a file descriptor:
 Exit status:
 
 ```ocaml
-# Eio_linux.run @@ fun _env ->
+# Eio_linux.run @@ fun env ->
+  let default_env = Eio.Stdenv.process_mgr env |> Eio.Process.environment in
   Switch.run @@ fun sw ->
   let child = Process.spawn ~sw Process.Fork_action.[
     execve "/usr/bin/env"
@@ -85,7 +86,8 @@ Exit status:
 Failure starting child:
 
 ```ocaml
-# Eio_linux.run @@ fun _env ->
+# Eio_linux.run @@ fun env ->
+  let default_env = Eio.Stdenv.process_mgr env |> Eio.Process.environment in
   Switch.run @@ fun sw ->
   Process.spawn ~sw Process.Fork_action.[
     chdir "/idontexist";
@@ -99,7 +101,8 @@ Exception: Unix.Unix_error(Unix.ENOENT, "chdir", "")
 Signalling a running child:
 
 ```ocaml
-# Eio_linux.run @@ fun _env ->
+# Eio_linux.run @@ fun env ->
+  let default_env = Eio.Stdenv.process_mgr env |> Eio.Process.environment in
   Switch.run @@ fun sw ->
   let child =
     Process.spawn ~sw Process.Fork_action.[
--- a/lib_eio_posix/test/spawn.md
+++ b/lib_eio_posix/test/spawn.md
@@ -7,8 +7,6 @@ open Eio.Std
 
 module Env = Eio.Process.Env
 module Process = Eio_posix.Low_level.Process
-
-let default_env = Unix.environment () |> Env.of_array
 ```
 
 ## Spawning processes
@@ -31,7 +29,8 @@ FOO=bar
 Changing directory:
 
 ```ocaml
-# Eio_posix.run @@ fun _env ->
+# Eio_posix.run @@ fun env ->
+  let default_env = Eio.Stdenv.process_mgr env |> Eio.Process.environment in
   Switch.run @@ fun sw ->
   let child = Process.spawn ~sw Process.Fork_action.[
     chdir "/";
@@ -47,7 +46,8 @@ Changing directory:
 Changing directory using a file descriptor:
 
 ```ocaml
-# Eio_posix.run @@ fun _env ->
+# Eio_posix.run @@ fun env ->
+  let default_env = Eio.Stdenv.process_mgr env |> Eio.Process.environment in
   Switch.run @@ fun sw ->
   let root = Eio_posix.Low_level.openat ~sw ~mode:0 Fs "/" Eio_posix.Low_level.Open_flags.(rdonly + directory) in
   let child = Process.spawn ~sw Process.Fork_action.[
@@ -64,7 +64,8 @@ Changing directory using a file descriptor:
 Exit status:
 
 ```ocaml
-# Eio_posix.run @@ fun _env ->
+# Eio_posix.run @@ fun env ->
+  let default_env = Eio.Stdenv.process_mgr env |> Eio.Process.environment in
   Switch.run @@ fun sw ->
   let child = Process.spawn ~sw Process.Fork_action.[
     execve "/usr/bin/env"
@@ -78,7 +79,8 @@ Exit status:
 Failure starting child:
 
 ```ocaml
-# Eio_posix.run @@ fun _env ->
+# Eio_posix.run @@ fun env ->
+  let default_env = Eio.Stdenv.process_mgr env |> Eio.Process.environment in
   Switch.run @@ fun sw ->
   Process.spawn ~sw Process.Fork_action.[
     chdir "/idontexist";
@@ -92,7 +94,8 @@ Exception: Unix.Unix_error(Unix.ENOENT, "chdir", "")
 Signalling a running child:
 
 ```ocaml
-# Eio_posix.run @@ fun _env ->
+# Eio_posix.run @@ fun env ->
+  let default_env = Eio.Stdenv.process_mgr env |> Eio.Process.environment in
   Switch.run @@ fun sw ->
   let child =
     Process.spawn ~sw Process.Fork_action.[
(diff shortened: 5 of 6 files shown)


lib_eio/unix/vars.ml
```diff
@@ -0,0 +1,19 @@
+type t = unit
+
+let split_in_two c s =
+  try
+    let first_occurence = String.index s c in
+    let first = String.sub s 0 first_occurence in
+    let second = String.sub s (first_occurence + 1) (String.length s - first_occurence) in
```
avsm:
```suggestion
    let first_occurrence = String.index s c in
    let first = String.sub s 0 first_occurrence in
    let second = String.sub s (first_occurrence + 1) (String.length s - first_occurrence) in
```

minor typo

avsm: Does this work for `A=b` ? Looks like there might be an off-by-one. Should probably have a testcase for get_all

lib_eio/vars.ml
```diff
@@ -0,0 +1,39 @@
...
+  let module X = (val (Resource.get ops Pi.Vars)) in
+  X.get_all t
+
+let get t name =
+  let (Resource.T (t, ops)) = t in
+  let module X = (val (Resource.get ops Pi.Vars)) in
+  try X.get t name
+  with Not_found as ex ->
+    let bt = Printexc.get_raw_backtrace () in
+    Exn.reraise_with_context ex bt "getting environment variable %s" name
```
avsm: Does this do anything if you're reraising a Not_found? The context only attaches to an Eio.Exn.Io I thought.

tests/vars.md
```diff
@@ -0,0 +1,34 @@
...
+- : string = "hello"
+```
+
+Getting environment variables that do not exist works:
+
+```ocaml
+# run @@ fun ~vars ->
+  Eio.Vars.get vars "THIS_ENV_PROBABLY_WILL_NOT_EXIST";;
+Exception: Not_found.
+```
```
avsm, line 44: Could use a get_all test here, and also just a couple of weird test cases (like A=b=c)

avsm (commented):
Some high level thoughts:

- getenv/putenv aren't multithreaded safe, so is this a good opportunity to provide domain safety by introducing a mutex, so that if accesses go via Eio.Vars they are synchronised? We cant do anything about direct uses of Unix but this is better than nothing
- bear in mind that secure_getenv will result in a null if setuid binary, but I think this is handled at the Unix library level.
- The Eio.Vars layer is where we should be dealing with portability, so a really useful future addition would be to add a PATH parser using our shiny new Nt_path and Posix_path modules. On Windows it's ; separated and on POSIX : separated.

patricoferris:
Thanks for the review Anil!! All very reasonable to me and I'll incorporate this soon!

patricoferris:
Re:threadsafety-- do you think, in practice, this is a big issue? It seems that the functions are mostly not thread safe when used together across domains. But are a lot of programs doing a lot of "putting"?

avsm:
I think the `putting` mainly happens at start-of-day. I do this, for example, when linking to third party libraries who need environmental configuration. It also happens most obviously at exec time but that's covered by other code.

lib_eio/eio.mli
```diff
@@ -249,6 +254,14 @@ module Stdenv : sig
       To use this, see {!Domain_manager}.
   *)
 
+  val vars : <vars : _ Vars.t as 'a; ..> -> 'a
+  (** [vars t] allows you to access environment variables. *)
```
avsm:
```suggestion
  (** [vars t] allows environment variables to be accessed and mutated. *)
```

lib_eio/vars.mli
```diff
@@ -0,0 +1,29 @@
+open Std
+
+type ty = [ `Vars ]
+type 'a t = ([> ty] as 'a) r
+
+val get_all : _ t -> (string * string) list
+(** [all system] gets the full list of environment variables. *)
+
+val get : _ t -> string -> string
+(** [get_var system name] will get the environment variable called [name].
```
avsm:
```suggestion
(** [get system name] will get the environment variable called [name].
```

lib_eio/vars.mli
```diff
@@ -0,0 +1,29 @@
...
+    If it does not exist, this function raises [Not_found]. *)
+
+val put : _ t -> name:string -> value:string -> unit
+(** [put_var system ~name ~value] adds the variable [name] associated with
```
avsm:
```suggestion
(** [put system ~name ~value] adds the variable [name] associated with
```

patricoferris:
> I think the putting mainly happens at start-of-day. I do this, for example, when linking to third party libraries who need environmental configuration. It also happens most obviously at exec time but that's covered by other code.

If it happens at start of day, do you think there's much chance of hitting race conditions with other domains? 

https://github.com/ocaml-multicore/eio/pull/923/commits/6b4f6d715f3a6de3011fa8d815a2ac18ec3ee5e4 adds `get_path` and `put_path` functions.

avsm:
> If it happens at start of day, do you think there's much chance of hitting race conditions with other domains?

I think if something's exposed in the API, it should be as safe as possible. If we're exposing `putenv`, then why not make it safe with a mutex? If we think it's not used, then it can also just not be included, which I'd prefer to a racy one. `Unix.putenv` still exists in practise for that.

patricoferris:
> I think if something's exposed in the API, it should be as safe as possible. If we're exposing putenv, then why not make it safe with a mutex?

That's fair enough. I have added a mutex in the `eio_unix` implementation which the three backends use: https://github.com/ocaml-multicore/eio/pull/923/commits/8872eeae5af327ddeb275fb8c88b955773ed3d13

avsm (approved):
This looks good to me! I took the liberty of pushing two changes:
- c20fae7c58814d48b0ace99f3cbb38555a2acd4d allocates the mutex once per stdenv, as otherwise each invocation of vars won't synchronise
- 15018f8128cc2b1f2a3b533078d292335a314b39 makes `~sep` a `char` for symmetry

lib_eio/vars.mli
```diff
@@ -0,0 +1,46 @@
...
+
+type ty = [ `Vars ]
+type 'a t = ([> ty] as 'a) r
+
+(** {1 Accessing environment variables} *)
+
+val get_all : _ t -> (string * string) list
+(** [get_all vars] gets the full list of environment variables. *)
+
+val get : _ t -> string -> string
```
talex5, line 11: I suggest either making this `get_opt`, or raising a more useful `Eio.Io` exception (giving the name of the variable that was missing).

lib_eio/vars.mli
```diff
@@ -0,0 +1,46 @@
...
+(** [get vars name] will get the environment variable called [name].
+
+    @raise Not_found if [name] does not exist.*)
+
+val get_path : _ t -> string list
+(** [get_path var] gets the ["PATH"] variable and returns the paths as a list.
+
+    @raise Not_found if ["PATH"] does not exist.*)
```
talex5, line 19: I'm not very keen on this. There are lots of PATH-like variables (e.g. `MANPATH`). Probably this should take the variable name too.

tests/vars.md
```diff
@@ -0,0 +1,86 @@
...
+open Eio.Std
+
+let run (fn : vars:_ Eio.Vars.t -> 'a) =
+  Eio_main.run @@ fun env ->
+  let vars = Eio.Stdenv.vars env in
+  fn ~vars
+
+let try_get t name =
+  try
+    Eio.traceln "%s is %a" name Fmt.(quote string) (Eio.Vars.get t name)
```
talex5, line 17:
```suggestion
    Eio.traceln "%s is %S" name (Eio.Vars.get t name)
```
shorter, and works correctly with strings containing quotes.

lib_eio/unix/eio_unix.mli
```diff
@@ -47,6 +47,9 @@ end
 module Net = Net
 (** Extended network API with support for file descriptors. *)
 
+module Vars = Vars
+(** An {! Eio.System} interface using standard {! Unix} functions. *)
```
talex5, line 51: Should be in `Private` (and remove the `System` bit, or the whole comment).

I'm not even sure it needs to be a module; we could just have `Private.vars` be the actual resource. It can work out the path separator for itself easily enough from `Sys.os_type`. Then each backend can just do `method vars = Eio_unix.Private.vars`.

lib_eio/eio.mli
```diff
@@ -181,6 +181,11 @@ module Fs = Fs
 (** Managing child processes. *)
 module Process = Process
 
+(** {2 Environment Variables} *)
```
talex5, line 184: Could leave out the heading. Putting environment variables under `Processes` seems fine.

lib_eio/vars.ml
```diff
@@ -0,0 +1,48 @@
...
+
+module Pi = struct
+  module type VARS = sig
+    type t
+
+    val get_all : t -> (string * string) list
+    val get : t -> string -> string
+    val get_path : t -> string list
+    val put : t -> name:string -> value:string -> unit
+    val put_path : t -> string list -> unit
```
talex5, line 14: `setenv` probably makes more sense for the name than `putenv` here (in C, `setenv` is the newer one).

Would be good to have `unsetenv` too.

lib_eio/vars.mli
```diff
@@ -0,0 +1,46 @@
+open Std
+
+type ty = [ `Vars ]
+type 'a t = ([> ty] as 'a) r
+
+(** {1 Accessing environment variables} *)
+
+val get_all : _ t -> (string * string) list
```
talex5, line 8: It's not very helpful that `get_all` returns a list of pairs, but `Process.spawn ~env` expects an array of strings.

lib_eio/vars.mli
patricoferris: Ok, maybe something like?

```ocaml
val get_paths : _ t -> string -> string list
(** [get_paths var name] gets the value associated with [name] and splits it using the backend-specific path separator. *)
```

lib_eio/vars.mli
patricoferris: The reasoning here was that I saw a few distinct use cases with `Process.spawn ~env`:

 1. The user wants to inherit the environment from the parent in which case they provide no value to `~env`, so there is no need for `get_all` to agree on the type.
 2. The user wants to provide a filtered set of the parent process's environment variables in which case they are likely going to want to have access to `(name, value)` pairs in order to do the filtering before having to rejoin them to be a `string array`. Of course, some filtering could be done on a version of `get_all` that returns a `string array`, for example, `String.starts_with ~prefix:"OCAML" v`.
 3. The user wants to extend the variables to include everything from the parent in addition to some executable-specific environment variables: this is the most compelling case for `get_all` to return a `string array`.

This does not take into account other use cases for `get_all` like finding some environment variables and their values for in-process OCaml logic where, I think, `(name, value)` pairs make the most sense.

I'm happy to make this return a `string array` (and we could provide a `split` function) but I thought I would lay out some of the reasoning for the current API.

Lemme know what you think :-)

talex5:
Now I'm wondering how useful it is to be able to mutate the environment globally like this. The uses I can think of are:

1. Changing the environment to be inherited by subprocesses. But then it would be safer to make a copy instead and pass it explicitly using `Process.spawn ~env`.
2. Passing values to other Eio code in the same process. But since you have to pass `env` around anyway, there's no advantage over passing the values directly. Or you could use fiber-local variables, which are typed and scoped sensibly.
3. Passing values to legacy/non-Eio code. But then the receiving code isn't using this system, so the mutex doesn't help, and you might as well use the `Unix` API directly.

It might be simpler if `Stdenv.vars` just returned the raw `string array`, and `Eio_unix` perhaps had some helpers to work with environments (e.g. convert to and from string maps, and convert values between strings and path lists).

lib_eio/vars.mli
avsm: That looks good to me. Probably also worth specifying what happens with repeated separators -- are empty segments preserved? (I'm not even sure if that's valid in PATHs)

lib_eio/vars.mli
avsm: If anything, this is case to change the Process.spawn API to take a list instead. It's a bit of a throwback to the C API to access an OCaml array here

patricoferris:
> It might be simpler if Stdenv.vars just returned the raw string array, and Eio_unix perhaps had some helpers to work with environments (e.g. convert to and from string maps, and convert values between strings and path lists).

I think this could be a fine approach, essentially making the environment as handled by Eio readonly (probably 90% of uses cases). It does not allow for @avsm's needs stated here https://github.com/ocaml-multicore/eio/pull/923#issuecomment-5354937134 (though as you say, you can always default to the racier `Unix.putenv` should you need).

However, I think it would be nice to be able to write portable Eio applications that can `Eio.Var.get_opt` at the very least (without linking `unix`). For example, in a jsoo application where we can provide the environment from the `jsooEnv`variable (see [the documentation](https://ocsigen.org/js_of_ocaml/latest/js_of_ocaml/options.html#env_vars)). Though, they have made that work with `unix` so maybe the point is a bit moot. I think this happens by default if `Stdenv.vars` returns a `(string * string) list` and users will naturally use `List.assoc_opt` or (for heavier use cases) transform it into a map (`StringMap.of_list`).

avsm:
I agree it's useful to not have a dependency on Eio_unix for 'pure' Eio applications. However, I'm also happy with a read-only environment for now, given how horrendously messy mutating it is...

lib_eio/vars.mli
talex5: I've opened #930, which changes the type `spawn` uses to `Process.Env.t`. Currently it's an alias of `string array`, but we can change it later once people have a chance to start using the new functions in there.

`get_all` can then use the same type. Though possibly it should just be `Process_mgr.current_environment` or something. The process manager already implicitly has access to the environment, since you can use `?env:None` so it probably makes sense to put it there.

patricoferris:
Thanks for adding those functions in #930 @talex5 ! 

Should the `vars` provided by the environment be a readonly copy of the vars when the program starts or should it be a function somewhat equivalent to:

```ocaml
let vars () = Unix.environment () |> Eio.Process.Env.of_array
```

so that it reflects `Unix.putenv`?

EDIT: I suspect it should be somewhat lazy. I'm running nix and my environment is pretty big (`env | wc -c` gives 14765). We probably shouldn't inflict that upon every Eio program?

talex5:
I think it should read the current environment each time you call it. Otherwise, there's no way to get that.

patricoferris:
I have updated this PR to only contain the readonly functions that were here originally: `get_all` and `get_opt`. Initially, I had only provided direct access to the `Eio.Process.Env.t`, that is, the capability was `method vars : unit -> Eio.Process.Env.t`. This seems wasteful for the most likely use case which is to pick out a few environment variables for your application (e.g. `XDG_DATA_DIRS`).

Here are some quick benchmarks comparing the various APIs. The `all` benchmark calls `get_all`every time and then finds a variable using `Eio.Process.Env.get_opt`, `all_once` calls `get_all` once and then `Eio.Process.Env.get_opt` every time and `direct` uses the `Eio.Vars.get_opt` directly every time: 

```
                  name,   major-allocated,   minor-allocated,   monotonic-clock
        get_env/all 10,        346.619865,      38577.227215,      29248.828738
        get_env/all 50,       1677.003302,     181788.755923,     138307.102538
       get_env/all 100,       3323.770915,     361218.802888,     274768.577259
   get_env/all_once 10,        121.955078,      14694.900546,      11652.096583
   get_env/all_once 50,        563.133065,      61951.683658,      48161.605588
  get_env/all_once 100,       1113.238600,     121431.915770,      93960.646400
     get_env/direct 10,          0.000000,          0.000000,        974.649587
     get_env/direct 50,          0.000000,          0.000000,       4369.486360
    get_env/direct 100,          0.000000,          0.000000,       8753.393503
```

talex5:
Having a separate function for the common case of reading one variable seems fine to me.
(though a bit surprising that `get_env/all_once` is 10x slower than C, given that it should be doing about the same thing)

How about moving these two functions to the process manager? It makes sense to provide a way to get an `Env.t` from there, since you typically need it in order to add things before calling `Process.spawn`, and the process manager implicitly has access to it anyway (with `?env:None`).

talex5:
> (though a bit surprising that get_env/all_once is 10x slower than C, given that it should be doing about the same thing)

OK, I had a look at https://github.com/lattera/glibc/blob/master/stdlib/getenv.c and it's doing some tricks! It has a special case for 1-character names, and if not then it hard-codes the test for the first two characters for the fast-path!

(also `String.starts_with` would probably do better to call `String.length prefix`, which is fast, on each iteration, rather than allocating a closure to hold it)

talex5:
> How about moving these two functions to the process manager?

To clarify, something like:
```ocaml
Eio.Process.environment : _ mgr -> Env.t
Eio.Process.getenv_opt : _ mgr -> string -> string option
```

patricoferris:
Happy to add that. Are we not concerned that applications that just want to check the environment are also capable of spawning subprocesses?

talex5:
That is the down-side of moving it.

But, assuming the environment doesn't change, you can get the `Process.Env.t` snapshot at startup and pass that around just as easily as passing the separate `Eio.Vars.t`. And if you're mutating the environment later then it's probably because you want to spawn a process with it.

So I suspect having a separate `Vars` module isn't worth it (do you have a use for it?).

patricoferris:
> But, assuming the environment doesn't change, you can get the Process.Env.t snapshot at startup and pass that around just as easily as passing the separate Eio.Vars.t. And if you're mutating the environment later then it's probably because you want to spawn a process with it.

I'm convinced :-) I've added those functions in the latest push.

> (though a bit surprising that get_env/all_once is 10x slower than C, given that it should be doing about the same thing)

Yes, I was a little surprised. Here is the benchmark when run inside a call to `env -` so all my Nix stuff isn't in there:

```
                  name,   major-allocated,   minor-allocated,   monotonic-clock
        get_env/all 10,          0.008949,        401.035481,        994.500899
        get_env/all 50,          0.038467,       1892.684723,       4655.264706
       get_env/all 100,          0.088574,       3866.065225,       9243.779433        
   get_env/all_once 10,          0.001615,        179.325236,        755.286173
   get_env/all_once 50,          0.026406,       1216.792224,       3414.363237
  get_env/all_once 100,          0.081945,       2738.526894,       6648.785586   
     get_env/direct 10,          0.000000,          0.000000,        415.258044
     get_env/direct 50,          0.000000,          0.000000,       1864.587791
    get_env/direct 100,          0.000000,        381.505302,       3698.969410     
```

lib_eio/unix/process.ml
```diff
@@ -136,6 +136,10 @@ end) = struct
 
   type tag = [ `Generic | `Unix ]
 
+  let environment _t = Unix.environment () |> Eio.Process.Env.of_array
+  let getenv_opt _t name =
+    try Some (Unix.getenv name) with Not_found -> None
```
talex5:
```suggestion
  let getenv_opt _t name = Sys.getenv_opt name
```

lib_eio/process.mli
```diff
@@ -109,6 +109,15 @@ module Env : sig
   val pp : t Fmt.t
 end
 
+(** {2 Accessing environment variables} *)
```
talex5: This heading should go higher up, to include the `Env` module (and maybe just be called "Environment variables").

lib_eio/process.mli
```diff
@@ -109,6 +109,15 @@ module Env : sig
...
+
+val environment : _ mgr -> Env.t
+(** [environment t] gets the full list of environment variables. *)
```
talex5: Maybe emphasise that this is taking a snapshot?
```suggestion
(** [environment t] returns a snapshot of this process's current environment. *)
```

talex5 (approved):
Looks good - thanks!

(some minor suggestions in the comments)

avsm:
This looks good to me. Just one nit is that Env.get_opt called validate (and so raised Invalid_argument), so we end up with getenv_opt mgr "" returning None while Env.get_opt "" raises (I think, not code tested this)

I can add Windows support for this into #936 pretty easily I think once this goes in

talex5:
> getenv_opt "" returning None while Env.get_opt "" raises

OK, this seems like a good argument to make our `getenv_opt` return `None` (like it did before), to match OCaml's behaviour.

talex5:
I've pushed the suggested changes, and also modified `Env.get_opt` to return `None` for invalid keys to match the behaviour of `Sys.getenv_opt` (which seems fine to me). I think this is ready to merge.

avsm:
Also re-read it, and looks great to me. Thanks @patricoferris !

patricoferris:
Thanks both :-)

PR #926, opened 2026-08-21
Add index.mld for eio

talex5:
The default auto-generated index is hard to navigate.

Before:
before

After:
after

### DIFF
--- a/lib_eio/index.mld
+++ b/lib_eio/index.mld
@@ -1,0 +1,40 @@
+{0 Eio}
+
+Eio provides support for concurrency (juggling many tasks) and
+parallelism (using multiple CPU cores for performance).
+
+See the {{:https://github.com/ocaml-multicore/eio/blob/main/README.md} Tutorial} for an introduction.
+
+The main {!Eio} module provides:
+
+- {{!Eio.Fiber} Fiber}, {{!Eio.Switch} Switch}, {{!Eio.Cancel} Cancel} (managing fibers)
+- {{!Eio.Stdenv} Stdenv} (getting access to host resources)
+- {{!Eio.Path} Path}, {{!Eio.File} File} (managing files and directories)
+- {{!Eio.Flow} Flow}, {{!Eio.Buf_read} Buf_read}, {{!Eio.Buf_write} Buf_write} (text streams)
+- {{!Eio.Net} Net} (networking)
+- {{!Eio.Process} Process} (running other programs)
+- {{!Eio.Time} Time} (clocks and timeouts)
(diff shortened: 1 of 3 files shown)


lib_eio/index.mld
```diff
@@ -0,0 +1,40 @@
...
+- {{!Eio.Path} Path}, {{!Eio.File} File} (managing files and directories)
+- {{!Eio.Flow} Flow}, {{!Eio.Buf_read} Buf_read}, {{!Eio.Buf_write} Buf_write} (text streams)
+- {{!Eio.Net} Net} (networking)
+- {{!Eio.Process} Process} (running other programs)
+- {{!Eio.Time} Time} (clocks and timeouts)
+- {{!Eio.Promise} Promise}, {{!Eio.Semaphore} Semaphore}, {{!Eio.Mutex} Mutex}, {{!Eio.Condition} Condition}, {{!Eio.Lazy} Lazy} (concurrency primitives)
+- {{!Eio.Stream} Stream}, {{!Eio.Pool} Pool} (collections)
+- {{!Eio.Domain_manager} Domain_manager}, {{!Eio.Executor_pool} Executor_pool} (using multiple domains)
+- {{!Eio.Std} Std} ([open Eio.Std] to get access to some common modules)
+- Less commonly-used: {{!Eio.Fs} Fs}, {{!Eio.Exn} Exn}, {{!Eio.Debug} Debug}, {{!Eio.Resource} Resource}
```
avsm: Should probably group Eio.Fs with Path (since it also has a bunch of error types users will see) and Eio.Exn is commonly seen if not used.

lib_eio/index.mld
```diff
@@ -0,0 +1,40 @@
...
+
+{2 Extended Unix APIs}
+
+{!Eio_unix} provides low-level access to file-descriptors, system threads, etc.
```
avsm:
```suggestion
{!Eio_unix} provides low-level but portable access to file-descriptors, threads and other system resources.
```

lib_eio/index.mld
```diff
@@ -0,0 +1,40 @@
...
+
+{2 Mocks}
+
+{!Eio_mock} is dummy backend for testing.
```
avsm:
```suggestion
{!Eio_mock} is a dummy backend for testing.
```

avsm (approved):

lib_eio/index.mld
talex5: OK, I've moved `Exn` into the main list.

I moved `Fs` away from there because you almost never want to view this file and it's distracting. i.e. if you want to do something with files and are searching for a suitable-looking module, this probably isn't the one you want. `Eio.Path` does link to it in its docs, though (including for errors).

PR #927, opened 2026-08-22
Add Eio.Fiber.List.partition_map

kentookura:
I could not implement this myself since `Limiter` is hidden from the interface.

### DIFF
--- a/lib_eio/core/fiber.ml
+++ b/lib_eio/core/fiber.ml
@@ -260,6 +260,7 @@ module List = struct
       aux items
 
   let map ?max_fibers fn = filter_map ?max_fibers (fun x -> Some (fn x))
+  let partition_map ?max_fibers fn items = List.partition_map Fun.id (map ?max_fibers fn items) 
   let filter ?max_fibers fn = filter_map ?max_fibers (fun x -> if fn x then Some x else None)
 
   let iter ?(max_fibers=max_int) fn items =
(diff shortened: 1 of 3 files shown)


lib_eio/core/fiber.ml
```diff
@@ -259,6 +259,29 @@ module List = struct
...
+      let rec aux left right = function
+        | [] -> (List.rev left, List.rev right)
+        | [x] ->
+          (match Limiter.use limiter fn x with
+           | Either.Left v -> (List.rev (v :: left), List.rev right)
+           | Either.Right v -> (List.rev left, List.rev (v :: right)))
+        | x :: xs ->
+          let x = Limiter.fork_promise_exn limiter fn x in
+          let left, right =
+            match Promise.await x with
```
avsm: This seems to a fork a promise and then immediately wait on it. Are you sure it's actually running concurrently?   It might just be simpler to have a one-liner `let partition_map ?max_fibers fn items = List.partition_map Fun.id (map ?max_fibers fn items)` (which allocates a bit more, but the dominant cost here would be fiber creation I expect).

talex5 (changes requested):
> Are you sure it's actually running concurrently?

Looks wrong to me too. And the test output also shows that it's not running concurrently.

> It might just be simpler to have a one-liner

I agree: it makes sense to start with the one-liner, and optimise it later if it's a problem. @avsm's suggested one-liner also fixes the test output.

talex5:
(I'm assuming this PR was closed because you decided it's easier to put the one-liner in your own code. Updating this PR to use it would also be fine.)

kentookura:
Will do so once I get the chance. Thanks!

PR #928, opened 2026-08-24
unix: reject out-of-range chown ids but allow -1 for unchanged

avsm:
posix allows -1 for 'unchanged'


talex5:
Won't the kernel check this for us anyway? Also, `uid_t` is unsigned on Linux!

avsm:
This is sufficiently weird that I thought a check here (in the portable layer) would be prudent The sign of uid_t/gid_t can be either signed or unsigned, so some casts here are probably necessary as well. e.g. https://github.com/python/cpython/issues/46289 discusses this as well.

PR #929, opened 2026-09-01
eio_windows: fix anonymous pipes on Windows

avsm:
This is a testcase failing on my machine atm, so opening to check windows CI here as well (I'm struggling with Windows Redirection Guard nuking my ssh setup due to the use of symlinks in Flexlink!)

### DIFF
--- a/lib_eio_windows/test/test_pipe.ml
+++ b/lib_eio_windows/test/test_pipe.ml
@@ -1,0 +1,40 @@
+(* Tests for anonymous pipes *)
+
+open Eio.Std
+
+let read_all flow =
+  let b = Buffer.create 16 in
+  Eio.Flow.copy flow (Eio.Flow.buffer_sink b);
+  Buffer.contents b
+
+let test_transfer () =
+  Switch.run @@ fun sw ->
+  let r, w = Eio_unix.pipe sw in
+  Eio.Flow.copy_string "hello" w;
+  Eio.Flow.close w;
+  Alcotest.(check string) "transfer" "hello" (read_all r)
+
+let test_read_before_write () =
+  Switch.run @@ fun sw ->
+  let r, w = Eio_unix.pipe sw in
+  Fiber.both
+    (fun () ->
+       let buf = Cstruct.create 8 in
+       let n = Eio.Flow.single_read r buf in
+       Alcotest.(check string) "data" "ping" (Cstruct.to_string ~len:n buf))
+    (fun () -> Eio.Flow.copy_string "ping" w)
+
+let test_eof () =
+  Switch.run @@ fun sw ->
+  let r, w = Eio_unix.pipe sw in
+  Eio.Flow.close w;
+  let buf = Cstruct.create 1 in
+  match Eio.Flow.single_read r buf with
+  | _ -> Alcotest.fail "read should have signaled eof"
+  | exception End_of_file -> ()
+
+let tests = [
+  "transfer", `Quick, test_transfer;
+  "read-before-write", `Quick, test_read_before_write;
+  "eof", `Quick, test_eof;
+]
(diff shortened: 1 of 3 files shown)


avsm:
Confirming it fails in CI too 
```
 [FAIL]        pipe             0   transfer.
  
    ... with 6 subsequent tests skipped.
  
  ┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ [FAIL]        pipe             0   transfer.                                                                         │
  └──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
  [exception] Eio.Io _
              Raised at Stdlib__Effect.Deep.discontinue.(fun) in file "effect.ml", line 63, characters 41-48
              Called from Eio_unix__Private.pipe in file "lib_eio/unix/private.ml", line 14, characters 13-37
              Called from Dune__exe__Test_pipe.test_transfer.(fun) in file "lib_eio_windows/test/test_pipe.ml", line 12, characters 13-29
              Called from Eio__core__Switch.run_internal in file "lib_eio/core/switch.ml", line 133, characters 8-12
              Re-raised at Eio__core__Switch.maybe_raise_exs in file "lib_eio/core/switch.ml", line 119, characters 21-56
              Called from Eio__core__Switch.run_internal in file "lib_eio/core/switch.ml", line 149, characters 4-21
              Called from Eio__core__Cancel.with_cc in file "lib_eio/core/cancel.ml", line 118, characters 8-12
              Re-raised at Eio__core__Cancel.with_cc in file "lib_eio/core/cancel.ml", line 120, characters 50-58
              Called from Alcotest_engine__Core.Make.protect_test.(fun) in file "src/alcotest-engine/core.ml", line 186, characters 17-23
              Called from Alcotest_engine__Monad.Identity.catch in file "src/alcotest-engine/monad.ml", line 24, characters 31-35
```

avsm:
I'm going to layer the process spawnign PR over this one; I can't do a full test case for a failing pipe without another process!

lib_eio_windows/test/test_pipe.ml
```diff
@@ -0,0 +1,40 @@
+(* Tests for anonymous pipes *)
+
+open Eio.Std
+
+let read_all flow =
```
talex5, line 5: We do have `Eio.Flow.read_all`.

talex5 (commented):
This doesn't look right - how will cancellation work?

According to https://stackoverflow.com/questions/34504970/non-blocking-read-on-os-pipe-on-windows, Windows does support non-blocking reads on pipes:

> [SetNamedPipeHandleState](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365787(v=vs.85).aspx) doc allows to put anonymous pipe to non-blocking mode.

Also, we probably want the same check for blocking mode that's in eio_posix. There, we have e.g.

https://github.com/ocaml-multicore/eio/blob/318abba840288a564fc57035d920a29b13774953/lib_eio_posix/low_level.ml#L67-L70

But in eio_windows, the wait is unconditional:

https://github.com/ocaml-multicore/eio/blob/318abba840288a564fc57035d920a29b13774953/lib_eio_windows/low_level.ml#L41-L44

lib_eio_windows/test/test_pipe.ml
avsm: Need this expanded to put debug printfs in! Not actually tried cancellations in, will do next: I think I've found one potential weirdness which is that the pipe behaviour changes if I SSH in _vs_ run at the machine directly! Windows is driving me mad, mad I tell you.

avsm:
I just can't see how to do cancellation without shifting to IOCP. Everything goes wrong otherwise; the latest thing is that Unix.write converts a 'full pipe' to EPIPE, and also writing an overly large buffer deadlocks. Rust uses overlapping pipes or a socketpair instead with IOCP, I'm going to investigate that.

PR #930, opened 2026-09-03
Add Process.Env

talex5:
This provides helper functions to get and set environment variables. For backwards compatibility the type is exposed, but having this module will make it easier to make the type abstract later.

(@avsm mentioned in https://github.com/ocaml-multicore/eio/pull/923#discussion_r3913887552 that we might want to change the type here at some point)

@patricoferris: this might be useful for #923.

### DIFF
--- a/lib_eio/process.ml
+++ b/lib_eio/process.ml
@@ -47,6 +47,100 @@ type 'tag mgr_ty = [ `Process_mgr | `Platform of 'tag ]
 type 'a mgr = 'a r
  constraint 'a = [> [> `Generic] mgr_ty]
 
+module Env = struct
+  let on_windows = (Sys.os_type = "Win32")
+
+  module Name = struct
+    type t = string
+
+    let normalise =
+      if on_windows then String.uppercase_ascii
+      else Fun.id
+
+    let compare x y =
+      String.compare (normalise x) (normalise y)
+
+    let starts_with ~prefix =
+      let prefix = normalise prefix in
+      fun x -> String.starts_with ~prefix (normalise x)
+
+    let validate t =
+      let bad_char = function
+        | '\000' | '=' -> true
+        | _ -> false
+      in
+      if t = "" || String.exists bad_char t then
+        Fmt.invalid_arg "Invalid environment variable name %S" t
+  end
+
+  module M = Map.Make(Name)
+
+  type t = string array
+
+  let of_array = Fun.id
+  let to_array = Fun.id
+  let empty = [| |]
+
+  let validate_value value =
+    if String.contains value '\000' then
+      Fmt.invalid_arg "Invalid environment variable value %S" value
+
+  let validate_binding (name, value) =
+    Name.validate name;
+    Option.iter validate_value value
+
+  let entry name value =
+    Printf.sprintf "%s=%s" name value
+
+  let get_opt name t =
+    Name.validate name;
+    let prefix = name ^ "=" in
+    Array.find_opt (Name.starts_with ~prefix) t
+    |> Option.map (fun e ->
+        let i = String.length prefix in
+        String.sub e i (String.length e - i)
+      )
+
+  let override bindings t =
+    List.iter validate_binding bindings;
+    let all_bindings = M.of_list bindings in
+    let bindings = ref all_bindings in
+    let updated =
+      Array.to_list t
+      |> List.filter_map (fun e ->
+          match String.index e '=' with
+          | exception Not_found -> Some e       (* Not a normal k=v entry *)
+          | i ->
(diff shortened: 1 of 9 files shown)


lib_eio/process.ml
```diff
@@ -47,6 +47,76 @@ type 'tag mgr_ty = [ `Process_mgr | `Platform of 'tag ]
...
+    let prefix = name ^ "=" in
+    Array.find_opt (String.starts_with ~prefix) t
+    |> Option.map (fun e ->
+        let i = String.length prefix in
+        String.sub e i (String.length e - i)
+      )
+
+  module M = Map.Make(String)
+
+  let override bindings t =
```
avsm, line 104: This is a deviation from execve behaviour, which passes through duplicate bindings and lets glibc/musl/etc handle that. Not necessarily a bad thing though, I haven't investigated what the libcs do yet.

lib_eio/process.ml
```diff
@@ -47,6 +47,76 @@ type 'tag mgr_ty = [ `Process_mgr | `Platform of 'tag ]
...
+    if String.contains value '\000' then
+      Fmt.invalid_arg "Invalid environment variable value %S" value
+
+  let entry name value =
+    validate_name name;
+    validate_value value;
+    Printf.sprintf "%s=%s" name value
+
+  let get_opt name t =
+    let prefix = name ^ "=" in
```
avsm, line 97: Does this deliberately not call `validate_name` on the `name`?

lib_eio/process.ml
talex5: I'll add some docs, but the idea is that the raw `of_array` API lets you do whatever you want (arbitrary array of strings), while the bindings APIs (that take lists of pairs) work like `setenv` (every entry added must be a key/value pair and duplicates are not created).

lib_eio/process.ml
talex5: Saying an invalid name wasn't found didn't seem too unreasonable. However, I've added a commit making it all a bit stricter now.

lib_eio/process.ml
avsm: This makes sense. What about case sensitivity? I think windows end keys are insensitive

lib_eio/process.mli
patricoferris, line 157: Do you think it is worth telling users to use `Process.Env.{of_bindings,of_array}` here (or up where the note is on the type of `Env.t`) so that eventually if that type is made abstract their code will not break? Existing users should also update their code.

lib_eio/process.ml
talex5: Good point. I guess the easiest way would be to change the behaviour of `Process.Env` depending on the host OS. I wonder what encoding Windows assumes for a case-insensitive compare? Are names always ASCII there?

talex5: I've pushed a commit that does case insensitive compares on Windows now. It's not very efficient, but there aren't usually many variables anyway.

lib_eio/process.mli
talex5: Could do, but I'm hoping that once #923 is in users won't normally need to create an `Env.t` themselves anyway.

I don't think OCaml has a way to generate a compiler warning when you rely on a type alias, so there's not much we can do about existing users (which is fine; the important thing is to have at least one version of Eio where the old and new APIs both work, so no one needs to wait for anyone else before updating their code).

lib_eio/process.ml
avsm: The change looks good, but I really dislike the portable interface varying behaviour based on which host it's running on. Here's a radical idea: why not specify our Eio environment interface as explicitly being case insensitive? We are already constraining it to forbid duplicates, and it seems like normalising on case should also be very safe.  We could also preserve the case at the Eio level (so it's passed through as-is) but is case-insensitive for comparisons.

talex5: Being case insensitive in general seems bad: it's a surprising change to the expected behaviour, and it causes trouble with non-ascii encodings.

The other option is to track whether a particular environment is Windows-style or not. But we can't use a flag at the moment because the type needs to be `string array`. It would be possible to track the type with a fake entry at the start (`EIO_OS=Windows` or something, that gets stripped out in `to_array`). Ugly, though.

A simpler solution is to recommend that environment variable names are upper-case (which they mostly are anyway). As long as all variables are uppercase, the Windows and POSIX behaviours are the same anyway.

avsm: > A simpler solution is to recommend that environment variable names are upper-case (which they mostly are anyway). As long as all variables are uppercase, the Windows and POSIX behaviours are the same anyway.

agreed! good idea

Issue #931, opened 2026-09-03
Windows: Paths are handed to NtCreateFile unqualified, raising Unix.ENOENT


kentookura:
NtCreateFile expects the paths to be fully qualified: https://learn.microsoft.com/en-us/windows/win32/api/winternl/nf-winternl-ntcreatefile

```ocaml
let nt_prefix = "\\??\\"

let try_open ~label env path =
  let eio_path = Eio.Path.(Eio.Stdenv.fs env / path) in
  match Eio.Path.with_open_in eio_path (fun _ -> ()) with
  | () -> Printf.printf "%s: OK   (path = %S)\n%!" label path
  | exception exn ->
    Printf.printf "%s: FAILED (path = %S): %s\n%!" label path
      (Printexc.to_string exn)

let () =
  Eio_main.run @@ fun env ->
  let path = Filename.temp_file "eio" "abs" in
  Fun.protect ~finally:(fun () -> Sys.remove path) @@ fun () ->
  Printf.printf "Sys.file_exists %S = %b\n%!" path (Sys.file_exists path);
  try_open ~label:"unqualified" env path;
  try_open ~label:"NT-qualified" env (nt_prefix ^ path)
```
Output:
```
Sys.file_exists "C:\\Users\\User\\AppData\\Local\\Temp\\eioddc8e3abs" = true
unqualified: FAILED (path = "C:\\Users\\User\\AppData\\Local\\Temp\\eioddc8e3abs"): Unix.Unix_error(Unix.ENOENT, "openat", "")
NT-qualified: OK   (path = "\\??\\C:\\Users\\User\\AppData\\Local\\Temp\\eioddc8e3abs")
```

talex5:
Thanks for testing this!

It would probably be good to add some tests for absolute paths to https://github.com/ocaml-multicore/eio/blob/main/lib_eio_windows/test/test_fs.ml. It checks eio_windows can manipulate absolute paths, but looks like it never actually tries to open one!

Also, this should probably be changed to include the path in the error:

https://github.com/ocaml-multicore/eio/blob/ab4dd74e0799230e756a883d23f0527f6be19897/lib_eio_windows/eio_windows_stubs.c#L164

@avsm: what's the best way to handle this? Should the `Eio.Path.t` itself be full-qualified, or should e.g. `caml_eio_windows_openat` be adding it?

avsm:
I'm working on this; there are some complexities around symlink following that I'm just figuring out.

[...]

PR #932, opened 2026-09-07
Fix NT path handling and stat in eio_windows

avsm:
While fixing #931 and adding tests, I went into a rabbithole of fixes involving Windows symlinks. For anyone who want to follow (pun intended) in my footsteps, beware of [Windows Redirect Guard](https://www.microsoft.com/en-us/msrc/blog/2025/06/redirectionguard-mitigating-unsafe-junction-traversal-in-windows) which is now active on OpenSSH by default and completely destroys the symlink-using Flexlink OCaml toolchain.

This fixes the issue reported in #931 and uses Nt_path consistently (exposing some new functions there too). I found some other bugs in symlink handling too:

- The `openat`/`unlinkat` stubs checked an uninitialized handle before the`NtCreateFile` status, so failures were misreported.
- `nofollow` replaced the create options with a Win32 flag, dropping the directory/non-directory constraints so stat kinds didnt' work
- `open_out` `ELOOP` recovery ran `readlink` relative to the process directory rather than the sandbox root.
- `openat` was only partially applied inside `Err.run`, so `Unix` errors from `open_in` escaped unmapped.
- Opening for reading or `stat` could create a missing file.
- `stat` always passed `FILE_NON_DIRECTORY_FILE`, so statting a directory failed.
- `stat` with `~follow:false` followed the link anyway; there is now a third open mode that opens the symlink, and `fstat` reports the target length as the size like the Posix backend does.

Best reviewed commit by commit and the testcases are separate so I could verify they failed before and worked after.

### DIFF
--- a/lib_eio/utils/nt_path.ml
+++ b/lib_eio/utils/nt_path.ml
@@ -53,20 +53,28 @@ let volume_prefix =
   <|> (sep *> sep *> component *> next)                  (* \\server\share, \\?\C: or \\.\device *)
   <|> (bslash *> qmark *> qmark *> component *> next)    (* \??\C: - the NT object-manager form (backslash only) *)
 
-let volume_end s = Option.value (volume_prefix s 0) ~default:0
-let volume s = String.sub s 0 (volume_end s)
-
 (* Win32 does no normalization in the verbatim and NT namespaces. *)
 let verbatim_prefix = bslash *> (bslash <|> qmark) *> qmark
 let verbatim s = Option.is_some (verbatim_prefix s 0)
 
+(* [\??\], [\\?\] and [\\.\] all name the NT object-manager namespace. *)
+let nt_prefix = (verbatim_prefix <|> (bslash *> bslash *> chr '.')) *> bslash
+
 (* [is_relative p] is [true] unless [p] begins with a volume or a separator. *)
 let is_relative s = Option.is_none ((volume_prefix <|> sep) s 0)
 
+let volume_end s = Option.value (volume_prefix s 0) ~default:0
+let drop n s = String.sub s n (String.length s - n)
+
+(* A path's volume prefix, and the rest of the path. *)
+let split_volume p =
+  let n = volume_end p in
+  String.sub p 0 n, drop n p
+
 let split p =
   let vend = volume_end p in
-  let sep_at = if verbatim p then Char.equal '\\' else is_sep in
-  let sep_at i = sep_at p.[i] in
+  let sep_char = if verbatim p then Char.equal '\\' else is_sep in
+  let sep_at i = sep_char p.[i] in
   (* Trailing separators are ignored; one is kept for a bare root. *)
   let rec trim i = if i > vend + 1 && sep_at (i - 1) then trim (i - 1) else i in
   let stop = trim (String.length p) in
@@ -74,7 +82,7 @@ let split p =
   else
     let rec rsep i = if i < vend then None else if sep_at i then Some i else rsep (i - 1) in
     match rsep (stop - 1) with
-    | None -> Some (volume p, String.sub p vend (stop - vend))
+    | None -> Some (String.sub p 0 vend, String.sub p vend (stop - vend))
     | Some idx ->
       let basename = String.sub p (idx + 1) (stop - idx - 1) in
       let dirname =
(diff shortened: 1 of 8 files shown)


lib_eio_windows/test/test_fs.ml
```diff
@@ -204,9 +244,7 @@ let test_symlink env () =
      Unix.mkdir "another" 0o700;
      print_endline @@ Unix.realpath "to-subdir" |}
   *)
-  if not (Unix.has_symlink ()) then
-    Printf.printf "Skipping test_symlink on systems that don't support symlinks.\n"
-  else
+  with_symlinks "test_symlink" @@ fun () ->
```
avsm: I'm just running the symlink tests unconditionally as it seems unlikely that we'll have developer mode unavailable, but I can change this back too.

avsm: (As in not erroring on them in non-developer mode)

talex5: That test got added in https://github.com/ocaml-multicore/eio/pull/771 so I guess it was useful for someone. If it comes back, it should probably use `Alcotest.skip ()`, so it shows up as skipped rather than as passed.

lib_eio/utils/nt_path.ml
```diff
@@ -97,3 +114,38 @@ let join p1 p2 =
...
+    let backslashes = String.map (fun c -> if c = '/' then '\\' else c) in
+    let vol, rest = split_volume (backslashes p) in
+    let cwd_vol, cwd_rest = split_volume (backslashes cwd) in
+    let rooted = rest <> "" && rest.[0] = '\\' in
+    let vol, base =
+      match vol with
+      | "" -> cwd_vol, (if rooted then "" else cwd_rest)
+      | v when rooted || v.[0] = '\\' -> v, ""
+      | v when String.uppercase_ascii v = String.uppercase_ascii cwd_vol -> cwd_vol, cwd_rest
+      | v -> v, ""    (* Win32 keeps a current directory per drive; but we sadly cant see it *)
```
talex5:
```suggestion
      | v -> v, ""    (* Win32 keeps a current directory per drive; but we sadly can't see it *)
```

lib_eio_windows/eio_windows_stubs.c
```diff
@@ -143,41 +147,63 @@ CAMLprim value caml_eio_windows_openat(value v_dirfd, value v_nofollow, value v_
...
+
+  if (!DeviceIoControl(h, FSCTL_GET_REPARSE_POINT, NULL, 0, &buffer, sizeof(buffer), &len, NULL)) {
+    caml_win32_maperr(GetLastError());
+    uerror("fstat", Nothing);
+  }
+
+  if (buffer.point.ReparseTag != IO_REPARSE_TAG_SYMLINK)
+    CAMLreturn(Val_none);
+
+  CAMLreturn(caml_alloc_some(Val_int(buffer.point.SymbolicLinkReparseBuffer.SubstituteNameLength / 2)));
```
talex5: Why `/ 2`? Is this is to convert from number of bytes to number of characters? But `Stat.size` says `Total size in bytes`.

lib_eio_windows/low_level.mli
```diff
@@ -39,15 +39,23 @@ val lseek : fd -> Optint.Int63.t -> [`Set | `Cur | `End] -> Optint.Int63.t
 val fsync : fd -> unit
 val ftruncate : fd -> Optint.Int63.t -> unit
 
+type follow = Follow | Nofollow | Open_link
+(** What to do with a path that is a symlink: open its target, fail with
+    [ELOOP], or open the link itself. *)
+
 val fstat : fd -> Unix.LargeFile.stats
+(** Unlike [Unix.fstat], a symlink opened with [Open_link] is reported as one,
+    with the length of its target as its size. *)
```
talex5: "length of its target" is a bit vague (it could mean the length of the file, not the path).

(and https://ocaml.org/manual/5.5/api/Unix.LargeFile.html#TYPEstats says the size is in bytes)

talex5 (approved):
Thanks - this looks like a huge improvement!

lib_eio_windows/eio_windows_stubs.c
avsm: Yep its because Windows gives us WCHAR 2-bytes so I had to half it.

lib_eio_windows/low_level.mli
avsm: I dont think there's really a clear definition of what 'the portable size of a symlink' is supposed to be. Leaving it as the raw byte size seems low level, but at least being conservative and larger is a reasonable default. I'll change it to bytes and simplify the code.

lib_eio_windows/test/test_fs.ml
avsm: Ah I didnt know about skip () - I'll change it to use that thanks

avsm: Fixed in 0410489befb7681d65040580f77a2869f0b298c0

lib_eio_windows/eio_windows_stubs.c
avsm: fixed in 1b098d1b74f8ccdbda83df381d9652684a3bab40 to be x2

lib_eio_windows/low_level.mli
avsm: I actually just elided the comments instead

avsm:
Weird unrelated (?) error on fedora:

```
+++ b/_build/default/tests/.mdx/process.md.corrected
@@ -190,7 +190,7 @@ A script whose interpreter is missing reports the executable as not found:
let f = Eio.Stdenv.cwd env / "badinterp" in
Eio.Path.save f "#!/nonexistent/interpreter\n" ~create:(`Exclusive 0o700);
Fun.protect ~finally:(fun () -> Eio.Path.unlink f) (fun () -> Process.run mgr ["./badinterp"]);;
-Exception: Eio.Io Process Executable_not_found "./badinterp"
+Exception: Unix.Unix_error(Unix.EUNKNOWNERR 26, "execve", "")
```

```

avsm:
That was a transient error, gone on re-run. This is good to go I think.

talex5:
`Unix.EUNKNOWNERR 26` is `ETXTBSY`. I'll investigate...

PR #933, opened 2026-09-08
Fix (harmless) gcc warning calloc-transposed-args

talex5:
fork_action.c:101:28: warning: ‘calloc’ sizes specified with ‘sizeof’ in the earlier argument and not in the later argument [-Wcalloc-transposed-args]
      101 |   char **c = calloc(sizeof(char *), n + 1);
          |                            ^~~~
    fork_action.c:101:28: note: earlier argument should specify number of elements, later size of each element

See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=112364.


avsm:
Hah, I wrote something like this for OpenBSD [back in 2003](https://man.openbsd.org/gcc-local.1#ATTRIBUTES)! Fix is obviously correct and we can't use caml_ functions in there.

PR #934, opened 2026-09-08
Windows: add rename support

avsm:
This is somewhat more complex than expected since Windows behaves differently for overwriting files or directories, and the more modern MoveFileEx is path based and so unsuitable for our 'at' based handle approach.

However, the lower level implementation here using the 'traditional' Win32 APIs works fine on the NTFS and FAT32 volumes I tried it on. Note that FAT just doesnt support renaming onto directories so will fail.

This let's forester cram tests get a little further! #931

### DIFF
--- a/lib_eio_windows/eio_windows_stubs.c
+++ b/lib_eio_windows/eio_windows_stubs.c
@@ -267,9 +267,78 @@ CAMLprim value caml_eio_windows_unlinkat(value v_dirfd, value v_pathname, value 
   CAMLreturn(Val_unit);
 }
 
+/* Missing from mingw's winternl.h but fixed by Windows kernel ABI */
+#define FileRenameInformationEx ((FILE_INFORMATION_CLASS)65)
+
 CAMLprim value caml_eio_windows_renameat(value v_old_fd, value v_old_path, value v_new_fd, value v_new_path)
 {
-  uerror("renameat is not supported on windows yet", Nothing);
+  CAMLparam4(v_old_fd, v_old_path, v_new_fd, v_new_path);
+  HANDLE h, old_dir, new_dir;
+  OBJECT_ATTRIBUTES obj_attr;
+  IO_STATUS_BLOCK io_status;
+  UNICODE_STRING relative;
+  wchar_t *old_path, *new_path;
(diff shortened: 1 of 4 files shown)


lib_eio_windows/eio_windows_stubs.c
```diff
@@ -267,9 +267,78 @@ CAMLprim value caml_eio_windows_unlinkat(value v_dirfd, value v_pathname, value
...
+  wchar_t *old_path, *new_path;
+  FILE_RENAME_INFO *info;
+  size_t name_len, info_size;
+  NTSTATUS r;
+
+  pNtCreateFile NtCreatefile = (pNtCreateFile)GetProcAddress(GetModuleHandle("ntdll.dll"), "NtCreateFile");
+  caml_unix_check_path(v_old_path, "renameat");
+  caml_unix_check_path(v_new_path, "renameat");
+  old_dir = Is_some(v_old_fd) ? Handle_val(Field(v_old_fd, 0)) : NULL;
+  new_dir = Is_some(v_new_fd) ? Handle_val(Field(v_new_fd, 0)) : NULL;
```
talex5:
```suggestion
  old_dir = Is_some(v_old_fd) ? Handle_val(Some_val(v_old_fd)) : NULL;
  new_dir = Is_some(v_new_fd) ? Handle_val(Some_val(v_new_fd)) : NULL;
```

talex5: Could use `Some_val` here to make it slightly clearer:
```suggestion
  old_dir = Is_some(v_old_fd) ? Handle_val(Some_val(v_old_fd)) : NULL;
  new_dir = Is_some(v_new_fd) ? Handle_val(Some_val(v_new_fd)) : NULL;
```

talex5 (approved):
I don't know much about Windows renaming, but this looks reasonable to me.

avsm:
I'm still digging through MSDN to see if there's something simpler for more atomic renames, but I can't find anything...

avsm:
I'm not having much success with MSDN, so given how localised this change it I've made a note to improve it later on, but the feature's useful now!

PR #936, opened 2026-09-09
Windows: process spawning support

avsm:
The long awaited Windows process spawn after 2 unsuccessful other attempts that I didn't open a PR for! This is layered over #929 as we need pipes to work for tests. 

This doesn't use Unix.create_process since I found it tricky to integrate:
- The stdlib one inherits every fd in the process, so (e.g.) pipes leak. This PR follows the Eio style and restricts inheritance to the three standard handles + an explicit inheritance list.
- This one allows the cwd to be set via the Eio sandbox
- We can cancel without blocking the domain here since we have a systhread for it.

Note that any signal sent will terminate the process. Should we actually parse the signal numbers and do something 'terminal' only on SIGKILL/TERM?

Another difference from Posix is that CreateProcess resolves the executable path itself if its blank, but I could replicate the PATH search that we do in Posix (Windows seems to prefer the search to be done by the call, I'm not sure). 

We may need to restrict to OCaml 5.4 or higher since that's when @dra27 added caml_stat_char_array_to_utf16

Caveat: I am very inexperienced in these APIs so this is the results of much rooting around in MSDN, but I am happy to stand corrected on any of these decisions. This does progress the Forester test suite due to it spawning `xcopy` quite significantly though! :-)

### DIFF
--- a/lib_eio_windows/process.ml
+++ b/lib_eio_windows/process.ml
@@ -1,0 +1,142 @@
+open Eio.Std
+
+module Fd = Eio_unix.Fd
+
+external eio_spawn :
+  string option -> string array ->
+  Unix.file_descr -> Unix.file_descr -> Unix.file_descr ->
+  string -> int * Unix.file_descr
+  = "caml_eio_windows_spawn_bytes" "caml_eio_windows_spawn"
+
+external eio_process_wait : Unix.file_descr -> int = "caml_eio_windows_process_wait"
+external eio_process_terminate : Unix.file_descr -> int -> bool = "caml_eio_windows_process_terminate"
+
+let check_env env =
+  Array.iter
+    (fun s ->
+       if s = "" || String.contains s '\000' then
+         Fmt.invalid_arg "spawn: invalid environment entry %S" s)
+    env;
+  env
+
+(* Follow OCaml stdlib and only quote only where needed, as apparently cmd.exe
+   treats quoted arguments differently even if they dont have spaces. *)
+let command_line args =
+  let quote_arg arg =
+    if arg = "" || String.exists (function ' ' | '\t' | '\n' | '\011' | '"' -> true | _ -> false) arg
+    then Filename.quote arg
+    else arg in
+  String.concat " " (List.map quote_arg args)
+
+let terminated_exit_code = Config.status_control_c_exit
+
+module Process = struct
+  type t = {
+    pid : int;
+    handle : Fd.t;
+    mutable signalled : int option;
+    exited : (int, exn) result Promise.t;
+    resolve : (int, exn) result Promise.u;
+    mutable waiting : bool;
+    mutable hook : Switch.hook;         (* Removed once the process has been reaped. *)
+  }
+  type tag = [ `Generic | `Unix ]
+
+  let pid t = t.pid
+
+  let start_waiting t =
+    t.waiting <- true;
+    Fd.use_exn "process_wait" t.handle @@ fun h ->
+    ignore (
+      Sched.enter (fun sched k ->
+          Sched.await_thread sched k ~finished:(Promise.resolve t.resolve)
+            (fun () -> eio_process_wait h))
+      : int)
+
+  let await t =
+    if not t.waiting then start_waiting t;
+    let code =
+      match Promise.await t.exited with
+      | Ok code -> code
+      | Error ex -> raise ex
+    in
+    ignore (Switch.try_remove_hook t.hook : bool);
+    t.hook <- Switch.null_hook;
+    match t.signalled with
+    | Some signum when code = terminated_exit_code -> `Signaled signum
+    | _ -> `Exited code
+
+  (* Windows has no signals: any signal terminates the process. *)
+  let signal t signum =
+    Fd.use t.handle ~if_closed:ignore (fun h ->
+        if eio_process_terminate h terminated_exit_code then t.signalled <- Some signum)
+
+  let stop t =
+    Eio.Cancel.protect @@ fun () ->
+    signal t Sys.sigkill;
+    ignore (await t : Eio.Process.exit_status)
+end
+
+let process_handler = Eio.Process.Pi.process (module Process)
+let process t = Eio.Resource.T (t, process_handler)
+
+module Impl = struct
+  module T = struct
+    type t = unit
+
+    (* CreateProcess takes a Win32 path and not NT object, so we must restrict this.
+       It may be possible to use NtCreateUserProcess directly in the future but
+       the details are pretty complicated so I didnt go into it just yet. *)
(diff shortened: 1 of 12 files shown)


avsm:
I'll need to back out the pipe changes here (see #929) , but the rest of the process spawning is still in reasonable shape.

lib_eio_windows/test/test_process.ml
```diff
@@ -0,0 +1,257 @@
+open Eio.Std
+
+module Process = Eio.Process
+
+let process env = Eio.Stdenv.process_mgr env
+
+let read_all flow =
+  let b = Buffer.create 100 in
+  Eio.Flow.copy flow (Eio.Flow.buffer_sink b);
+  Buffer.contents b
```
talex5, line 10: This is just `Eio.Flow.read_all`.

lib_eio_windows/test/test_process.ml
```diff
@@ -0,0 +1,257 @@
...
+
+let check_status msg expected = function
+  | `Exited code when code = expected -> ()
+  | status ->
+    Alcotest.failf "%s: expected exit %d, got %a" msg expected Process.pp_status status
```
talex5, line 15: Could `Alcotest.of_pp` help here? I'd expect something like:
```ocaml
let status = Alcotest.of_pp Eio.Process.pp_status
```
That would also eliminate the need for a separate `check_signalled` below.

lib_eio_windows/test/test_process.ml
```diff
@@ -0,0 +1,257 @@
...
+  let from_parent, to_child = Eio_unix.pipe sw in
+  let child = Process.spawn ~sw (process env) ~stdin:from_parent ~stdout:to_parent ["findstr"; "hello"] in
+  Eio.Flow.close to_parent;
+  Eio.Flow.copy_string "hello\r\n" to_child;
+  Eio.Flow.close to_child;
+  let out = read_all from_child in
+  check_status "findstr" 0 (Process.await child);
+  Alcotest.(check string) "roundtrip" "hello" (String.trim out)
+
+(* A handle leaking into a sibling would delay its pipe's EOF *)
```
talex5, line 68: I don't see how this tests that. Even if another process inherits a pipe, it will just exit quickly anyway and the test will still pass. Maybe this comment belongs to the next test instead?

lib_eio_windows/test/test_process.ml
```diff
@@ -0,0 +1,257 @@
...
+    check_status token 0 (Process.await child);
+    Alcotest.(check string) token token (String.trim out)
+  in
+  Fiber.List.iter ~max_fibers:8 spawn_one (List.init 1_000 Fun.id)
+
+let test_no_handle_leak env () =
+  Switch.run @@ fun sw ->
+  let r, w = Eio_unix.pipe sw in
+  let w_fd = Option.get (Eio_unix.Resource.fd_opt w) in
+  Eio_unix.Fd.use_exn "pipe" w_fd Unix.clear_close_on_exec;
```
talex5, line 87: If `close_on_exec` is clear, then inheriting the FD is the correct behaviour (at least, that's what the other backends do, and it's what I'd expect).

lib_eio_windows/test/test_process.ml
```diff
@@ -0,0 +1,257 @@
...
+let test_fds_above_2_rejected env () =
+  Switch.run @@ fun sw ->
+  Alcotest.check_raises "fd 3"
+    (Invalid_argument "spawn: only fds 0-2 are supported on Windows (got fd 3)")
+    (fun () ->
+       ignore (Eio_unix.Process.spawn_unix ~sw (process env) ~executable:"cmd.exe"
+                 ~fds:(std_fds @ [3, Eio_unix.Fd.stdin, `Blocking])
+                 ["cmd"; "/c"; "exit"; "0"]))
+
+(* An unlisted descriptor is inherited, as on Unix. *)
```
talex5, line 170: How do we know this worked? `exit` doesn't try to use any of the FDs.

lib_eio_windows/process.ml
```diff
@@ -0,0 +1,142 @@
...
+          Sched.await_thread sched k ~finished:(Promise.resolve t.resolve)
+            (fun () -> eio_process_wait h))
+      : int)
+
+  let await t =
+    if not t.waiting then start_waiting t;
+    let code =
+      match Promise.await t.exited with
+      | Ok code -> code
+      | Error ex -> raise ex
```
talex5, line 61:
```suggestion
      Promise.await_exn t.exited
```

lib_eio_windows/sched.mli
```diff
@@ -38,6 +38,14 @@ val await_timeout : t -> unit Eio_utils.Suspended.t -> Mtime.t -> exit
 
     When [time] is reached, [k] is resumed. Cancelling [k] removes the entry from the timer. *)
 
+val await_thread : t -> 'a Eio_utils.Suspended.t -> ?finished:(('a, exn) result -> unit) -> (unit -> 'a) -> exit
```
talex5, line 41: This `await_thread` thing seems really odd. It looks like it waits for the process *unless* cancelled. But if the spawning switch is cancelled, we want to signal the process and wait for it to exit, so there's no reason to do this.

Maybe something like this:
```ocaml
let watcher = Thread.create (fun h -> Promise.resolve t.exited (eio_process_wait h)) h;
```

The you can just await `t.exited`, and signal the process if the switch ends. `eio_posix` has some suitable logic for this already (might not need `reap` on Windows, not sure whether calling `Thread.join` is necessary):

https://github.com/ocaml-multicore/eio/blob/a8ae279302310566132fbeb16fbae9be0a7689ba/lib_eio_posix/low_level.ml#L681-L695

(it's slightly complicated because we decided that the switch ending should terminate the child, rather than waiting for it to finish)

talex5 (commented):
Just a quick review (I know it needs updating to remove the pipe stuff, so I've ignored those bits).

I think the logic for waiting for the child isn't quite right, and could probably be simplified.

The changes to `nt_path` look generally useful and could be split into another PR if you want to get that out of the way first.

lib_eio_windows/sched.mli
talex5: Here's a simpler way to do the same thing, that doesn't require interrupting and restarting `reap`:
```ocaml
    Fiber.fork_daemon ~sw (fun () ->
        Switch.run @@ fun sw ->
        Fiber.fork_daemon ~sw (fun () ->
            try Fiber.await_cancel ()
            with ex -> signal t Sys.sigkill; raise ex
          );
        Eio.Cancel.protect (fun () -> reap t set_exit_status; `Stop_daemon)
      );
```
If the top `sw` ends, the two daemon fibers will be cancelled. The first signals the process, while the second continues running (because it's protected from cancellation) until it exits.

If the process ends by itself, the inner `sw` finishes and cancels the other fiber.

Issue #899, opened 2026-07-22
open_in/open_out is missing ~follow


samoht:
Thanks you for adding `O_NOFOLLOW` to many path functions, that's super useful. It seems the only remaining ones are open_in/open_out?

talex5:
The functions that have a `follow` argument at the moment are the ones where it makes sense either way (e.g. do you want to `stat` the symlink itself or the thing it points to?).

Adding it to the open calls seems reasonable though (i.e. fail with an error if the last component is a symlink).

But I'm curious what you plan to use it for. `with_subtree` will prevent following a link out of the subtree, so this would only be useful to prevent following symlinks within a subtree.

samoht:
I guess I have 2 kinds of use-cases in mind:

- opening a raw block device by name (/dev/sda, or a disk image path), through Stdenv.fs.  I just want "open the device I was named, or fail". I had to write is a C stub doing open(path, O_RDWR | O_NOFOLLOW | O_CLOEXEC), and since a Unix.file_descr cannot become an Eio.File.rw I had to bind pread/pwrite too. ~follow:false on open_out ~create:Never`` retires all three.

-  building a disk image from a source tree. I stat ~follow:false each entry to classify it, then open the regular files to read them. Another process on the build host can replace an entry with a symlink in between: with_subtree stops that link pointing outside the tree, but a link to another file inside the tree is still followed.

avsm:
Creating a disk image in memory is an excellent reason for distinguishing symlinks indeed!

talex5:
> I had to write is a C stub doing open(path, O_RDWR | O_NOFOLLOW | O_CLOEXEC), and since a Unix.file_descr cannot become an Eio.File.rw I had to bind pread/pwrite too.

I've also added support for this in #937.

PR #937, opened 2026-09-09
Add Eio_unix.File module

talex5:
Allows creating Eio files and flows from FDs.

In many cases it was possible to do this before using `Eio_unix.Net.import_socket_stream`, but that wasn't obvious and doesn't provide support for e.g. `pread` (see e.g. https://github.com/ocaml-multicore/eio/issues/899#issuecomment-5072465263).

### DIFF
--- a/lib_eio/unix/file.mli
+++ b/lib_eio/unix/file.mli
@@ -1,0 +1,17 @@
+open Eio.Std
+
+type rw_ty = [`Unix_fd | Eio.File.rw_ty]
+type ro_ty = [`Unix_fd | Eio.File.ro_ty]
+
+val import_rw : sw:Switch.t -> close_unix:bool -> Unix.file_descr -> [< rw_ty ] r
+(** [import_rw ~sw ~close_unix fd] is a read/write Eio file that uses [fd].
+
+    The file resource will be closed when [sw] finishes.
+
+    The backend takes ownership of [fd] and may change whether it is
+    non-blocking.
+
+    The [close_unix] and [sw] arguments are passed to {!Fd.of_unix}. *)
+
+val import_ro : sw:Switch.t -> close_unix:bool -> Unix.file_descr -> [< ro_ty ] r
+(** [import_ro] is like {!import_rw}, but casts the result to be read-only. *)
--- a/tests/fs.md
+++ b/tests/fs.md
@@ -1547,3 +1547,25 @@ Exception: Failure "Simulated error".
 +"/" / "" = "/"
 - : unit = ()
 ```
+
+# Importing files from FDs
+
+```ocaml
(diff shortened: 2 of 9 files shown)


lib_eio/unix/file.mli
```diff
@@ -0,0 +1,16 @@
+open Eio.Std
+
+type rw_ty = [`Unix_fd | Eio.File.rw_ty]
+type ro_ty = [`Unix_fd | Eio.File.ro_ty]
+
+val import_rw : Fd.t -> [< rw_ty ] r
+(** [import_rw fd] is a read/write Eio file that uses [fd].
+
+    [fd] should be non-blocking (at least for pipes).
```
avsm: I think that pipes do actually block in our current implementation of them on Windows (#929), at least until we shift to IOCP (which does support it I think but is a huge change for that backend)

tests/fs.md
```diff
@@ -1547,3 +1547,23 @@ Exception: Failure "Simulated error".
...
+     Unix.openfile (Eio.Path.native_exn path) [O_CREAT; O_RDWR; O_NONBLOCK] 0o600
+     |> Eio_unix.Fd.of_unix ~sw ~close_unix:true
+     |> Eio_unix.File.import_rw
+     |> Eio.Flow.copy_string "test-data"
+  );
+  Switch.run (fun sw ->
+     Unix.openfile (Eio.Path.native_exn path) [O_RDONLY; O_NONBLOCK] 0
+     |> Eio_unix.Fd.of_unix ~sw ~close_unix:true
+     |> Eio_unix.File.import_ro
+     |> Eio.Flow.read_all
```
avsm: Should we have a test for pseek here as that's what motivated the PR

avsm (approved):
LGTM!

lib_eio/unix/file.mli
talex5: Does the comment need changing then? What happens if you open a pipe in non-blocking mode and pass it to the Windows backend?

tests/fs.md
talex5: OK, done.

lib_eio/unix/file.mli
avsm: It raises ENOTSOCK and all hell breaks loose. But #929 does fix this now by running them in a systhread (i.e. treat it like a file). It does mark a pipe fd as blocking, though (which is what contradicts the comment above)

talex5: OK, I've removed the stuff about pipes and blocking modes. It's now the backend's responsibility to set it to be non-blocking, if appropriate. Opening a pipe as a file doesn't really make sense anyway.

PR #938, opened 2026-09-14
Eio.Path: add ?follow arguments to open functions

### LINKED ISSUE #899: open_in/open_out is missing ~follow

talex5:
Fixes #899, requested by @samoht.


talex5:
@avsm: I tested this on Windows (in CI) and got this:
```
+write <cwd:dir1\\file> -> ok
+write <cwd:link1\\file> -> ok
+write <cwd:link1\\link> -> ok
+write <cwd:dir1\\file> -> ok (no-follow)
+write <cwd:link1\\file> -> ok (no-follow)
+Eio.Io Unix_error (Unknown error, "nofollow", ""), opening <cwd:link1\\link> (no-follow)
```
Not sure what the `Unknown error` is. Would be good if this could turn into an `Eio.Io Fs Symlink` exception.

talex5:
Ah, this doesn't work:

https://github.com/ocaml-multicore/eio/blob/2ae74d1cceb3ef13a9ee75169bfbe3940a8c5fd1/lib_eio_windows/eio_windows_stubs.c#L99

`caml_unix_error` is looking for `ELOOP`, but it previously redefined it!
```c
#undef ELOOP
#define ELOOP                   -WSAELOOP
```
(https://github.com/ocaml/ocaml/blob/774a05dfa6ff8cedca636ffd4cd0b418a7577eba/otherlibs/unix/unixsupport_win32.c#L161-L162)

However, there's another problem now:
```
  caml_eio_windows_openat \??\D:\a\eio\eio\_build\default\lib_eio_windows\test\dir1\file follow=Nofollow
  ASSERT read <cwd:link1\\link> follow=false
  ASSERT Expected symlink error on load
```
When saving, `caml_eio_windows_openat` got the local path `"link"`. But, when loading, the path gets converted to a full path with the symlinks already resolved for some reason.

talex5:
I've modified `open_in` on Windows so it works the same way as `open_out`, and the tests are now passing.

